PR #15, opened 2015-03-09
Fix refresh for publishingInfoList

pjenicot:
Tested on 5.8-HF31

Explanation: The "a4j: outputPanel" with id "publishingInfoList" does not refresh when you would select a publicationTree


atchertchian:
Hi,
Can you please explain how to reproduce the issue, from a user point of view?
Thanks

pjenicot:
Hi Anahide,

I performed the tests with the Nuxeo version 5.8-HF30.

For your information, we have *many domains* in our Nuxeo repository.
In this example, I put only two.

My document "Examen - Statistiques.pdf" is published in the domain 
"*etablissements-domain*".

1] I click on the "publication" tab. By default, It is the domain 
"Default-domain" is displayed


2] I select from the list of domains "etablissements-domain"


3] The publication of my document doesn't appear under the tree publications


4] When I reselect the domain "default-domain", a list of publications 
is refreshed and my document appears.



My fix adds a good result I think :)

Have a nice day !

Pierre

atchertchian:
Hi,

Many thanks for the explanation and the screenshots, i'll create a JIRA issue with this information.

I'm a bit confused by the fix since the table iteration value is still relying on "publishedDocuments" instead of "publishActions.publishedDocuments", so there seems to be a strange behaviour here, i'll try to have a closer look soon.

Cheers

PR #33, opened 2015-10-22
SUPNXP-14791 Retrieve Latest non-Versionable Rendition

ron1:
Patch includes new test TestRenditionService.shouldStoreLatestNonVersionedRendition() that exposes the problem. it also includes code modifications to fix the problem by preventing retrieval of stale, non-versionable renditions. Maven test was successfully run against the patched nuxeo-platform-rendition module using three back-end databases namely, H2, PostgreSQL, and MySQL.


efge:
Merged as 1c3689a2d49c517fd96112dd84a3f7ba2877fb40.

PR #34, opened 2015-10-27
SUPNXP-14841: Fix Rendition Definition Filter Perm checking

ron1:
Fix Rendition Definition (Provider) Filter Permission checking


troger:
Merged, see abf9a8a1073863bdda71a3437719d0fd06eaa4e1.

PR #35, opened 2015-11-04
SUPNXP-15005/14858 Fix mgmt of non-versionable stored renditions and introduce StoredRenditionManager

ron1:
Fix SUPNXP-15005 fix admin reuse of non-versionable stored renditions


efge:
Committed through 557fdfa564f41dd1a395011b2e9f1c83e843b2a1.

PR #37, opened 2015-11-20
Feature nxp 18423 runtime context auto closeable

kevinleturc:
Could you review this ?


efge:
That's good. Could you actually rename the existing destroy() to close(), re-create a deprecated compat destroy() method that just calls close(), and change all existing callers of destroy() to call close() ?

PR #40, opened 2015-12-08
SUPNXP-15301: Return error on failed Lazy Rendition work blob computation

ron1:


efge:
Pushed with changes as 1e0f366c9789f6aac92ed95dc2282a0bfaa1ad0c.

PR #42, opened 2016-03-10
NXP-17875 Set SMTP timeouts in testSendMail()

ron1:
This patch fixes a failure in unit test EmbeddedAutomationClientTest.testSendMail() while running in my environment


vpasquier:
Thank you Ron for this patch.

PR #43, opened 2016-03-11
SUPNXP-15683 Fix Web UI view/delete of placeless doc

ron1:


efge:
Pushed with small edits: a3f11ce8ac3998ac56a3aee452eb3f12b5ad8062

PR #44, opened 2016-03-14
NXP-19143 Allow to configure 401 return code instead of handling prompt for specific auth chains

dmetzler:
I'm opening a PR to be able to discuss on implementation with pairs for review.
So.... do not merge, a T&P will take care of it.


dmetzler:
@chapurlatn @troger @tiry When returning a 401, I've added a `Location` header with the URL of the /login that really handle the login. 

I'm also adding something in the payload which is not localized (Please login at : +url). I don't think it's a really big issue since it's should not be shown to the end user. (the normal login process should be shown to the user)

dmetzler:
PR has been merged by a Test&Push

PR #46, opened 2016-03-23
NXP-19258: Add nuxeoctl register-trial command

akervern:


nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1267,48 @@ public void run() {
         }
     }
 
+    protected boolean registerTrial() throws IOException {
+        if (params.length < 2) {
+            printLongHelp();
+            errorValue = EXIT_CODE_ERROR;
+            return false;
```
jcarsique: Ask for the missing parameters if in interactive mode

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1267,48 @@ public void run() {
...
+        String projectName = "project1";
+        if (params.length >= 3) {
+            projectName = params[2];
+        }
+        registration.put("connectreg:projectName", projectName);
+
+        try {
+            LauncherRegisterHelper.registerTrial(registration);
+        } catch (RegistrationException e) {
+            System.err.println(e.getMessage() + ":");
```
jcarsique: Don't use temp sysout/syserr during development: this class has a few cases of wanted use of sysout instead of log.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1267,48 @@ public void run() {
...
+            e.getErrors().forEach(err -> System.err.println("\t- " + err.getMessage()));
```
jcarsique: Use either `org.nuxeo.launcher.NuxeoLauncher.cset`, either the try/catch in `org.nuxeo.launcher.NuxeoLauncher.main(String[])`. Adapt it to handle multiple errors if necessary.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterHelper.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.commons.io.IOUtils;
+import org.nuxeo.connect.NuxeoConnectClient;
+import org.nuxeo.connect.registration.RegistrationException;
+
+/**
+ * @author <a href="mailto:ak@nuxeo.com">Arnaud Kervern</a>
+ * @since 8.2
+ */
+public class LauncherRegisterHelper {
+    public static String[] promptPasswords() throws IOException {
```
jcarsique: why later delegating the password comparison? 
use char[] as much as possible when dealing with passwords

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterHelper.java
```diff
@@ -0,0 +1,67 @@
...
+        String[] ret = new String[2];
+
+        Console console = System.console();
+
+        if (console != null) {
+            ret[0] = new String(console.readPassword("Please enter your password: "));
+            ret[1] = new String(console.readPassword("Please re-enter your password: "));
+        } else {
+            String password = new String(IOUtils.toByteArray(System.in)).trim();
```
jcarsique: See for instance `org.nuxeo.common.codec.Crypto.getChars(byte[])`

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterHelper.java
```diff
@@ -0,0 +1,67 @@
...
+        return ret;
+    }
+
+    public static boolean promptAcceptTerms() {
+        Console console = System.console();
+        if (console != null) {
+            String terms;
+            do {
+                terms = console.readLine("Read and accept the Nuxeo Trial Terms and Conditions - https://www.nuxeo.com/en/services/nuxeo-trial-tc (y/n): ");
+            } while (!(terms.startsWith("y") || terms.startsWith("n")));
```
jcarsique: why loop? terms can be null. I would check for `"y".equals(terms) || "yes".equals(terms)` once.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterHelper.java
```diff
@@ -0,0 +1,67 @@
...
+
+            return terms.startsWith("y");
+        } else {
+            return true;
```
jcarsique: If not interactive, then the user is supposed to have accepted the terms? If so, it's worth to display an informational message.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1267,48 @@ public void run() {
...
+        registration.put("termsAndConditions", "true");
+
+        String projectName = "project1";
+        if (params.length >= 3) {
+            projectName = params[2];
+        }
+        registration.put("connectreg:projectName", projectName);
+
+        try {
+            LauncherRegisterHelper.registerTrial(registration);
```
jcarsique: registration is missing the instance description

pom.xml
```diff
@@ -121,7 +121,7 @@
     <nuxeo.webengine.version>8.2-SNAPSHOT</nuxeo.webengine.version>
     <nuxeo.features.version>8.2-SNAPSHOT</nuxeo.features.version>
 
-    <nuxeo.connect.client.version>1.4.24</nuxeo.connect.client.version>
+    <nuxeo.connect.client.version>1.4.25-SNAPSHOT</nuxeo.connect.client.version>
```
jcarsique: missing exclude in maven-enforcer-plugin in nighlty profile, you will break the nightly build

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -515,6 +517,7 @@ protected boolean commandRequiresNoGUI() {
             + "        nuxeoctl [mp-list|mp-listall|mp-init|mp-update] [command parameters] [-d [<categories>]|-q|--clid <arg>|--xml|--json]\n\n"
             + "        nuxeoctl [mp-reset|mp-purge|mp-hotfix|mp-upgrade] [command parameters] [-d [<categories>]|-q|--clid <arg>|--xml|--json|--accept <true|false|yes|no|ask>]\n\n"
             + "        nuxeoctl [mp-add|mp-install|mp-uninstall|mp-remove|mp-set|mp-request] [command parameters] [-d [<categories>]|-q|--clid <arg>|--xml|--json|--nodeps|--relax <true|false|yes|no|ask>|--accept <true|false|yes|no|ask>|-s|-im]\n\n"
+            + "        nuxeoctl [register-trial] <email> <company_name> [<project_name>]\n\n"
```
jcarsique: square-brackets are wrong on those 10 commands, please remove them: the command is not optional

as said later, I would make email and company_name optional too (note that the other commands use camelcase or simple spaces but no underscores => companyName, projectName)

list your command in `org.nuxeo.launcher.NuxeoLauncher.OPTION_HELP_DESC_COMMANDS`

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1267,48 @@ public void run() {
...
     }
 
+    protected boolean registerTrial() throws IOException {
+        if (params.length < 2) {
+            printLongHelp();
+            errorValue = EXIT_CODE_ERROR;
+            return false;
+        }
+
+        Console console = System.console();
```
jcarsique: unused

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1267,48 @@ public void run() {
...
+        Map<String, String> registration = new HashMap<>();
+        registration.put("email", params[0]);
+        registration.put("company", params[1]);
+
+        String[] pwds = LauncherRegisterHelper.promptPasswords();
+        registration.put("password", pwds[0]);
+        registration.put("password_verif", pwds[1]);
```
jcarsique: duplicate

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -337,7 +338,7 @@
             "mp-update", "status", "showconf", "mp-show", "mp-set", "config", "encrypt", "decrypt", OPTION_HELP };
 
     private static final String[] COMMANDS_NO_RUNNING_SERVER = { "pack", "mp-init", "mp-purge", "mp-add", "mp-install",
-            "mp-uninstall", "mp-request", "mp-remove", "mp-hotfix", "mp-upgrade", "mp-reset", "mp-update", "mp-set" };
+            "mp-uninstall", "mp-request", "mp-remove", "mp-hotfix", "mp-upgrade", "mp-reset", "mp-update", "mp-set", "register-trial" };
```
jcarsique: Is a running server an issue? (just a question, I don't know)
The must be some available code in the server to get a refresh since the registration is also available in the GUI.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1277,94 @@ public void run() {
         }
     }
 
+    protected boolean registerRemoteInstance() throws IOException, ConfigurationException, PackageException {
+        String username;
+        String password;
+        String project;
+        String description = "";
+        NuxeoClientInstanceType type = null;
```
efge: `= null` not needed I think

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1277,94 @@ public void run() {
...
+            registration.put("company", LauncherRegisterPromptHelper.promptCompany());
+
+            String pwd = LauncherRegisterPromptHelper.promptPassword(true);
+            registration.put("password", pwd);
+            registration.put("password_verif", pwd);
+
+            registration.put("connectreg:projectName", LauncherRegisterPromptHelper.promptProjectName());
+            registration.put("description", LauncherRegisterPromptHelper.promptDescription());
+        } else {
+            String[] input = new String(IOUtils.toByteArray(System.in)).split("\n");
```
efge: Should then probably check `input` length and display error messages if malformed.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -2298,6 +2401,14 @@ protected ConnectBroker getConnectBroker() throws IOException, PackageException
         return connectBroker;
     }
 
+    protected ConnectRegistrationBroker getConnectRegisterationBroker() throws IOException, PackageException {
+        if (connectRegistrationBroker == null) {
+            getConnectBroker(); // Ensure ConnectBroker is instanciated too.
```
efge: instanciated -> instantiated

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
```diff
@@ -0,0 +1,175 @@
...
+    public static String promptMail() {
+        return prompt("Mail Adress: ", "^.+@.+\\..{2,4}", "Mail is invalid.");
+    }
+
+    public static String promptUsername() {
+        return prompt("Username: ", "^\\w+", "Username can not be empty.");
+    }
+
+    public static String promptCompany() {
+        return prompt("Company: ", "^\\w+", "Company can't be empty.");
```
efge: "Username can not" vs "Company can't" -> use "cannot" in both places.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
```diff
@@ -0,0 +1,175 @@
...
+        }
+        return pwd;
+    }
+
+    public static NuxeoClientInstanceType promptInstanceType() {
+        NuxeoClientInstanceType ret;
+        Console console = System.console();
+
+        do {
+            String s = console.readLine("Instance type (dev|preprod|prod): ");
```
efge: Maybe show in the message that `dev` is the default.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
```diff
@@ -0,0 +1,175 @@
...
+
+    protected static String promptPassword(String message) throws IOException {
+        Console console = System.console();
+        return new String(console.readPassword(message));
+    }
+
+    public static boolean promptAcceptTerms() {
+        Console console = System.console();
+        if (console != null) {
+            String terms = console.readLine("Read and accept the Nuxeo Trial Terms and Conditions - https://www.nuxeo.com/en/services/nuxeo-trial-tc (Y/n): ");
```
efge: Put the URL itself in a constant at the top of the file, we may want to change it separately at some point.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
```diff
@@ -0,0 +1,175 @@
...
+            return terms != null && (terms.isEmpty() || "y".equalsIgnoreCase(terms) || "yes".equalsIgnoreCase(terms));
+        } else {
+            log.info("Read Nuxeo Trial Terms and Conditions - https://www.nuxeo.com/en/services/nuxeo-trial-tc");
```
efge: Indeed the TOC URL it's being reused here.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
```diff
@@ -0,0 +1,175 @@
...
+            return true;
+        }
+    }
+
+    public static String promptProjectId(List<ConnectProject> projects) throws ConfigurationException {
+        Console console = System.console();
+        if (console == null) {
+            throw new ConfigurationException("Can't configure project in an non-interactive shell.");
+        }
+
```
efge: What if `projects` is empty?

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
```diff
@@ -0,0 +1,175 @@
...
+            String projectName = "";
+            do {
+                if (i > 0 && !PlatformUtils.isWindows()) {
+                    // Remove "next page" line to only have projects
+                    System.out.print("\33[1A\33[2K");
+                }
+
+                int fromIndex = i * PAGE_SIZE;
+                int toIndex = (i + 1) * PAGE_SIZE;
+                if (toIndex > projects.size()) {
```
efge: `>=` I think.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -506,15 +513,17 @@ protected boolean commandRequiresNoGUI() {
...
+            + "        nuxeoctl wizard [-d [<categories>]|-q|--clid <arg>|--gui <true|false|yes|no>]\n\n"
+            + "        nuxeoctl stop [-d [<categories>]|-q|--gui <true|false|yes|no>]\n\n"
+            + "        nuxeoctl start|restart|console|startbg|restartbg [-d [<categories>]|-q|--clid <arg>|--gui <true|false|yes|no>|--strict|-hdw]\n\n"
+            + "        nuxeoctl mp-show [command parameters] [-d [<categories>]|-q|--clid <arg>|--xml|--json]\n\n"
+            + "        nuxeoctl mp-list|mp-listall|mp-init|mp-update [command parameters] [-d [<categories>]|-q|--clid <arg>|--xml|--json]\n\n"
+            + "        nuxeoctl mp-reset|mp-purge|mp-hotfix|mp-upgrade [command parameters] [-d [<categories>]|-q|--clid <arg>|--xml|--json|--accept <true|false|yes|no|ask>]\n\n"
+            + "        nuxeoctl mp-add|mp-install|mp-uninstall|mp-remove|mp-set|mp-request [command parameters] [-d [<categories>]|-q|--clid <arg>|--xml|--json|--nodeps|--relax <true|false|yes|no|ask>|--accept <true|false|yes|no|ask>|-s|-im]\n\n"
+            + "        nuxeoctl register\n\n"
+            + "        nuxeoctl register-trial\n\n"
```
jcarsique: Parameters and options are missing for both commands

akervern: Nope, guess not. As you requested there is no parameters and not sure which options are taken into account... as i don't see any global options; but no one is needed in the new commands.

To be clearer; following what you expect, parameters are passed using STDIN or prompted to the user if makes sense. So; as this a new pattern in the whole possible commands; I have no idea how to comment them.

jcarsique: I see what you mean. That makes the parameters optional to call the command but when you know them, you should be able to pass them as parameters in the command line (as a single call to nuxeoctl). I was thinking about the same behavior as for the `config --set` command for instance.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1277,96 @@ public void run() {
...
+        CommandInfo commandInfo = cset.newCommandInfo("register-trial");
+
+        Map<String, String> registration = new HashMap<>();
+        if (System.console() != null) {
+            registration.put("email", LauncherRegisterPromptHelper.promptMail());
+            registration.put("company", LauncherRegisterPromptHelper.promptCompany());
+
+            String pwd = LauncherRegisterPromptHelper.promptPassword(true);
+            registration.put("password", pwd);
+            registration.put("password_verif", pwd);
```
jcarsique: that is really too much weird and useless

pom.xml
```diff
@@ -121,7 +121,7 @@
     <nuxeo.webengine.version>8.3-SNAPSHOT</nuxeo.webengine.version>
     <nuxeo.features.version>8.3-SNAPSHOT</nuxeo.features.version>
 
-    <nuxeo.connect.client.version>1.4.24</nuxeo.connect.client.version>
+    <nuxeo.connect.client.version>1.4.25-SNAPSHOT</nuxeo.connect.client.version>
```
jcarsique: accordingly relax the SNAPSHOT plugin enforcement

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+        return pwd;
+    }
+
+    public static NuxeoClientInstanceType promptInstanceType() {
+        NuxeoClientInstanceType ret;
+        Console console = System.console();
+
+        do {
+            String s = console.readLine("Instance type (dev|preprod|prod) (default: dev): ");
```
jcarsique: Format usage on nuxeoctl for the default value on prompt is: 
`Do you want to continue (yes/no)? [yes]`

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
akervern: Yes, you are right. But you know... using the same stuff twice... do not have two separate endpoints doing the same... Sometimes you have to do this kind of stuff.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
```diff
@@ -0,0 +1,183 @@
...
+
+        Console console = System.console();
+        String value = null;
+        do {
+            if (value != null) {
+                console.printf(error + "\n", value);
+            }
+            value = console.readLine(message);
+        } while (!value.matches(regex));
+
```
jcarsique:
```
String value = console.readLine(message);
while (!value.matches(regex)) {
    console.printf(error + "\n", value);
    value = console.readLine(message);
}
```

Also, `console.readLine(message)` returns << A string containing the line read from the console, not including any line-termination characters, or **null** if an end of stream has been reached. >>
So you must add `value == null ||` or more likely `value !=null &&` to the condition, depending on the wanted behavior (which must be described in the Javadoc).

akervern: Thanks you. Very interesting.

pom.xml
akervern: I still not understand why a snapshot branch is not allowing snapshot dependencies... but ok. It's just a pain. 😓

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1277,96 @@ public void run() {
...
+                throw new IOException("Wrong number of arguments.");
+            }
+            username = input[0];
+            password = input[1];
+            project = input[2];
+            type = NuxeoClientInstanceType.fromString(input[3]);
+            if (input.length > 3) {
+                description = StringUtils.join(Arrays.copyOfRange(input, 4, input.length), "\n");
+            }
+        }
```
jcarsique: This case (the else block) must be described in the usage, it cannot be guessed.
Anyway, I'm wondering why not rather do like in other similar nuxeoctl commands: 
```
if username not set (passed as parameter, option or context) {
    if console !=null {
        read username from console
    } else {
        read username from stdin
}
and so on
```
Refactor that code up to the prompt method and don't bother with the console in this method.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1262,6 +1277,96 @@ public void run() {
...
+
+        if (type == null) {
+            type = NuxeoClientInstanceType.DEV;
+        }
```
jcarsique: type should default to DEV if not provided, like in `promptInstanceType()` but checked here, a null value can also mean a wrong input (an invalid instance type) and you should raise an exception and exit.

akervern: I'd agree with you; but as the value is not used anywhere except stored... I prefer to ask you to `create a ticket` and we'll handling it when it will be relevant and we know what we really want to do with the InstanceType.

pom.xml
jcarsique: Not a pain: one line two uncomment in the same file.

It is allowed on development branches, not on master or maintenance branches. And certainly not on release.

SNAPSHOT dependencies are usually a mistake, coming from a transitive path or a forgotten work in progress and lead to random errors. It also increases the workload and the risk at release time when it comes mandatory to replace them with a release version in order to satisfy the most important rule: **a release build must be reproducible**.
That's why the default rule is not to use SNAPSHOT dependencies on master or maintenance branches, except during one development cycle, on purpose, during their own parallel development and knowing that they will have to be replaced with a release version before the release of the current code.
So, for an average of three or four library cases over all the code, it's worth to explicitly mention those very few exceptions for which the tooling will not raise an alert.

Please read https://doc.nuxeo.com/display/CORG/Maven+Coding+Rules#MavenCodingRules-SNAPSHOTdependencies

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
akervern: Can you point me an example, please? I didn't find anything matching the same behavior we expect here. Sorry, i'm a bit lost in this class.

pom.xml
akervern: Thank you but I was not discussing the whole global idea about snapshot dependencies; even if you list some generic cases that can be discussed, but it's out of topic.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterPromptHelper.java
jcarsique: !fixed

pom.xml
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/connect/LauncherRegisterHelper.java
jcarsique: !fixed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1277,79 +1285,348 @@ public void run() {
...
+     * @throws ConfigurationException If the user input is read from stdin and is {@code null} or does not match the
+     *             {@code regex}
+     * @throws IOException
+     * @since 8.3
+     */
+    public String promptEmail() throws IOException, ConfigurationException {
+        final String message = "Email Address: ";
+        final String error = "Invalid email address.";
+        EmailValidator validator = EmailValidator.getInstance();
+        String email = null;
```
efge: `= null` not needed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1277,79 +1285,348 @@ public void run() {
...
+     * @return The user input. Never null.
+     * @throws ConfigurationException If the user input is read from stdin and is {@code null} or does not match the
+     *             {@code regex}
+     * @throws IOException
+     * @since 8.3
+     */
+    public String prompt(String message, String regex, String error) throws ConfigurationException, IOException {
+        boolean doRegexMatch = StringUtils.isNotBlank(regex);
+        String value = null;
+        Console console = System.console();
```
efge: `= null` not needed

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1277,79 +1285,348 @@ public void run() {
...
+     * Prompts for a valid email address according to RFC 822 standards. The remote service may apply stricter
+     * constraints on email validation such as some black listed domains.
+     *
+     * @return The user input. Never null.
+     * @throws ConfigurationException If the user input is read from stdin and is {@code null} or does not match the
+     *             {@code regex}
+     * @throws IOException
+     * @since 8.3
+     */
+    public String promptEmail() throws IOException, ConfigurationException {
```
efge: Code looks very similar to `prompt()`, probably can be factored by passing the regex match and the validator as lambdas.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1277,79 +1285,348 @@ public void run() {
...
 
-        if (type == null) {
-            type = NuxeoClientInstanceType.DEV;
+    /**
+     * Register the instance, generating the CLID.
+     *
+     * <pre>
+     * {@code
+     * nuxeoctl register [<username> [<project> [<type> <description>] [pwd]]]
+     * 0/1 param:        [<username>]
```
efge: `[<username> [pwd]]`

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1277,79 +1285,348 @@ public void run() {
...
         return true;
     }
 
-    protected boolean registerTrial() throws IOException, ConfigurationException, PackageException {
-        getConnectBroker(); // Initialize NuxeoConnectClient Oo
+    /**
+     * @throws IOException
+     * @throws ConfigurationException
+     * @throws PackageException
+     * @since 8.3
```
efge: Uninformative `@throws`, to be removed.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1263,6 +1286,368 @@ public void run() {
...
+            registration.put("email", email);
+
+            String company = params[1];
+            if (!Pattern.matches("^\\w+$", company)) {
+                throw new ConfigurationException("Company field is mandatory.");
+            }
+            registration.put("company", company);
+
+            String projectName = params[2];
+            if (!Pattern.matches("^(?:[-\\w]+|)$", projectName)) {
```
akervern: Same regex using twice. (1) Should be extracted.

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1263,6 +1286,368 @@ public void run() {
...
+            }
+            registration.put("connectreg:projectName", projectName);
+
+            registration.put("description", params[3]);
+        } else {
+            registration.put("email", promptEmail());
+            registration.put("company", prompt("Company: ", "^\\w+$", "Company field is mandatory."));
+            registration.put(
+                    "connectreg:projectName",
+                    prompt("Project name: ", "^(?:[-\\w]+|)$",
```
akervern: (1)

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1263,6 +1286,368 @@ public void run() {
...
+        Map<String, String> registration = new HashMap<>();
+        if (params.length > 3) {
+            String email = params[0];
+            if (!EmailValidator.getInstance().isValid(email)) {
+                throw new ConfigurationException("Invalid email address.");
+            }
+            registration.put("email", email);
+
+            String company = params[1];
+            if (!Pattern.matches("^\\w+$", company)) {
```
akervern: Same here (2)

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1263,6 +1286,368 @@ public void run() {
...
+            String projectName = params[2];
+            if (!Pattern.matches("^(?:[-\\w]+|)$", projectName)) {
+                throw new ConfigurationException("Project name can only contain alphanumeric characters and dashes.");
+            }
+            registration.put("connectreg:projectName", projectName);
+
+            registration.put("description", params[3]);
+        } else {
+            registration.put("email", promptEmail());
+            registration.put("company", prompt("Company: ", "^\\w+$", "Company field is mandatory."));
```
akervern: (2)

nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1277,79 +1285,348 @@ public void run() {
...
-            if (input.length < 3) {
-                throw new IOException("Wrong number of arguments.");
+        if (params.length > 3) {
+            String email = params[0];
+            if (!EmailValidator.getInstance().isValid(email)) {
+                throw new ConfigurationException("Invalid email address.");
             }
-            registration.put("email", input[0]);
-            registration.put("company", input[1]);
+            registration.put("email", email);
```
efge: Please extract these as constants ("email", "company", etc.).

PR #49, opened 2016-04-05
Compute the blob digest using a buffer

jendib:
The current implementation is computing the digest byte by byte, which is quite slow. Using a buffer is roughly 10x faster.


efge:
Hi,
Thanks for the patch but this code is already using org.apache.commons.codec.binary.Base64 since 2012:
https://github.com/nuxeo/nuxeo/blob/2b9cbdf45cc868faf8f6f29a27beb61c2082f3b4/src/main/java/org/nuxeo/common/utils/Base64.java

Why did you need this?

efge:
Thanks for the report. We'll go even further and use Commons Codec directly, see 87a02f223afbc68054e93c4797372d938f993275.

PR #52, opened 2016-04-07
NXP-19425: Rework transaction management on reload

kevinleturc:
Duplicates by https://jira.nuxeo.com/browse/NXP-19150


efge:
This needs an explanation (in NXP and code) of why suspending/resuming the transaction was bad, and why committing the previous one and starting a new one is better and acceptable.

PR #105, opened 2016-05-31
NXP-19664 add a parameter to define the user/group suggestion widget …

dmetzler:
Added  a parameter to define the default `,` character used to separate values in the user/group suggestion widgets. This allows to define its own and thus use group with `,` character in their id.


review-ninja:
<a href="https://app.review.ninja/nuxeo/nuxeo/pull/105" target="_blank">ReviewNinja

efge:
-1, on master and 7.10 `nuxeo.userwidget.separator.char` should not be a nuxeo.conf property but a configuration service property.

dmetzler:
As it is used in a contribution (`layouts-contrib.xml`), defining it in another contribution doesn't work. I tried different `<require>` directives but it doesn't work either. The variable expansion seems to take place when the file is read on bundle scan.
and then it wouldn't be possible to override it, since the override would take place at the end.

Merge has been done by T&P

PR #139, opened 2016-06-28
NXP-19824: fix version label after copy then checkin

efge:


review-ninja:
<a href="https://app.review.ninja/nuxeo/nuxeo/pull/139" target="_blank">ReviewNinja

PR #179, opened 2016-08-03
NXP-20198: Override the getGroup() method

glefevre:
Override the getGroup() method in order to get the computed group when the context is provided.


review-ninja:
<a href="https://app.review.ninja/nuxeo/nuxeo/pull/179" target="_blank">ReviewNinja

efge:
If this solves an NPE then I'd like a unit test that fails with that NPE when your change is not applied.

efge:
👍

PR #219, opened 2016-09-09
NXP-19012: automation trace fails in 7.10

richardsd:
In an automation chain which includes an automation script, and automation tracing is enabled, a UnsupportedOperationException was thrown when generating the trace. This was due to a problem with the context unwrap from **DocumentScriptingWrapper** to **DocumentModel**.

This commit provides a fix (and unit test) for the issue in 7.10, but it is not necessary for master, as the new **WrapperHelper** solves the problem.


efge:
👍

kevinleturc:
👍 
if issue doesn't occur on master due to recent changes, you can add your unit test to master in order to check further regression

nelsonsilva:
👍

PR #243, opened 2016-09-23
NXP-20344: deactivate no-cache header on resources but keep it on pages

nelsonsilva:
@vlad @florent wdyt? cache control was already configurable in RequestControllerService through filterConfig (cached=true|false) so just made sure we kept a single Cache-Control header and removed "Pragma" since specs say it's mean to be used in clients' requests not server responses.

Note: Could use a review of default filter config for more aggressive caching of static resources.


review-ninja:
<a href="https://app.review.ninja/nuxeo/nuxeo/pull/243" target="_blank">ReviewNinja

efge:
👍

PR #249, opened 2016-09-27
NXP-20401: fix repository retrieval for comments in multi-repo setup

Gabez0r:


review-ninja:
<a href="https://app.review.ninja/nuxeo/nuxeo/pull/249" target="_blank">ReviewNinja

egiuly-nuxeo:
@Gabez0r 
What do you think about adding a unit test so that this bug won't happen again?
Other question: You've changed `CommentManagerImpl.java`, but should some other classes using the `getGraphByName` method be changed too?

efge:
Estelle has good points Gabriel. A unit test for multi-repo is a PITA though. But the other questions about the API stand.

Gabez0r:
Tests would prevent this from happening again. I will create a separate non-supcom ticket to address unit tests on the CommentManager as soon as time permits.

As for the other points,  there are indeed several components using it throughout the platform. A thorough review should be performed to assess how each of those components are affected and act accordingly. I will also create another ticket for this one.

PR #303, opened 2016-10-24
NXP-20860: gitfunctions.sh and nxutils.sh rewrite

jcarsique:
https://jira.nuxeo.com/browse/NXP-20860
https://jira.nuxeo.com/browse/NXBT-1151
(later for https://jira.nuxeo.com/browse/NXBT-1343 too)

Expected backward compliance with 'gitf' and 'gitfa', as much as possible
Less porosity on variables
Allow to retrieve the addons list
Uniformize behavior regardless the current directory (ie: root or addons)
Verbose (default) and quiet modes
BASH, ZSH and (supposed) POSIX compliance
New features: 'shr' and modules list export. Needed to fix race conditions such as module addition/removal.

```
Usage: gitf [-r] [-q] [-h|-?] <[--] Git instructions>
Recursively executes the given Git command on the current and its direct children Git repositories.

Usage: gitfa [-q] [-h|-?] <[--] Git instructions>
Recursively executes the given Git command on the current, its direct children Git repositories and browse the child Maven modules of $PARENT_MODULES directories.

Usage: shr [-a] [-q] [-h|-?] <[--] Shell instructions>
Recursively executes the given Shell command on the current and its direct children Git repositories.

    -h,-?   Show this help message and exit.
    -a       The recurse will continue on the child Maven modules of $PARENT_MODULES directories.
    -q       Quiet mode (default is verbose).
    -r        The recurse will continue on all children Git repositories.
    --        Optional separator between the options and the instructions which may start with a dash.
    PARENT_MODULES Environment variable. Defaults to 'addons addons-core' if not set.
    LIST_<parent module> Environment variable that can be set to restrict the children modules to a fixed list. For instance: LIST_ADDONS_CORE="nuxeo-core-storage-marklogic" or LIST_ADDONS="nuxeo-shell nuxeo-quota".
```


scripts/gitfunctions.sh
```diff
@@ -2,43 +2,196 @@
...
-      (cd "$dir" ; git "$@")
+      [ "$quiet" != true ] && echo "\$> git $git_args"
+      (eval git $git_args) || true
       echo
     fi
-  done
+    for dir in $(ls -d */); do
+      dir=${dir%%/};
+      if [ -e "$dir"/.git ]; then
+        ( cd "$dir" || continue
```
efge: Are you sure you can do `continue` in a subshell???

scripts/gitfunctions.sh
```diff
@@ -2,43 +2,196 @@
...
+fi
 . $HERE/nxutils.sh
 
-# usage: "gitf [git instructions]"
 gitf() {
-  for dir in . *; do
-    if [ -e "$dir"/.git ]; then
+  # Parse options
+  local OPTIND=1 # getopts reset
+  local recurse=false
```
efge: I prefer the idiom of `recurse=y` for true and `recurse=` (empty) for false, and test with `-n` and `-z`

scripts/gitfunctions.sh
```diff
@@ -2,43 +2,196 @@
...
-HERE=$(cd $(dirname $BASH_SOURCE); pwd -P)
+
+if [ -n "$BASH" ]; then
+  HERE=$(cd $(dirname $BASH_SOURCE); pwd -P)
+elif [ -n "ZSH_NAME" ]; then
+  HERE=$(cd $(dirname ${(%):-%N}); pwd -P)
+else
+  HERE=$(cd $(dirname $_); pwd -P)
+fi
 . $HERE/nxutils.sh
```
efge: I hate shell scripts that include other scripts. See the hacks to find HERE... If you do that you might as well write in python. Can't this be expanded directly in the main script?

scripts/nxutils.sh
```diff
@@ -17,38 +17,32 @@
...
+  local __resultvar=$1
+  local _list=LIST_$(echo $(basename $PWD)|tr '-' '_'|tr '[:lower:]' '[:upper:]')
+  if [ -z "${!_list}" ]; then
+    [ "$quiet" != true ] && echo "Modules list calculated from POM"
+    local _result="$(grep '<module>' pom.xml |cut -d ">" -f 2 |cut -d "<" -f 1|sort|uniq|tr '\n' ' ')"
+  else
+    [ "$quiet" != true ] && echo "Modules list set from environment variable: $_list"
+    local _result="${!_list}"
+  fi
+  eval "$__resultvar='$_result'"
```
efge: Doing an `eval` to put the result in the variable whose name is passed as an argument looks like an ugly hack. Why not just `echo` the result, and the caller does VAR="$(thisfunction)"

scripts/gitfunctions.sh
```diff
@@ -2,43 +2,196 @@
...
+  [ "$1" = "--" ] && shift
+  local git_args=$@
+
+  function git_command {
+    local dir
+    if [ -e .git ]; then
+      dir=$(basename $PWD)
       echo "[$dir]"
-      (cd "$dir" ; git "$@")
+      [ "$quiet" != true ] && echo "\$> git $git_args"
```
efge: This would be simply `test -n $quiet && echo...` if you use my suggestion for true/false.

scripts/gitfunctions.sh
```diff
@@ -2,43 +2,196 @@
...
+      (eval git $git_args) || true
```
efge: Is an `eval` really needed?

scripts/gitfunctions.sh
```diff
@@ -2,43 +2,196 @@
...
       echo
     fi
-  done
+    for dir in $(ls -d */); do
```
efge: `for dir in $(find . -type d -depth 1 -prune); do`

scripts/gitfunctions.sh
```diff
@@ -2,43 +2,196 @@
...
+  local dir=$(basename $PWD)
+  git_command $dir
+  # Detect if we are at Nuxeo Platform repository's root
+  if [ -d "nuxeo-common" ]; then
+    local dir
+    for dir in $(ls -d */); do
+      ( dir=${dir%%/};
+        cd "$dir" || continue
+        git_command "$dir"
+        if [[ "$parent_modules" =~ (^|[[:space:]])"$dir"($|[[:space:]]) ]]; then
```
efge: I'd just do (portable) (note the spaces):

    case " $parent_modules " in
        *" $dir "*) _execute_on_modules "git_command"
            ;;
    esac

scripts/gitfunctions.sh
jcarsique: @kevinleturc any particular need for creating nxutils.sh or was it design purpose only?

scripts/gitfunctions.sh
jcarsique: Understood. I'll look at it. However I need the variable being set to ensure that it will be local, determined and not contaminated from outside or contaminating outside (especially the `recurse` one).

scripts/gitfunctions.sh
jcarsique: Yes, for variable expansion.

scripts/nxutils.sh
jcarsique: Not a hack but a commonly recommended way to set a local var by reference (from a function in this case); while using echo is a hack (or a bad practice in complex cases).
The called function must be able to echo text, debugging information and be resilient to wrong outputs.
Note that the "namerefs" is a built-in feature since BASH 4.3.

efge: On macOS bash is version 3.2.57.

nxmatic:
Cannot run gitfa on macos/zsh. 

```
diff --git i/scripts/gitfunctions.sh w/scripts/gitfunctions.sh
index 107fc34..36db951 100755
--- i/scripts/gitfunctions.sh
+++ w/scripts/gitfunctions.sh
@@ -25,7 +25,7 @@
 function _retrieve_modules {
   local __resultvar=$1
   local _list=LIST_$(echo $(basename $PWD)|tr '-' '_'|tr '[:lower:]' '[:upper:]')
-  if [ -z "${!_list}" ]; then
+  if [ -z "${(P)_list}" ]; then
     [ "$quiet" != true ] && echo "Modules list calculated from POM"
     local _result="$(grep '<module>' pom.xml |cut -d ">" -f 2 |cut -d "<" -f 1|sort|uniq|tr '\n' ' ')"
   else
@@ -42,7 +42,7 @@ function _execute_on_modules {
   local _modules=""
   _retrieve_modules _modules
   [ "$quiet" != true ] && echo -e "Execute on modules: $_modules\n"
-  for dir in $_modules; do
+  for dir in $(echo $_modules|tr '\n' ' '); do
     if [ -d $dir ]; then
       ( cd $dir
         $f $dir

```

scripts/nxutils.sh
```diff
@@ -1,54 +0,0 @@
...
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
-# Contributors:
-#   Kevin Leturc
-#
-
-function _retrieve_modules {
```
kevinleturc: Be aware that T&P template use these functions.

jcarsique: I did not find any reference to `_retrieve_modules` function, only to `_execute_on_modules` which has not changed.

jcarsique:
merged

PR #305, opened 2016-10-25
NXP-20802: Avoid HTML detection in DefaultFulltextParser

ron1:


nuxeo-core/nuxeo-core-storage/src/main/java/org/nuxeo/ecm/core/storage/FulltextExtractorWork.java
```diff
@@ -209,4 +214,17 @@ protected BlobHolder convert(BlobHolder blobHolder) throws ConversionException {
         return conversionService.convert(ANY2TEXT, blobHolder, null);
     }
 
+    protected static class MimetypeAndText implements Serializable {
```
efge: Can you please use `StringBlob` to hold the mime type and text rather than this specialized class?

nuxeo-core/nuxeo-core-storage/src/main/java/org/nuxeo/ecm/core/storage/DefaultFulltextParser.java
```diff
@@ -61,8 +74,8 @@ public String parse(String s, String path) {
      * This can be subclassed.
      */
     @Override
-    public void parse(String s, String path, List<String> strings) {
-        s = preprocessField(s, path);
+    public void parse(String s, String path, String mimeType, DocumentRef docRef, List<String> strings) {
```
efge: `docRef` is not used in the code. If you actually need it in some custom code, please make a mention in the implementation that this parameter is currently unused but has some use cases for potential subclasses.

However a `DocumentRef` is not enough to uniquely determine a document, as you have no way of knowing the repository name. This would be manifest if there was actual code trying to use this parameter. You should use a `DocumentLocation` instead.

nuxeo-core/nuxeo-core-storage/src/main/java/org/nuxeo/ecm/core/storage/FulltextExtractorWork.java
```diff
@@ -183,6 +185,9 @@ protected String blobsToText(List<Blob> blobs, String docId) {
                 if (blob == null) {
                     continue;
                 }
+                if (StringUtils.isNotEmpty(mimeType) && StringUtils.isNotEmpty(blob.getMimeType())) {
```
efge: Don't you mean `StringUtils.isEmpty(mimeType)` ?

ron1:
Review feedback has been incorporated into commit https://github.com/nuxeo/nuxeo/pull/305/commits/9e3280dc9f208b82aba99a9b442681773a7fad86.

efge:
Thanks, pushed with squash and slight cleanups as bffa2b3dba3d98df1ccbc9854bf1a03837a7e8e0.

PR #331, opened 2016-11-07
task-NXP-19790-nuxeo-server-it

ataillefer:


efge:
👍 seems ok but it's hard to review as it's so big a change

jcarsique:
:+1:

PR #332, opened 2016-11-07
NXP-20641: allow mim-type check to be disabled on FileManager.createDocumentFromBlob

Gabez0r:
Private method checkMimeType retrieves the mime-type based on the file's name, and if that fails, it relies on the blob's mime-type, but only if it's normalized. Otherwise, it tries to read the blob, and fails if the blob as no stream available. This happens for a native Google Drive document. I propose a flag to to disable the checkMimeType. Default behavior is not changed.

### DIFF
--- a/nuxeo-services/nuxeo-platform-filemanager-core/src/test/java/org/nuxeo/ecm/platform/filemanager/TestFileManagerService.java
+++ b/nuxeo-services/nuxeo-platform-filemanager-core/src/test/java/org/nuxeo/ecm/platform/filemanager/TestFileManagerService.java
@@ -45,6 +45,7 @@ import org.nuxeo.ecm.core.api.Blobs;
 import org.nuxeo.ecm.core.api.CoreSession;
 import org.nuxeo.ecm.core.api.DocumentModel;
 import org.nuxeo.ecm.core.api.DocumentRef;
+import org.nuxeo.ecm.core.blob.binary.BinaryBlob;
 import org.nuxeo.ecm.core.test.CoreFeature;
 import org.nuxeo.ecm.core.test.annotations.Granularity;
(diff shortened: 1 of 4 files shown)


efge:
👍

nuxeo-services/nuxeo-platform-filemanager-core/src/test/java/org/nuxeo/ecm/platform/filemanager/TestFileManagerService.java
```diff
@@ -108,6 +108,14 @@ public void testDefaultCreateFromBlob() throws Exception {
...
         assertEquals("hello.doc", doc.getProperty("file", "filename"));
         assertNotNull(doc.getProperty("file", "content"));
+
+        // let's make the same test but this time without mime-type checking
+        doc = service.createDocumentFromBlob(coreSession, input, workspace.getPathAsString(), true,
+            "test-data/hello2.doc", true);
+        assertNotNull(doc);
+        assertEquals("hello2.doc", doc.getProperty("dublincore", "title"));
+        assertEquals("hello2.doc", doc.getProperty("file", "filename"));
+        assertNotNull(doc.getProperty("file", "content"));
```
troger: I'm not sure to understand what is tested there... Wether `noMimeTypeCheck` is `true` or `false`, it changes nothing?
Shouldn't you check the mime-type or something like that?

Gabez0r: The main goal of this tests is to show that, for a blob with a mime-type that matches the file's name, using the mime-type check on or off should not make a difference. I will improve the test and add another one to make this more clear.

PR #364, opened 2016-11-21
feature-NXP-21117-add-pdf-utils-to-distribution

mnixo:
Includes the nuxeo-platform-pdf-utils addon in the nuxeo server distribution.


atchertchian:
:+1:

PR #391, opened 2016-11-30
NXP-13846: Release script must not replace version in all files

pgmillon:


scripts/release.py
```diff
@@ -651,7 +651,11 @@ def prepare(self, dodeploy=False, doperform=False, dryrun=False, upgrade_only=Fa
                 # Maintenance branches are kept, so update their versions
                 log("\n[INFO] Maintenance branch...")
                 msg_commit = "Update %s to %s" % (self.tag, self.maintenance_version)
-                self.update_versions(self.tag, self.maintenance_version)
+                self.repo.git_recurse("reset HEAD~", with_optionals=True)
+                self.update_versions(self.snapshot, self.maintenance_version)
+                for other_version in self.other_versions:
+                    if len(other_version) > 0:
+                        self.update_versions(other_version[0], other_version[2])
```
atchertchian: do you think that would it be possible to write non-regression tests for this bugfix? thanks

scripts/release.py
```diff
@@ -651,7 +651,11 @@ def prepare(self, dodeploy=False, doperform=False, dryrun=False, upgrade_only=Fa
                 # Maintenance branches are kept, so update their versions
                 log("\n[INFO] Maintenance branch...")
                 msg_commit = "Update %s to %s" % (self.tag, self.maintenance_version)
-                self.update_versions(self.tag, self.maintenance_version)
+                self.repo.git_recurse("reset HEAD~", with_optionals=True)
```
jcarsique: This removes a commit.
=> `revert -n HEAD~`

scripts/release.py
jcarsique: Nice catch! but it will be `self.update_versions(other_version[0], other_version[1])`
The other versions stay aligned on a release (`other_version[1]`) on the maintenance branch.

pgmillon:
Updated with ❤️

jcarsique:
:+1: without 2e72f1781b61ceb0311689e847eaa65b4a8a55a0

pgmillon:
I rewrote the commits with cherry picks from #393 + added some asserts to check that the 8.10 version in the layouts-summary-contrib.xml file is not modified by the release (that's the issue we address here).

jcarsique:
For information, the follow-up is on https://jira.nuxeo.com/browse/NXP-22177
Closing this too much old PR.

PR #394, opened 2016-12-01
NXP-21068 Enable to go to parent with breadcrumb button

egiuly-nuxeo:
https://jira.nuxeo.com/browse/NXP-21068


egiuly-nuxeo:
@atchertchian
I did this new commit, using StartupHelper.SERVERS_VIEW instead of hardcoding view_servers.
What do you think?

atchertchian:
It's always good to extract hardcoded strings, but we should review use cases here, seems the logics to hide the breadcrumb entirely on the servers and domains views are broken....

PR #395, opened 2016-12-05
Fix nxp 20500 validate email when editing user

egiuly-nuxeo:
When editing a user, the email was not verified (whereas it was when creating a user).
So I added a pattern constraint on the email in the `user.xsd` schema used when editing a user.

For more information, please check the JIRA issue: https://jira.nuxeo.com/browse/NXP-20500

T&P succeeded


efge:
👍 but I think it's worth adding comments in `user.xsd` and `userinfo.xsd` to mention the fact that the same pattern is used in the other schema.

PR #403, opened 2016-12-09
NXP-19242 Change error message from invalid to empty file when upload…

egiuly-nuxeo:
…ing empty file

https://jira.nuxeo.com/browse/NXP-19242


troger:
:+1:

PR #412, opened 2016-12-15
NXP-21311: fix PictureConversion merge

efge:
https://jira.nuxeo.com/browse/NXP-21311

### DIFF
--- a/nuxeo-features/nuxeo-platform-imaging/nuxeo-platform-imaging-api/src/main/java/org/nuxeo/ecm/platform/picture/api/PictureConversion.java
+++ b/nuxeo-features/nuxeo-platform-imaging/nuxeo-platform-imaging-api/src/main/java/org/nuxeo/ecm/platform/picture/api/PictureConversion.java
@@ -22,6 +22,7 @@ package org.nuxeo.ecm.platform.picture.api;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.builder.EqualsBuilder;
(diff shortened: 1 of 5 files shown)


efge:
Indeed, current behavior is still problematic as the test override contribution does ` <pictureConversion id="Original" enabled="false" />` but this still doesn't disable it as it still inherits the fact that it's the default from the base contribution.

troger:
Default picture conversions were introduced to have required conversions so that the JSF UI can work correctly. That does not make any sense now when working on a Nuxeo Server where you want to configure precisely which conversion you need for your app to avoid the overhead of generating conversions not used.

Code has been updated to remove the `default` attribute on master. The picture conversions can now be enabled or disabled as other standard contributions.

nuxeo-features/nuxeo-platform-imaging/nuxeo-platform-imaging-api/src/main/java/org/nuxeo/ecm/platform/picture/api/PictureConversion.java
```diff
@@ -222,20 +206,49 @@ public PictureConversion clone() {
...
+        }
+        if (other.renditionVisible != null) {
+            renditionVisible = other.renditionVisible;
+        }
+    }
+
     @Override
     public String toString() {
         return String.format(
                 "PictureConversion [id=%s, description=%s, tag=%s, maxSize=%d, order=%d, chainId=%s, enabled=%s, default=%s]",
```
efge: You must also remove the `default=%s`

efge:
👍

PR #459, opened 2017-01-09
NXP-21266: allow blob uris to be used as content source

nelsonsilva:
Fix for CSP issue in PDF preview with images to be backported to 8.10


efge:
👍

kevinleturc:
👍

PR #480, opened 2017-01-17
feature-NXP-21456-relax-property-check

kevinleturc:
DO NOT MERGE, work in progress.

### DIFF
--- a/nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/model/impl/AbstractProperty.java
+++ b/nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/model/impl/AbstractProperty.java
@@ -31,8 +31,11 @@ import org.nuxeo.ecm.core.api.model.PropertyNotFoundException;
 import org.nuxeo.ecm.core.api.model.ReadOnlyPropertyException;
 import org.nuxeo.ecm.core.api.model.resolver.PropertyObjectResolver;
 import org.nuxeo.ecm.core.api.model.resolver.PropertyObjectResolverImpl;
+import org.nuxeo.ecm.core.schema.types.Field;
 import org.nuxeo.ecm.core.schema.types.Schema;
 import org.nuxeo.ecm.core.schema.types.resolver.ObjectResolver;
+
+import com.google.common.base.Strings;
 
 public abstract class AbstractProperty implements Property {
 
@@ -303,6 +306,12 @@ public abstract class AbstractProperty implements Property {
 
     @Override
     public void setValue(Object value) throws PropertyException {
+        // 1. check if it's a deprecated property with a fallback property
+        Field field = getField();
+        if (field.isDeprecated() && !Strings.isNullOrEmpty(field.getFallbackXpath())) {
+            setValue('/' + field.getFallbackXpath(), value);
+            return;
+        }
         // 1. check the read only flag
         if (isReadOnly()) {
             throw new ReadOnlyPropertyException(getPath());
@@ -343,7 +352,9 @@ public abstract class AbstractProperty implements Property {
 
     @Override
     public Serializable getValue() throws PropertyException {
-        if (isPhantom() || isRemoved()) {
+        if (!(this instanceof DocumentPart) && getField().isDeprecated() && !Strings.isNullOrEmpty(getField().getFallbackXpath())) {
+            return getValue('/' + getField().getFallbackXpath());
+        } else if (isPhantom() || isRemoved()) {
             return getDefaultValue();
         }
         return internalGetValue();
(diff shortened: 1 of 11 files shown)


kevinleturc:
First of all, I choose to put the checks in AbstractProperty, but not very sure it's the right place.
There's still missing the check to not create removed property in VCS, in DBS a null property should not be persisted.

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/model/impl/AbstractProperty.java
```diff
@@ -303,6 +306,12 @@ protected void setIsMoved() {
 
     @Override
     public void setValue(Object value) throws PropertyException {
+        // 1. check if it's a deprecated property with a fallback property
+        Field field = getField();
+        if (field.isDeprecated() && !Strings.isNullOrEmpty(field.getFallbackXpath())) {
+            setValue('/' + field.getFallbackXpath(), value);
+            return;
+        }
```
kevinleturc: The three checks here don't fully satisfy me.
First in some case we need to check if current instance is not a DocumentPart, because it doesn't implement getField() method.
Maybe a refactoring of AbstractProperty to rely on a common method to do these logic is a good way. Waiting to validate this POC before spending more time on it.

nuxeo-core/nuxeo-core-schema/src/main/java/org/nuxeo/ecm/core/schema/SchemaManagerImpl.java
```diff
@@ -280,13 +283,57 @@ protected void recompute() {
...
+                    }
+                    if ("scalar2scalar".equals(fieldName)) {
+                        field.setFallbackXpath("scalarfallback");
+                    } else if ("scalar2complex".equals(fieldName)) {
+                        field.setFallbackXpath("complexfallback/scalar");
+                    } else if ("complex".equals(fieldName)) {
+                        field.setFallbackXpath("complexfallback");
+                    } else if ("filename".equals(fieldName)) {
+                        field.setFallbackXpath("content/name");
+                    }
```
kevinleturc: Here we can't use a real xpath (naming to change), because we can't fallback on a property which doesn't belong to the same schema.
I assume that for this first version we won't handle schema deprecation (not the scope) and not handle fallback in a different schema.

nuxeo-core/nuxeo-core-schema/src/main/java/org/nuxeo/ecm/core/schema/SchemaManagerImpl.java
```diff
@@ -280,13 +283,57 @@ protected void recompute() {
...
+            Schema commonSchema = schemas.get("common");
+            commonSchema.addField("size", RemovedType.INSTANCE, null, Field.DEPRECATED, null);
+            // Removed property with fallback
+            Schema fileSchema = schemas.get("file");
+            fileSchema.addField("filename", RemovedType.INSTANCE, null, Field.DEPRECATED, null);
+            // Deprecated property with fallback
+
+        } else if (step == 2) {
+            // Second step, declare them as a list
+            List<String> properties = Arrays.asList("common:size", "file:filename", "deprecated:scalar2scalar", "deprecated:scalar2complex", "deprecated:complex");
```
kevinleturc: Waiting to validate this POC to put this declaration in a Descriptor.

nuxeo-core/nuxeo-core-api/src/test/java/org/nuxeo/ecm/core/api/impl/TestDocumentModel.java
```diff
@@ -0,0 +1,173 @@
...
+        doc.setPropertyValue("file:filename", "test filename");
+        assertEquals("test filename", doc.getPropertyValue("file:content/name"));
+    }
+
+    @Test
+    public void testSetRemovedScalarPropertiesWithFallbackOnBlob() {
+        DocumentModel doc = new DocumentModelImpl("/", "doc", "File");
+        doc.setProperties("file", Collections.singletonMap("filename", "test filename"));
+        assertEquals("test filename", ((Blob) doc.getProperties("file").get("content")).getFilename());
+    }
```
kevinleturc: Tests on blobs failed because we don't have any blob under _content_. 
- Leave the exception happening, as it could happened only in cases such as setting filename before the blob, so even if we handle it, filename will be erased by blob filename.
```
org.nuxeo.ecm.core.api.NuxeoException: Trying to access a member of a null object: name
    at org.nuxeo.ecm.core.api.model.impl.primitives.BlobProperty.getMemberValue(BlobProperty.java:187)
    at org.nuxeo.ecm.core.api.model.impl.primitives.BlobProperty$ScalarMemberProperty.internalGetValue(BlobProperty.java:234)
    at org.nuxeo.ecm.core.api.model.impl.AbstractProperty.setValue(AbstractProperty.java:322)
    at org.nuxeo.ecm.core.api.model.impl.AbstractProperty.setValue(AbstractProperty.java:340)
```
- Create a StringBlob at the right moment to set filename

PR #498, opened 2017-01-25
fix-NXP-21364-pdf-preview-shows-wrong-characters-7.10

mnixo:
Context: https://jira.nuxeo.com/browse/NXP-21364

When detecting the encoding of the provided html file in `UTF8CharsetConverter`, `CharsetDetector` generates a `CharsetMatch` with “Big5” encoding (with a value of confidence of 10).
This sample html file contains an inline image in base64. When this image is removed, the encoding of the html file is correctly evaluated as “UTF-8”.
I tried to create a new test case based on the sample html file, using the inline image block, which resulted also in a “UTF-8” match. Apparently it’s the match between that specific image block and the rest of the file that causes the evaluation to result in a “Big5” match (making it difficult to find a failing html resource that is not this particular one).

(T&P did not create the PR: https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-mnixo-7.10/1/)

### DIFF
--- a/nuxeo-services/nuxeo-platform-convert/src/main/java/org/nuxeo/ecm/platform/convert/plugins/UTF8CharsetConverter.java
+++ b/nuxeo-services/nuxeo-platform-convert/src/main/java/org/nuxeo/ecm/platform/convert/plugins/UTF8CharsetConverter.java
@@ -36,6 +36,9 @@ import org.nuxeo.ecm.core.convert.extension.ConverterDescriptor;
 
 import com.ibm.icu.text.CharsetDetector;
 import com.ibm.icu.text.CharsetMatch;
+import org.nuxeo.ecm.core.convert.plugins.text.extractors.Html2TextConverter;
+
+import static javax.ws.rs.core.MediaType.TEXT_HTML;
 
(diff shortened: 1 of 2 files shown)


nuxeo-services/nuxeo-platform-convert/src/main/java/org/nuxeo/ecm/platform/convert/plugins/UTF8CharsetConverter.java
```diff
@@ -70,7 +73,15 @@ protected Blob convert(Blob blob) throws IOException, ConversionException {
             return blob;
         }
         if (StringUtils.isEmpty(encoding)) {
-            try (InputStream in = blob.getStream()) {
+            Blob unencodedBlob = blob;
+            if (mimetype.equals(TEXT_HTML)) {
+                // convert to text to strip the html of any inline base64 resources
```
efge: The comment should explain *why* we do this, i.e., `to avoid mis-detecting the encoding`.

nelsonsilva:
:+1:

efge:
👍

PR #515, opened 2017-02-06
NXP-21580 skip if session is null

andreinechaev:


efge:
👎 please include a unit test.

PR #539, opened 2017-02-13
NXP-21559: Add user check when assigning a permission

mincong-h:
https://jira.nuxeo.com/browse/NXP-21559

### DIFF
--- a/nuxeo-features/nuxeo-automation/nuxeo-automation-core/pom.xml
+++ b/nuxeo-features/nuxeo-automation/nuxeo-automation-core/pom.xml
@@ -115,14 +115,17 @@
     <dependency>
       <groupId>org.nuxeo.ecm.core</groupId>
       <artifactId>nuxeo-core-test</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.nuxeo.runtime</groupId>
       <artifactId>nuxeo-runtime-test</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.assertj</groupId>
--- a/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/AddPermission.java
+++ b/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/AddPermission.java
@@ -113,11 +113,17 @@ public class AddPermission {
             throw new IllegalParameterException("'end' parameter must be set when adding a permission for an 'email'");
         }
 
-        String username = user;
-        if (username == null) {
+        String username;
+        if (user == null) {
+            // share a document with someone not registered in Nuxeo, by using only an email
             username = NuxeoPrincipal.computeTransientUsername(email);
-        } else if (Framework.getService(UserManager.class).getUserModel(username) == null) {
-            throw new NuxeoException("Username '" + username + "' does not exist. Please provide a valid username.");
+        } else {
+            username = user;
+            UserManager userManager = Framework.getService(UserManager.class);
+            if (userManager.getUserModel(username) == null && userManager.getGroup(username) == null) {
+                String errorMsg = "User or group name '" + username + "' does not exist. Please provide a valid name.";
+                throw new NuxeoException(errorMsg);
+            }
         }
 
         ACP acp = doc.getACP() != null ? doc.getACP() : new ACPImpl();
--- a/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/AddPermission.java
+++ b/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/AddPermission.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright 2015 Nuxeo SA (http://nuxeo.com/) and others.
+ * (C) Copyright 2015-2017 Nuxeo (http://nuxeo.com/) and others.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/AddPermission.java
+++ b/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/AddPermission.java
@@ -16,6 +16,7 @@
  * Contributors:
  *     dmetzler
  *     Vladimir Pasquier <vpasquier@nuxeo.com>
+ *     Mincong Huang <mhuang@nuxeo.com>
  */
 package org.nuxeo.ecm.automation.core.operations.document;
 
@@ -33,12 +34,15 @@ import org.nuxeo.ecm.automation.core.collectors.DocumentModelCollector;
 import org.nuxeo.ecm.core.api.CoreSession;
 import org.nuxeo.ecm.core.api.DocumentModel;
 import org.nuxeo.ecm.core.api.DocumentRef;
+import org.nuxeo.ecm.core.api.NuxeoException;
 import org.nuxeo.ecm.core.api.NuxeoPrincipal;
 import org.nuxeo.ecm.core.api.security.ACE;
 import org.nuxeo.ecm.core.api.security.ACL;
 import org.nuxeo.ecm.core.api.security.ACP;
 import org.nuxeo.ecm.core.api.security.impl.ACPImpl;
+import org.nuxeo.ecm.platform.usermanager.UserManager;
 import org.nuxeo.ecm.webengine.model.exceptions.IllegalParameterException;
+import org.nuxeo.runtime.api.Framework;
 
 /**
  * Operation that adds a permission to a given ACL for a given user.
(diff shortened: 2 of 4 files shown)


nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/AddPermission.java
```diff
@@ -136,4 +140,12 @@ protected void addPermission(DocumentModel doc) {
         }
     }
 
+    /**
+     * @since 8.10
+     */
+    private boolean isUserFound(String username) {
```
efge: There's only two lines in the method and it's only used once, so you can just inline the code at the caller site.

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/security/PermissionAutomationTest.java
```diff
@@ -78,6 +82,15 @@
...
     AutomationService automationService;
 
+    @Inject
+    private UserManager userManager;
+
+    /**
+     * Transient user for test. It is created before each test and deleted after each test. Do NOT manipulate this
+     * object during the test.
+     */
+    private DocumentModel _user;
```
efge: Just call it `testUser`, no need for an initial `_`. And actually in `@After` you only need the user id so just store that, `testUserId`, not the whole object.

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/security/PermissionAutomationTest.java
```diff
@@ -88,14 +101,26 @@ public void initRepo() throws Exception {
...
+     * @throws DirectoryException if user does not exist
+     */
+    @After
+    public void deleteUser() throws DirectoryException {
+        userManager.deleteUser(_user.getId());
     }
 
     @Test
-    public void canAddPermission() throws OperationException {
+    public void canAddPermissionForExistentUser() throws OperationException {
```
efge: _existent_ -> _existing_ everywhere

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/AddPermission.java
```diff
@@ -110,8 +114,8 @@ protected void addPermission(DocumentModel doc) {
         }
 
         String username = user;
-        if (username == null) {
-            username = NuxeoPrincipal.computeTransientUsername(email);
+        if (!isUserFound(username)) {
```
troger: What if `user` is `null` but the `email` is not? This operation allows creating transient permissions for a user given only an `email`.
Seems we are missing a test here... (my fault, will add one after).

You should do your check only if  `user` is != `null`, otherwise keep the existing code creating a TransientUsername.

mincong-h: I don't agree that we should create a transient username when `user == null && email != null`, because this is the objective of this ticket: deny creating permission if user is not found. Instead, maybe we can do

```java
if (!isUserNameFound(user) || !isUserEmailFound(email)) { ... }
```

Would it be a good idea?

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/AddPermission.java
mincong-h: I did this because I found the method name `isUserFound` more semantic, further more, it allows to keep the version since which the check is added. But based on your feedback, adding a version `@since` should be unnecessary ...

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/security/PermissionAutomationTest.java
mincong-h: Should I keep the comment?

mincong-h:
> BTW, seems like this PR was not created using a T&P on QA, it should have :)

My bad!

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/AddPermission.java
troger: Agree or not, the point is that we can share a document with someone not registered in Nuxeo, by using only an email :)

See https://doc.nuxeo.com/userdoc/managing-permissions/#granting-permissions-to-external-users-instant-share- for the functional use case.

mincong-h: Thanks for sharing this helpful link. I'll update my code.

mincong-h:
PR updated and please review again. T&P will be launched if the modification is good.

efge:
👍

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/security/PermissionAutomationTest.java
```diff
@@ -87,14 +101,26 @@ public void initRepo() throws Exception {
 
         when(administratorGroupsProvider.getAdministratorsGroups()).thenReturn(
                 Collections.singletonList("administrators"));
+
+        DocumentModel testUser = userManager.getBareUserModel();
```
troger: You could have mocked the `userManager.getUserModel` method to avoid requiring the `PlatformFeature` creating and deleting a user.

mincong-h: In our case, what are the benefices to mock instead using real test preparation? We can have less dependencies when using mock, but `PlatformService` is in the `test` scope, so it should be alright, isn't?

Obviously, only the User ID `testUserId` is stored here, but it doesn't mean it's the only thing we use: A real user need to be created before each test. If the target user `existingUser ` does not exist, all the tests based on this user will fail. Such failure is raised by the user-existence-check in `AddPermission.java`, added in this ticket.

troger: A faster test with less stuff to deploy + one less maven dependency.

There is no real need to create a user here, you just assume "existingUser" is a real user that exists (+ there are already stuff mocked in the test).

mincong-h: I agree with you. Code updated: please see the commit 500ba9.

nuxeo-features/nuxeo-automation/nuxeo-automation-core/pom.xml
```diff
@@ -115,19 +115,27 @@
...
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.nuxeo.ecm.platform</groupId>
+      <artifactId>nuxeo-platform-test</artifactId>
```
troger: I don't think you need this dependency anymore now that you are using a mock?

nuxeo-jsf/nuxeo-platform-webapp-base/pom.xml
```diff
@@ -210,11 +210,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.nuxeo.ecm.platform</groupId>
```
troger: Does this remove belong to your fix?

mincong-h: My bad. I was working on 2 issues in the same time and deleted the wrong file.

mincong-h: Updated again: see commit 0ee4d9f.

nuxeo-features/nuxeo-automation/nuxeo-automation-test/src/test/java/org/nuxeo/ecm/automation/server/test/EmbeddedAutomationClientTest.java
```diff
@@ -1001,22 +1005,54 @@ public void shouldReadContentEnricher() throws IOException {
...
+               .set("username", "existingMembers")
                .set("permission", "Write")
                .set("begin", begin)
                .set("end", end)
                .execute();
         // TODO NXP-17232 to use context parameters in json payload response with automation and automation client.
         // Once NXP-17232 resolved: assertions possible to get related doc ACLs.
+
+        // Tear down
+        userManager.deleteUser(testUser.getId());
```
troger: Please, do the tear down in a `try {} finally {}` to ensure that the user is always deleted.

nuxeo-features/nuxeo-automation/nuxeo-automation-test/src/test/java/org/nuxeo/ecm/automation/server/test/EmbeddedAutomationClientTest.java
```diff
@@ -1001,22 +1005,56 @@ public void shouldReadContentEnricher() throws IOException {
...
+                   .set("end", end)
+                   .execute();
+        } catch (RemoteException e) {
+            String expectedMsg = "Failed to invoke operation: Document.AddPermission";
+            assertEquals(e.getMessage(), expectedMsg, e.getMessage());
+        }
+    }
+
+    /**
+     * @since 7.4
```
troger: 8.10?

mincong-h:
Merged by @nuxeojenkins, see commit 3683359. Thanks for reviewing my PR, @efge @troger.

PR #563, opened 2017-02-23
NXP-21447: Use a different policy for subscriber, because they are lo…

bdelbosc:
…ng running operation exponential backoff with timeout is not approriated, change to retry for ever with constant delay


bdelbosc:
tried 2 times to t&p but got random on unrelated webdriver tests

efge:
👍

kevinleturc:
👍

PR #573, opened 2017-03-02
NXP-21601: add super type name when appending document type descriptors

Gabez0r:


nuxeo-core/nuxeo-core-schema/src/main/java/org/nuxeo/ecm/core/schema/DocumentTypeDescriptor.java
```diff
@@ -130,6 +130,13 @@ public DocumentTypeDescriptor merge(DocumentTypeDescriptor other) {
             }
         }
 
+        // update supertype
+        if ((superTypeName == null || superTypeName.isEmpty()) &&
+            (other.superTypeName != null && !other.superTypeName.isEmpty())) {
+            superTypeName = other.superTypeName;
```
Gabez0r: The main idea here is to update the super type only and only if there is no supertype defined so far. I believe this would be an expectable behavior from an `append="true"`.

efge:
👍

atchertchian:
:+1:

PR #584, opened 2017-03-07
NXP-21810: bind gatling.skip on skipITs

jcarsique:
Default gatling-maven-plugin configuration which will skip the Gatling tests if `skipITs=true`.

https://jira.nuxeo.com/browse/NXP-21810


mnixo:
👍

PR #664, opened 2017-03-28
Update RedisSentinelDescriptor.java

pjenicot:
The "host" variable is not set for the class "RedisSentinelDescriptor".

The sentinel mode on Redis for Nuxeo version 6.0 can not work without it.

Can you update the code in the next hotfix ?

Thanks

### DIFF
--- a/nuxeo-core/nuxeo-core-redis/src/main/java/org/nuxeo/ecm/core/redis/RedisSentinelDescriptor.java
+++ b/nuxeo-core/nuxeo-core-redis/src/main/java/org/nuxeo/ecm/core/redis/RedisSentinelDescriptor.java
@@ -12,6 +12,7 @@ import java.util.HashSet;
 import java.util.Set;
 
 import org.nuxeo.common.xmap.annotation.XNode;
+import org.nuxeo.common.xmap.annotation.XNodeList;
 import org.nuxeo.common.xmap.annotation.XObject;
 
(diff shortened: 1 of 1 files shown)


nxmatic:
On that version, the way of contributing nuxeo for a sentinel deployment was different from the current version. You should use this format instead
...
  <extension target="org.nuxeo.ecm.core.redis"
    point="configuration">
        <sentinel>
          <master>master</master>
          <hap name="localhost" port="26379"/>
          <hap name="localhost" port="26380"/>
          <password>secret</password>
          <database>0</database>
          <timeout>2000</timeout>
        </sentinel>
  </extension>
...

Can you have a try with that way of doing ? In case of problems, I suggest you to ask for on our site https://answers.nuxeo.com/ before starting a patch.

pjenicot:
Thank you for your reply.

But in this case, it will be necessary to modify the contribution  "nuxeo / nuxeo-distribution / nuxeo-distribution-resources / src / main / resources / templates-tomcat / common-base / nxserver / config / redis-config.xml.nxftl" because the requested XML tag is "host" and not "hap"...

Can we change to be like the current version?

nxmatic:
you're right, the redis template is wrong in that version, and we should the tag hap instead of host. We're going to fix it (see NXP-22012)

nxmatic:
I've just merged the fix in 6.0 branch.

PR #755, opened 2017-04-20
NXP-22202: Ensure to not rely on nuxeo.home if not temporary

akervern:


jcarsique:
Looks weird. Please describe the usage. Which use cases do you cover?
See parameters at https://github.com/nuxeo/tools-nuxeo-ftest/blob/master/README.md#available-parameters
Wasn't it possible to transpose the standard logic based on the custom Nuxeo Tomcat on the static WAR deployment architecture based on a stock Tomcat?
Why not using the "light" Tomcat?

akervern:
Why not using light tomcat; as I answered you in the associated jira ticket: because it does not contain `catalina.sh` script.

I cover the cases you asked for:
 - using `nuxeo.home` to create the pack; then deploy/test it in a separate folder on a stock Tomcat
 - using `zip.file` extract to create the pack; then replace it by the stock Tomcat
 - default mode; same before modifications.

Have you tested it? Isn't it working?
What do you mean by "transposing the standard logic"? Isn't it what I do by overriding used variable to make them point to the stock tomcat instead?

akervern:
Not sure to understand how do you create a pack file.

jcarsique:
:+1: successfully tested

Added commits to:
- make use of `nuxeoURL`
- cleanup and format

I found that `tomcat.home` always values `${out.dir}/tomcat`, isn't it?
```
# nuxeo-ftest.xml
if env.NUXEO_HOME
    nuxeo.home=${env.NUXEO_HOME}
    no.download defined
else
    nuxeo.home=${out.dir}/tomcat

# itest.xml
if env.NUXEO_HOME
    tomcat.home=${out.dir}/tomcat
else
    tomcat.home=${nuxeo.home}

```

akervern:
Yes, you're right! Small typo!

PR #808, opened 2017-05-24
Obey the MIME-TYPE on multipart file upload if set via the X-File-Type

cocevk:
On multipart upload (/api/v1/upload/ endpoint), the X-File-Type header is not respected. 

This is a problem when trying to set the correct mime-type for svg, css, json and other files that are not recognised in the Nuxeo backend.

### DIFF
--- a/nuxeo-features/rest-api/nuxeo-rest-api-server/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/BatchUploadObject.java
+++ b/nuxeo-features/rest-api/nuxeo-rest-api-server/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/BatchUploadObject.java
@@ -169,7 +169,10 @@ public class BatchUploadObject extends AbstractResource<ResourceTypeImpl> {
                     if (!UPLOAD_TYPE_CHUNKED.equals(uploadType)) {
                         fileName = blob.getFilename();
                     }
-                    mimeType = blob.getMimeType();
+                    // Don't change the mime-type if it was forced via the X-File-Type header
+                    if (mimeType == null || mimeType.isEmpty()) {
+                        mimeType = blob.getMimeType();
+                    }
                     uploadedSize = String.valueOf(blob.getLength());
                 }
             } else {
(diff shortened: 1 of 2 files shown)


efge:
Thanks, I created [NXP-22408](https://jira.nuxeo.com/browse/NXP-22408) to track this. Would you mind changing your commit message to start with `NXP-22408: ` to follow our conventions?

Also a unit test would be nice. You could update `BatchUploadFixture` and run the test via `BatchUploadTest`.

cocevk:
Sure thing!

cocevk:
@efge I get the following error while building the rest-api: 
```Failed to read artifact descriptor for org.nuxeo.common:nuxeo-common:jar:8.10-HF08-SNAPSHOT: Could not transfer artifact org.nuxeo.common:nuxeo-common:pom:8.10-HF08-SNAPSHOT from/to hotfix-snapshots (https://maven.nuxeo.org/nexus/content/repositories/hotfix-snapshots): Not authorized , ReasonPhrase:Unauthorized.```
Can you help?

efge:
We don't distribute SNAPSHOT artifacts. You should recompile this locally then use maven with `-nsu`.

cocevk:
@efge DONE, ```BatchUploadFixture``` unit test also updated.

efge:
Thanks for introducing constants in the tests but please do so in a separate commit, so that there is a minimal commit with your changes and the associated test change.
But really just this simple commit is enough for us, don't bother with the cleanup. If you really want to though, please make the constants `public` and add an empty line between each to follow our style guide.
Thanks again.

cocevk:
@efge ok, done, really sorry for the trouble

efge:
If I comment out the `mimeType = blob.getMimeType()` you added, then the tests still pass, which means we're not testing this failure case. Could you review your test changes so that something fails if we don't add your fix, and then passes if your fix is there?

cocevk:
@efge done.

cocevk:
@efge any progress on PR?

efge:
Thanks, this has now gone through our internal QA process and I'll inform you of the commit when it's actually merged.

cocevk:
@efge great, thanks!

efge:
Committed as 644642efd93e82ddc8fd79de237ac8378f1ac717

PR #849, opened 2017-06-05
NXP-22042: support SSL with Elasticsearch Shield

funshodavid:


kevinleturc:
👍

nuxeo-features/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/main/java/org/nuxeo/elasticsearch/config/ESClientInitializationDescriptor.java
```diff
@@ -42,6 +42,12 @@
     @XNode("password")
     protected String password;
 
+    @XNode("sslKeystorePath")
+    protected String sslKeystorePath;
+
+    @XNode("sslKeystorePassword")
+    protected String sslKeystorePassword;
+
```
atchertchian: a detail, but it's a good practice to keep since annotations on xmap-annotated attributes too

PR #892, opened 2017-06-20
NXP-21430: Do not force JSP as UTF-8

akervern:


efge:
Yes, not closing tickets before push is actually done is very important. I'm reopening.

PR #1088, opened 2017-08-23
NXP-22954: avoid NumberLong when specifying 0/1/-1 as flags

efge:
Fix for https://jira.nuxeo.com/browse/NXP-22954


kevinleturc:
👍

PR #1092, opened 2017-08-23
NXP-22954: avoid NumberLong when specifying 0/1/-1 as flags (8.10)

efge:
https://jira.nuxeo.com/browse/NXP-22954 8.10 backport


kevinleturc:
👍

PR #1115, opened 2017-08-28
Fix nxp 22951 template jsf optims 7.10

atchertchian:
https://jira.nuxeo.com/browse/NXP-22951
https://jira.nuxeo.com/browse/NXBT-1860


efge:
👍

PR #1233, opened 2017-10-02
NXP-22724: unique collection position constraints on PostgreSQL

efge:
https://jira.nuxeo.com/browse/NXP-22724

Note that we don't know the speed impact of having a constraint on all of our lists on PostgreSQL, so maybe the feature should be off by default?


bdelbosc:
I think this should be on by default on new instance, the cost of the constraint is very low and the concurrent exception will help to pinpoint cases where it happens.
:+1:

kevinleturc:
👍

PR #1272, opened 2017-10-13
Feature nxp 23142 es multi client test

bdelbosc:


nuxeo-features/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/test/java/org/nuxeo/elasticsearch/test/ESTestClientFactory.java
```diff
@@ -0,0 +1,96 @@
...
+    public ESClient create(ElasticSearchEmbeddedNode node, ElasticSearchClientConfig config) {
+        // we don't use the provided config for the client
+        String clientType = System.getProperty(CLIENT_PROPERTY);
+        switch (clientType != null ? clientType : DEFAULT_CLIENT) {
+            case TRANSPORT_CLIENT:
+                return createTransportClient(node);
+            case REST_CLIENT:
+                return createRestClient(node);
+            case RANDOM_CLIENT:
+                getRandomClient();
```
kevinleturc: We shouldn't rely on a random client, even if we don't return it here.

kevinleturc:
Does the rest of platform tests will benefit of this client switch ? Maybe we should implement a feature such as `CoreFeature` which decides according to the system properties, which client to use.
Wdyt ?

nuxeo-features/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/test/resources/elasticsearch-test-contrib.xml
```diff
@@ -8,14 +8,17 @@
...
+    <elasticSearchClient class="org.nuxeo.elasticsearch.test.ESTestClientFactory" />
+
+    <!--<elasticSearchClient class="org.nuxeo.elasticsearch.client.ESTransportClientFactory">-->
+      <!--<option name="addressList">localhost:9300</option>-->
+      <!--<option name="clusterName">elasticsearch</option>-->
+      <!--<option name="clientTransportSniff">false</option>-->
+      <!--<option name="clientTransportIgnoreClusterName">false</option>-->
+      <!--<option name="clientTransportPingTimeout">5s</option>-->
+      <!--<option name="clientTransportNodesSamplerInterval">5s</option>-->
+    <!--</elasticSearchClient>-->
```
ataillefer: Comments kept on purpose?

ataillefer:
In fact, @kevinleturc @bdelbosc, I wonder if `elasticsearch-test-contrib.xml` should not simply be part of the `RepositoryElasticSearchFeature`? This would possibly:
- Make any test using this feature rely on the ES client implementation switch mechanism.
- Allow to remove all the `@LocalDeploy({"org.nuxeo.elasticsearch.core.test:elasticsearch-test-contrib.xml"` occurrences used in almost every test using `RepositoryElasticSearchFeature`.

nuxeo-features/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/test/java/org/nuxeo/elasticsearch/test/ESTestClientFactory.java
bdelbosc: will be removed

nuxeo-features/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/test/resources/elasticsearch-test-contrib.xml
bdelbosc: yes because it the way run test locally using a different configuration like against a remote ES cluster

bdelbosc:
It benefits to all tests that use the test configuration, it can also work for ftests if they are configured to use this EsTestClientFactory.
Adding the test configuration by default to the RepositoryElasticSearchFeature sounds a good idea, but out of the scope of this ticket

PR #1295, opened 2017-10-19
NXP-23347: Enabling property name nxql remapping and minor_version su…

covolution:
This is a Work in Progress, specific to MongoDB.

### DIFF
--- a/nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSDocument.java
+++ b/nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSDocument.java
@@ -198,6 +198,13 @@ public class DBSDocument extends BaseDocument<State> {
 
     public static final String APPLICATION_OCTET_STREAM = "application/octet-stream";
 
+    public static final String PROP_UID_MAJOR_VERSION = "uid:major_version";
+
+    public static final String PROP_UID_MINOR_VERSION = "uid:minor_version";
+
+    public static final String PROP_MAJOR_VERSION = "major_version";
+
+    public static final String PROP_MINOR_VERSION = "minor_version";
     /**
      * @since 9.3
      */
@@ -962,9 +969,9 @@ public class DBSDocument extends BaseDocument<State> {
     @Override
     protected String internalName(String name) {
         switch (name) {
-        case "major_version":
+        case PROP_MAJOR_VERSION:
             return KEY_MAJOR_VERSION;
-        case "minor_version":
+        case PROP_MINOR_VERSION:
             return KEY_MINOR_VERSION;
         }
         return name;
--- a/nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSSession.java
+++ b/nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSSession.java
@@ -1898,6 +1898,12 @@ public class DBSSession implements Session {
             return KEY_FULLTEXT_BINARY;
         case NXQL.ECM_ACL:
             return KEY_ACP;
+        case DBSDocument.PROP_UID_MAJOR_VERSION:
+        case DBSDocument.PROP_MAJOR_VERSION:
+            return DBSDocument.KEY_MAJOR_VERSION;
+        case DBSDocument.PROP_UID_MINOR_VERSION:
+        case DBSDocument.PROP_MINOR_VERSION:
+            return DBSDocument.KEY_MINOR_VERSION;
         case NXQL.ECM_FULLTEXT:
             throw new UnsupportedOperationException(name);
         }
--- a/nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSStateFlattener.java
+++ b/nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSStateFlattener.java
@@ -22,6 +22,7 @@ package org.nuxeo.ecm.core.storage.dbs;
 import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.KEY_PREFIX;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -36,18 +37,37 @@ import org.nuxeo.ecm.core.storage.State;
  */
 public class DBSStateFlattener {
 
-    public static Map<String, Serializable> flatten(State state) {
+    protected final Map<String, String> keyMappings;
+
+    public DBSStateFlattener() {
+        this.keyMappings = Collections.emptyMap();
+    }
+
+    public DBSStateFlattener(Map<String, String> keyMappings) {
+        this.keyMappings = keyMappings != null ? keyMappings : Collections.emptyMap();
+    }
+
+    /**
+     * Flattens with optional property key mappings.
+     * @param state state
+     * @return flattened result
+     * @since 9.3
+     */
+    public Map<String, Serializable> flatten(State state) {
         Map<String, Serializable> map = new HashMap<>();
         flatten(map, state, null);
         return map;
     }
 
-    private static void flatten(Map<String, Serializable> map, State state, String prefix) {
+    protected void flatten(Map<String, Serializable> map, State state, String prefix) {
         for (Entry<String, Serializable> en : state.entrySet()) {
             String key = en.getKey();
             Serializable value = en.getValue();
             String name;
-            if (key.startsWith(KEY_PREFIX)) {
+            String realName = keyMappings.get(key);
+            if (realName != null) {
+                name = realName;
+            } else if (key.startsWith(KEY_PREFIX)) {
                 name = DBSSession.convToNXQL(key);
                 if (name == null) {
                     // present in state but not returned to caller
--- a/nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBQueryBuilder.java
+++ b/nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBQueryBuilder.java
@@ -26,6 +26,10 @@ import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.KEY_ACL;
 import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.KEY_ACL_NAME;
 import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.KEY_ACP;
 import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.KEY_FULLTEXT_SCORE;
+import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.PROP_MAJOR_VERSION;
+import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.PROP_MINOR_VERSION;
+import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.PROP_UID_MAJOR_VERSION;
+import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.PROP_UID_MINOR_VERSION;
 import static org.nuxeo.ecm.core.storage.mongodb.MongoDBRepository.MONGODB_ID;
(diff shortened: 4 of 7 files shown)


nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSDocument.java
```diff
@@ -198,6 +198,13 @@
 
     public static final String APPLICATION_OCTET_STREAM = "application/octet-stream";
 
+    public static final String UID_MAJOR_VERSION = "uid:major_version";
+
+    public static final String UID_MINOR_VERSION = "uid:minor_version";
+
+    public static final String SUFFIX_MAJOR_VERSION = "major_version";
```
efge: No need for `SUFFIX`. But please call these 4 `PROP_something` to stress the fact that they are "properties" names from the high level representation of documents, not low-level "keys".

So use `PROP_UID_MAJOR_VERSION`, `PROP_UID_MINOR_VERSION`, `PROP_MAJOR_VERSION`, `PROP_MINOR_VERSION`.

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSSession.java
```diff
@@ -1454,8 +1454,8 @@ protected static ACP memToAcp(Serializable acpSer) {
             String idKey = idKeyHolder.getValue();
             List<String> ids = new ArrayList<>(pl.size());
             for (Map<String, Serializable> map : pl) {
-                String id = (String) map.get(idKey);
-                ids.add(id);
+                Serializable id = map.get(idKey);
+                if (id != null) ids.add(String.valueOf(id));
```
efge: We always use `{}` blocks even for very short conditions:

    if (id != null) {
        ids.add(String.valueOf(id));
    }

However I'd keep the cast `(String) id` instead of `String.valueOf` as the key should already have been turned into a `String` by lower layers, and I don't want to hide an unexpected change in behavior by doing `String.valueOf`

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSSession.java
```diff
@@ -1898,6 +1898,12 @@ public static String convToInternal(String name) {
             return KEY_FULLTEXT_BINARY;
         case NXQL.ECM_ACL:
             return KEY_ACP;
+        case DBSDocument.SUFFIX_MAJOR_VERSION:
+        case DBSDocument.UID_MAJOR_VERSION:
+            return DBSDocument.KEY_MAJOR_VERSION;
```
efge: You can `import static` the various `DBSDocument` constants to follow the style of the others.

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSStateFlattener.java
```diff
@@ -37,17 +38,31 @@
...
+    }
+
+    /**
+     * Flatten with optional property mappings
+     * @param state State
+     * @param keyMappings A map of key/values to replace
+     * @return Flattened result
+     * @since 9.3
+     */
+    public static Map<String, Serializable> flatten(State state, Map<String, String> keyMappings) {
```
efge: Please make all methods non-`static` and pass the `keyMappings` in the constructor, to avoid propagating it in all recursive calls.

    protected static final Map<String, String> keyMappings;

    public DBSStateFlattener(Map<String, String> keyMappings) {
        this.keyMappings = keyMappings;
    }

This will require changes to the few callers of course: `new DBSStateFlattener(mappings).flatten(state)`

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBQueryBuilder.java
```diff
@@ -211,6 +214,11 @@ protected void walkProjection() {
                 throw new QueryParseException("Projection not supported: " + op);
             }
             FieldInfo fieldInfo = walkReference((Reference) op);
+            if (!NXQL.ECM_UUID.equals(fieldInfo.prop) &&
+                !fieldInfo.prop.equals(fieldInfo.projectionField) &&
+                !fieldInfo.prop.contains("/")) {
```
efge: Put `fieldInfo.prop` in a local variable.

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBQueryBuilder.java
```diff
@@ -1014,6 +1022,12 @@ protected FieldInfo walkReference(String name) {
                 }
             }
             Type type = field.getType();
+
+            if (prop.contains(DBSDocument.SUFFIX_MAJOR_VERSION) || prop.contains(DBSDocument.SUFFIX_MINOR_VERSION)) {
```
efge: Please don't use `contains` but exact string equality on what's really expected.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestSQLRepositoryQuery.java
```diff
@@ -2295,6 +2293,69 @@ public void testSelectColumnsSameName() throws Exception {
...
+    /**
+     * @since 9.3
+     */
+    @Test
+    public void testSelectColumnMappings() throws Exception {
+        String query;
+        IterableQueryResult res;
+        Map<String, Serializable> map;
+
+        DocumentModel file = new DocumentModelImpl("/", "testfileForColumns", "File");
```
efge: You're probably copy/pasting from other tests, but now we prefer to not rely on the implementation and do instead:

    DocumentModel file = session.createDocumentModel("/", "testfileForColumns", "File");

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSStateFlattener.java
```diff
@@ -37,17 +38,31 @@
 public class DBSStateFlattener {
 
     public static Map<String, Serializable> flatten(State state) {
+        return flatten(state, Collections.emptyMap());
+    }
+
+    /**
+     * Flatten with optional property mappings
+     * @param state State
+     * @param keyMappings A map of key/values to replace
```
efge: Javadoc style nitpicks: First sentence should end with a `.` and should be 3rd person declarative rather than 2nd person imperative (—> Flattens). `@param` and `@return` are not full sentences so no initial capital and no final `.`.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestSQLRepositoryQuery.java
```diff
@@ -2295,6 +2293,69 @@ public void testSelectColumnsSameName() throws Exception {
...
+        res = session.queryAndFetch(query, "NXQL");
+        assertEquals(1, res.size());
+        map = res.iterator().next();
+        assertEquals("special1", map.get("dc:title"));
+
+        query = "SELECT uid:major_version FROM File WHERE dc:title = 'special1' AND ecm:isVersion = 1";
+        res = session.queryAndFetch(query, "NXQL");
+        assertEquals(1, res.size());
+        map = res.iterator().next();
+        assertEquals(1l, map.get("uid:major_version"));
```
efge: Please use `1L` instead of `1l`, it's much more readable. Personally I even tend to avoid boxing and be explicit here: `Long.valueOf(1)`

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSSession.java
covolution: I add this because a Long (major_version) was failing to cast to a String.  That was halfway through my fix so could have been a result of something else.  Let me take this out and see if it still breaks.

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSDocument.java
covolution: It would be great if there were some constants I could reuse.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestSQLRepositoryQuery.java
covolution: Something to remember for the next time.  I will leave it how it is for this test so its consistent with the many other tests in this class.

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSStateFlattener.java
```diff
@@ -36,18 +37,37 @@
  */
 public class DBSStateFlattener {
 
-    public static Map<String, Serializable> flatten(State state) {
+    private final Map<String, String> keyMappings;
```
efge: Please use `protected` and not `private` for fields. You never know when subclassing something can be useful.

 (except `log` and `serialVersionUID` which should never be shared even by subclasses)

efge:
👍

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSDocument.java
```diff
@@ -198,6 +198,13 @@
 
     public static final String APPLICATION_OCTET_STREAM = "application/octet-stream";
 
+    public static final String PROP_UID_MAJOR_VERSION = "uid:major_version";
+
+    public static final String PROP_UID_MINOR_VERSION = "uid:minor_version";
+
+    public static final String PROP_MAJOR_VERSION = "major_version";
+
+    public static final String PROP_MINOR_VERSION = "minor_version";
```
bdelbosc: we tend to use suffix instead of prefix (FOO_PROP instead of PORP_FOO)

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSStateFlattener.java
```diff
@@ -36,18 +37,37 @@
...
+        this.keyMappings = Collections.emptyMap();
+    }
+
+    public DBSStateFlattener(Map<String, String> keyMappings) {
+        this.keyMappings = keyMappings != null ? keyMappings : Collections.emptyMap();
+    }
+
+    /**
+     * Flattens with optional property key mappings.
+     * @param state state
```
bdelbosc: if you have nothing to say about this param remove the line

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSStateFlattener.java
```diff
@@ -36,18 +37,37 @@
...
+     * @since 9.3
+     */
+    public Map<String, Serializable> flatten(State state) {
         Map<String, Serializable> map = new HashMap<>();
         flatten(map, state, null);
         return map;
     }
 
-    private static void flatten(Map<String, Serializable> map, State state, String prefix) {
+    private void flatten(Map<String, Serializable> map, State state, String prefix) {
```
bdelbosc: we don't use private for methods only protected

bdelbosc:
:+1:

bdelbosc:
:+1:

covolution:
I have a T&P which works except for the intermittent web-ui failures.  https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-gethin/18/console

PR #1299, opened 2017-10-19
NXP-19412 migration service

efge:
WIP for https://jira.nuxeo.com/browse/NXP-19412


nuxeo-runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -0,0 +1,286 @@
...
+
+        runnable.run(); // TODO XXX in a thread
+    }
+
+    /**
+     * Execute while setting a lock
+     */
+    protected <T> T atomic(String id, Function<KeyValueStore, T> function) {
+        KeyValueStore kv = getKeyValueStore();
+        String lock = String.valueOf(System.currentTimeMillis());
```
bdelbosc: may be rename "lock" to "value" here, could be nice to have the node name + timestamp for debug purpose

nuxeo-runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -0,0 +1,286 @@
...
+        if (step == null) {
+            // migration never started, assumed at first state
+            state = descr.states.keySet().iterator().next(); // LinkedHashMap guarantees order
+            return new MigrationStatus(state);
+        }
+        long startTime = Long.parseLong(kv.getString(id + START_TIME));
+        long pingTime = Long.parseLong(kv.getString(id + PING_TIME));
+        String progressMessage = kv.getString(id + PROGRESS_MESSAGE);
+        long progressNum = Long.parseLong(kv.getString(id + PROGRESS_NUM));
+        long progressTotal = Long.parseLong(kv.getString(id + PROGRESS_TOTAL));
```
bdelbosc: to be consistent the read need to be done atomicly with the same write lock

bdelbosc:
:+1:

nuxeo-runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -0,0 +1,286 @@
...
+        }
+    }
+
+    protected void atomic(String id, Consumer<KeyValueStore> consumer) {
+        atomic(id, kv -> {
+            consumer.accept(kv);
+            return null;
+        });
+    }
+
```
funshodavid: Would it be worth exposing these atomic methods in the `LockHelper` class ?

funshodavid:
:+1:

nuxeo-runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
efge: They're small enough that we don't need to factor them for now, but yes eventually something similar will be helpful. Although here there's a convention on related key names which isn't needed in the general case.

nuxeo-runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
efge: Yes but I don't want the read path to need locks, as it can be used in polling. The idea is that if the "state" key is present then that's the status, otherwise the "step" and other keys determine the running status, in which case atomicity is not really important. I'll try to rework this to be clearer. The lock is here just to allow atomic transition between "state set and not step" and the opposite.

kevinleturc:
👍

kevinleturc:
👍

funshodavid:
:+1:

funshodavid:
:+1:

nuxeo-features/nuxeo-admin-center/nuxeo-admin-center-jsf/src/main/java/org/nuxeo/ecm/admin/MigrationAdminBean.java
```diff
@@ -0,0 +1,85 @@
...
+                    }
+                }
+                // sort steps by id
+                Collections.sort(steps, (a, b) -> a.getId().compareTo(b.getId()));
+                migrationInfo.put("steps", steps);
+            }
+            migrationInfos.add(migrationInfo);
+        });
+        // sort migrationInfos by id
+        Collections.sort(migrationInfos, (a, b) -> ((String) a.get("id")).compareTo((String) b.get("id")));
```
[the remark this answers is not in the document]
efge: Agreed it's a better pattern.
However, as in practice these lists will have around 3 elements max 😀, I'll change this only if I need to do other changes on the PR.

efge:
PR succeeded: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/272/

PR #1451, opened 2017-11-21
NXP-23706 adding leading wildcards to H2 for fulltext

andreinechaev:
leading wildcards for H2 for fulltext


efge:
See comments in https://jira.nuxeo.com/browse/NXP-23706

PR #1454, opened 2017-11-21
NXP-23491: add indexes to speed up MongoDB audit query

efge:
https://jira.nuxeo.com/browse/NXP-23491

### DIFF
--- a/nuxeo-features/nuxeo-platform-audit/nuxeo-platform-audit-mongodb/src/main/java/org/nuxeo/mongodb/audit/MongoDBAuditBackend.java
+++ b/nuxeo-features/nuxeo-platform-audit/nuxeo-platform-audit-mongodb/src/main/java/org/nuxeo/mongodb/audit/MongoDBAuditBackend.java
@@ -20,6 +20,7 @@ package org.nuxeo.mongodb.audit;
 
 import static org.nuxeo.ecm.platform.audit.api.BuiltinLogEntryData.LOG_CATEGORY;
 import static org.nuxeo.ecm.platform.audit.api.BuiltinLogEntryData.LOG_DOC_PATH;
+import static org.nuxeo.ecm.platform.audit.api.BuiltinLogEntryData.LOG_DOC_UUID;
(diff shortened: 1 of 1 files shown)


kevinleturc:
👍

andreinechaev:
👍

nuxeo-features/nuxeo-platform-audit/nuxeo-platform-audit-mongodb/src/main/java/org/nuxeo/mongodb/audit/MongoDBAuditBackend.java
```diff
@@ -122,7 +124,9 @@ public void onApplicationStarted() {
         MongoDBConnectionService mongoService = Framework.getService(MongoDBConnectionService.class);
         MongoDatabase database = mongoService.getDatabase(AUDIT_DATABASE_ID);
         collection = database.getCollection(collName);
-        // TODO migration ?
+        collection.createIndex(Indexes.ascending(LOG_DOC_UUID)); // query by doc id
```
egiuly-nuxeo: Question @efge please:
why not LOG_PRINCIPAL_NAME or LOG_DOC_LIFE_CYCLE too for instance?

efge: I added doc id, date range, and event type because that's what most of the queries I saw were doing. If you think we have other queries that only filter on principal name or lifecycle state, but have no associated filter on doc id, date range or event type, then I'm interested.

efge: The idea is that pre-filtering by doc id or date range or event type is usually enough to get efficient results even if we do filter by principal or lifecycle *in addition*.

egiuly-nuxeo: Ok I see, makes sense, thanks

PR #1464, opened 2017-11-22
NXP-23412: Fixes duplicate labels introduced by PR #1460

rsertelon:
> I'm creating this PR manually as suggested by @kevinleturc

This pull request aims at fixing synchronization with crowdin for translations since PR #1460 introduced duplicated text which is forbidden and couldn't be foreseen with standard test and push process.

Resolution is to rename the older duplicate to a name that matches more what it should represent.


efge:
👍

kevinleturc:
👍

PR #1494, opened 2017-11-30
 NXP-23722: Prevent directory sessions from modifying their parameters

egiuly-nuxeo:
Prevent directory sessions from modifying their parameters
in order to be able to use session.createEntry(Collections.singletonMap(k,v))

This code will be used in https://github.com/nuxeo/nuxeo-audit-storage/pull/2 and https://github.com/nuxeo/nuxeo-signature/pull/7

### DIFF
--- a/nuxeo-services/nuxeo-platform-directory/nuxeo-platform-directory-ldap/src/main/java/org/nuxeo/ecm/directory/ldap/LDAPSession.java
+++ b/nuxeo-services/nuxeo-platform-directory/nuxeo-platform-directory-ldap/src/main/java/org/nuxeo/ecm/directory/ldap/LDAPSession.java
@@ -145,6 +145,9 @@ public class LDAPSession extends BaseSession {
 
     @Override
     protected DocumentModel createEntryWithoutReferences(Map<String, Object> fieldMap) {
+        // Make a copy of fieldMap to avoid modifying it
+        fieldMap = new HashMap<>(fieldMap);
+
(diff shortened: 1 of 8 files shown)


nuxeo-services/nuxeo-platform-directory/nuxeo-platform-directory-ldap/src/main/java/org/nuxeo/ecm/directory/ldap/LDAPSession.java
```diff
@@ -145,10 +145,13 @@ public DirContext getContext() {
 
     @Override
     protected DocumentModel createEntryWithoutReferences(Map<String, Object> fieldMap) {
+        // Make a copy of fieldMap to avoid modifying it
+        Map<String, Object> fieldMapCopy = new HashMap<>(fieldMap);
```
efge: You can simply say `fieldMap = new HashMap<>(fieldMap)`. This just puts the result into `fieldMap` instead of creating a new local variable `fieldMapCopy` (we don't have a code style prohibition on never changing passed-in parameters) so the diff will also be minimized.

egiuly-nuxeo: ok 👍

efge:
Or at least squash the first two commits, the third one can remain by itself (but change the commit message).

efge:
👍

funshodavid:
:+1:

PR #1497, opened 2017-12-01
NXP-23788: optimize Read ACL update on big volume

efge:


nuxeo-core/nuxeo-core-storage-mem/src/main/java/org/nuxeo/ecm/core/storage/mem/MemRepository.java
```diff
@@ -266,12 +282,21 @@ public boolean hasChild(String parentId, String name, Set<String> ignored) {
...
+
+    @Override
+    public Stream<State> getDescendants(String rootId, Set<String> keys, int limit) {
         if (log.isTraceEnabled()) {
             log.trace("Mem: QUERY " + KEY_ANCESTOR_IDS + " = " + rootId);
         }
-        return states.values() //
-                     .stream()
-                     .filter(state -> hasAncestor(state, rootId));
+        Stream<State> stream = states.values() //
```
egiuly-nuxeo: extra ` // ` I guess

efge: It's here as a common hack to force some IDEs (Eclipse) to indent this correctly (aligning each call in a fluent style on the same column)

PR #1505, opened 2017-12-05
NXP-23395: fix totalSize computation in Query SELECTing from lists

efge:
https://jira.nuxeo.com/browse/NXP-23395


nuxeo-core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/sql/model/WhereClause.java
```diff
@@ -28,13 +28,13 @@
 
     private static final long serialVersionUID = -6192874148042567401L;
 
-    public final Predicate predicate;
+    public final Expression predicate;
```
covolution: It would be clearer if the variable is called `expression`.

covolution:
:+1:

nuxeo-core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/sql/model/WhereClause.java
efge: Agreed but I wanted to minimize the diff

PR #1512, opened 2017-12-06
NXP-23852: Make DocumentModelComparator ignore case for orderBy

egiuly-nuxeo:
JIRA: https://jira.nuxeo.com/browse/NXP-23852

T&P running: https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-egiuly/122/

### DIFF
--- a/nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/DocumentModelComparator.java
+++ b/nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/DocumentModelComparator.java
@@ -106,7 +106,7 @@ public class DocumentModelComparator implements Sorter {
             DataModel d2 = doc2.getDataModel(schemaName);
             for (Entry<String, String> e : orderBy.entrySet()) {
                 final String fieldName = e.getKey();
-                final boolean asc = ORDER_ASC.equals(e.getValue());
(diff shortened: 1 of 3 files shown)


nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/DocumentModelComparator.java
```diff
@@ -106,7 +106,7 @@ public int compare(DocumentModel doc1, DocumentModel doc2) {
             DataModel d2 = doc2.getDataModel(schemaName);
             for (Entry<String, String> e : orderBy.entrySet()) {
                 final String fieldName = e.getKey();
-                final boolean asc = ORDER_ASC.equals(e.getValue());
+                final boolean asc = ORDER_ASC.equalsIgnoreCase(e.getValue());
```
efge: For APIs I don't see the point in ignoring case. The developer should use whatever case is specified in the API (here uppercase I guess, or better, the constants — maybe we could add a `ORDER_DESC` to be clear).

So -1 on this change as it is. I'll comment on the ticket.

egiuly-nuxeo: ok thanks

PR #1702, opened 2018-02-08
Fix NXP-24343 servlet tx

efge:


pom.xml
```diff
@@ -1943,14 +1944,13 @@
       <dependency>
         <groupId>javax.servlet</groupId>
         <artifactId>javax.servlet-api</artifactId>
-        <version>3.0.1</version>
-        <scope>provided</scope>
```
kevinleturc: Why do we need to remove `provided` scope ?

efge: When it's `provided` it's not seen transitively, so I would have to add the `servlet-api` dependency in all modules that depend on something that itself depends on `servlet-api`. There are a lots of them. So instead we keep it normal for the build, and it's only in `nuxeo-distribution` that it's declared `provided` to not be included in the final assemblies.

This changed because I now exclude dependencies on an older version servlet-api coming from Jetty that we don't want to actually depend on.

ataillefer:
:+1:

PR #1707, opened 2018-02-09
NXP-24309: fix assert in zip to html convert test.

richardsd:
It seems that the zip order is not deterministic. I think it makes sense to keep testing the conversion despite the order changing, and validating this. Feedback would be welcome, as this was breaking too many builds. What do you think? I'll go through the T&P after discussing it.

### DIFF
--- a/nuxeo-services/nuxeo-platform-convert/src/test/java/org/nuxeo/ecm/platform/convert/tests/TestZip2HtmlConverter.java
+++ b/nuxeo-services/nuxeo-platform-convert/src/test/java/org/nuxeo/ecm/platform/convert/tests/TestZip2HtmlConverter.java
@@ -19,14 +19,15 @@
 
 package org.nuxeo.ecm.platform.convert.tests;
 
-import org.junit.Ignore;
 import org.junit.Test;
 import org.nuxeo.ecm.core.api.Blob;
 import org.nuxeo.ecm.core.api.blobholder.BlobHolder;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -34,7 +35,6 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
-@Ignore("NXP-24309")
 public class TestZip2HtmlConverter extends BaseConverterTest {
 
     @Test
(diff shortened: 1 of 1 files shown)


efge:
I don't think it's zip order that's not deterministic but rather the list of the filesystem directory into which the zip is unzipped. ZipCachableBlobHolder.load -> SimpleCachableBlobHolder.addDirectoryToList -> directory.listFiles() has no deterministic order.

nuxeo-services/nuxeo-platform-convert/src/test/java/org/nuxeo/ecm/platform/convert/tests/TestZip2HtmlConverter.java
```diff
@@ -60,14 +60,20 @@ public void testConvertToMimeType(String mimeType) throws IOException {
...
 
-        content = DocumentUTUtils.readContent(blobs.get(1).getFile());
-        assertTrue(content, content.contains("Hello from a xml <b>document</b>"));
-
-        content = DocumentUTUtils.readContent(blobs.get(2).getFile());
-        assertTrue(content, content.contains("Hello from a text document"));
+        for (int i = 1; i < blobs.size(); i++) {
+            File file = blobs.get(i).getFile();
+            content = DocumentUTUtils.readContent(file);
+            assertTrue(content, content.contains(zipContent.get(file.getName())));
```
efge: Put `file.getName()` in a local variable, and add the iteration number and the filename in the failure message: `assertTrue("file " + i + " does not contain " + filename + ": " + content, ...)`

nuxeo-services/nuxeo-platform-convert/src/test/java/org/nuxeo/ecm/platform/convert/tests/TestZip2HtmlConverter.java
```diff
@@ -60,14 +60,20 @@ public void testConvertToMimeType(String mimeType) throws IOException {
...
+            zipContent.remove(file.getName());
+        }
+        assertEquals(zipContent.size(), 0, zipContent.size());
```
kevinleturc: Could you replace it by an `assertTrue` with a call on `isEmpty` ? Message could be something about expected files not found.

efge: Yes like `assertTrue("Unexpected remaining files: " + zipContent.keys(), zipContent.isEmpty())`

richardsd: Yes, much clear 👍 thanks

richardsd:
Pushed the requested changes to the branch (running the T&P).

efge:
👍

kevinleturc:
👍

PR #1800, opened 2018-03-04
NXP-24540: added a way to customize Quartz jobs and triggers

fmunch:
Added an optional `jobFactoryClass` attribute to the `schedule` extension
point of `SchedulerService` to allow a custom Quartz job and trigger to be
used.

Allows for example the use of `@DisallowConcurrentExecution` on the job or
`withMisfireHandlingInstructionIgnoreMisfires()` on the trigger.

See https://answers.nuxeo.com/general/q/497114904ba6475d9426ca61c9423b51/Feature-Custom-Quartz-Job-and-Trigger

### DIFF
--- a/nuxeo-core/nuxeo-core-event/src/test/java/org/nuxeo/ecm/core/scheduler/EventJobFactoryTest.java
+++ b/nuxeo-core/nuxeo-core-event/src/test/java/org/nuxeo/ecm/core/scheduler/EventJobFactoryTest.java
@@ -1,0 +1,68 @@
+/*
+ * (C) Copyright 2007-2018 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Florent Munch
+ */
+package org.nuxeo.ecm.core.scheduler;
+
+import javax.inject.Inject;
+import java.lang.reflect.Field;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.nuxeo.runtime.test.runner.Deploy;
+import org.nuxeo.runtime.test.runner.Features;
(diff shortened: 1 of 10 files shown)


efge:
Thanks for this, I opened [NXP-24540](https://jira.nuxeo.com/browse/NXP-24540) to track this proposal, would you mind updating the commit message?

We'll do a review of the PR asap.

fmunch:
Thank you. Commit message updated.

nuxeo-core/nuxeo-core-event/src/test/java/org/nuxeo/ecm/core/scheduler/EventJobFactoryTest.java
```diff
@@ -0,0 +1,76 @@
...
+        Thread.sleep(5000L);
+
+        // without DisallowConcurrentExecution the 2s job must be executed every second (even if executions overlap),
+        // meaning that the listener should be called at least four or five times
+        verify(eventListener, atLeast(4)).handleEvent("testSchedulerMultipleExecutions");
+
+        // with DisallowConcurrentExecution the 2s job cannot be executed every second and must wait for one execution
+        // to finish to run again, meaning that the listener cannot be called more than 3 times
+        verify(eventListener, atLeast(2)).handleEvent("testSchedulerSingleExecution");
+        verify(eventListener, atMost(3)).handleEvent("testSchedulerSingleExecution");
```
efge: Generally we don't like to have tests that sleep because it slows down the whole run — but for a scheduler test there's not many other ways to do testing.

In a perfect world these tests would be ok, however on slow machines we've seen that time-based tests are often not very reliable. I would prefer tests that do explicit waits and releases of the job based on `CountDownLatch`. If you see what that entails then please try to update the PR for that. Otherwise we'll push as is but will need to carefully watch for failures and if that's the case we'll have to disable the test.

fmunch:
It should be fine now.

I removed most of the previous test. The new one just checks that `jobDetail.isConcurrentExectionDisallowed()` has the expected value on both jobs. After all we just want to make sure that the factory is used and that it injects the expected jobClass. The job synchronization is up to Quartz, not Nuxeo.

efge:
👍

dmetzler:
Could you update https://github.com/nuxeo/nuxeo/blob/master/nuxeo-core/nuxeo-core-event/src/main/resources/OSGI-INF/scheduler-service.xml to reflect the additional configuration?

fmunch:
Done, is it detailed enough?

dmetzler:
👍

dmetzler:
Launching a T&P on the branch https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-dmetzler/65/

efge:
T&P in progress: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/374/

efge:
See #1868 for followup testing and merge

PR #1880, opened 2018-03-27
NXP-24540: added a way to customize Quartz jobs and triggers (9.10)

fmunch:
Backport of #1800.

Since overriding the `EventJob` class is a pain before 10.2 it would be nice to have this PR be backported to 9.10. Would this be possible?


efge:
Thanks Florent but this is a new feature / improvement and we don't backport those unless we have a specific motivated customer request through a support issue.

fmunch:
This actually comes from the work done for a platinum customer. We can live without it though.

efge:
Ok then could you please ask them to open a SUPNXP ticket referencing this PR? This will help move things forward.

PR #1968, opened 2018-04-26
NXP-24756: Fix WebDav file creation on MacOSX

kevinleturc:
I wasn't able to do a T&P, 5-6 fails with connection error to firefox or on some selenium tests.
Even if tests seem to be the same between the two last run, I wasn't able to reproduce it locally.

Here the three [tests](https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-kleturc-7.10/58/artifact/nuxeo-distribution/nuxeo-distribution-cap-selenium-tests/target/results/result-suite1.html#testresult24)
- modifyNote
- createSubGroupAdmin
- createSubGroupMembers

Does it ring a bell to someone ? My changes shouldn't affect these tests.
Note: 8.10, 9.10 and on going FT has passed T&P on first try.


efge:
Yes these are "standard" failing tests on 7.10.

efge:
👍

PR #1974, opened 2018-05-02
fix-NXP-24884-encrypted-mail.transport.password

ataillefer:
T&P OK but failed because of `bowerInstall` in Web UI: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-ataillefer/94/


jcarsique:
Is it expected that the EmailAuthenticator is not initiated with the runtime properties? The current fix is relying on the Framework to either create a new CryptoProperties, either call expandVars twice for each property. Seems strange.

nuxeo-features/nuxeo-platform-notification/nuxeo-platform-notification-core/src/main/java/org/nuxeo/ecm/platform/ec/notification/email/EmailAuthenticator.java
```diff
@@ -32,32 +32,20 @@ public EmailAuthenticator(Properties properties) {
...
-        }
-        return defaultValue(name);
-    }
-
     protected String protocolValue(String name) {
         String key = "mail." + getRequestingProtocol() + "." + name;
         return properties.getProperty(key);
     }
 
-    protected String defaultValue(String name) {
```
jcarsique: are you sure it's unused? it seems designed for being implemented by a subclass

efge:
👍

nuxeo-features/nuxeo-platform-notification/nuxeo-platform-notification-core/src/main/java/org/nuxeo/ecm/platform/ec/notification/email/EmailAuthenticator.java
ataillefer: Yes, according to a grep in the platform and released addons repositories.

ataillefer:
@jcarsique Yes it is expected, the `EmailAuthenticator` instance is initiated with the `Mail` Catalina Resource, see https://github.com/nuxeo/nuxeo/blob/master/nuxeo-distribution/nuxeo-nxr-server/src/main/resources/templates/common-base/conf/Catalina/localhost/nuxeo.xml.nxftl#L40.

In this case, when the `mail.transport.password` runtime property is encrypted, it is kept non expanded in the configuration files after preprocessing, thus the `nuxeo.xml` file ends up with:
```
mail.smtp.password="${mail.transport.password}"
```

That's why the `EmailAuthenticator` needs to expand it.

jcarsique:
@ataillefer that's what I suspected; then, why not rather fixing `org.nuxeo.ecm.platform.ec.notification.email.EmailResourceFactory` in order to initiate the right properties object?

ataillefer:
@jcarsique Yes, you're right, better let the factory expand the properties only once instead of doing it systematically in the getter.
PR updated.

jcarsique:
:+1:

PR #2055, opened 2018-06-07
NXP-25013: sanitize file name before download

tmartins:


nuxeo-core/nuxeo-core-io/src/main/java/org/nuxeo/ecm/core/io/download/DownloadServiceImpl.java
```diff
@@ -334,7 +334,7 @@ protected void handleDownload(HttpServletRequest req, HttpServletResponse resp,
                 tx = TransactionHelper.startTransaction();
             }
             String xpath = downloadBlobInfo.xpath;
-            String filename = downloadBlobInfo.filename;
+            String filename = FILENAME_SANITIZATION_REGEX.matcher(downloadBlobInfo.filename).replaceAll("");
```
efge: Why not just call `String filename = getSanitizedFilenameWithoutPath(downloadBlobInfo.filename)`?
But actually that's already called by `getDownloadUrl` which is where `filename` is used, so I'm not sure I see the point of the fix (and given there's no unit test it's really hard to know).

tmartins: getDownloadUrl is called to build the download URL. And despite this call, some download URLs are built with a jsessionid. So to be sure we never download a file whose name is not sanitized, I sanitize it here.
ok to use getSanitizedFilenameWithoutPath

PR #2058, opened 2018-06-08
NXP-25097: first bulk processor implementation

funshodavid:


nuxeo-runtime/nuxeo-runtime-stream/src/main/java/org/nuxeo/runtime/stream/BatchedProcessorTopology.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.nuxeo.lib.stream.computation.AbstractComputation;
+import org.nuxeo.lib.stream.computation.ComputationContext;
+import org.nuxeo.lib.stream.computation.Record;
+import org.nuxeo.lib.stream.computation.Topology;
+
+/**
+ * Abstract implementation of a topology processing its entries by batches.
+ *
```
bdelbosc: A computation processes records from input streams and can produces records to output streams.
A topology is a DAG of computations.
A processor run the topology of computations according to some settings.
This class assumes that there is only one computation per topology and that the processor itself handles the records that are now called entries :thinking:.
My plan was to add batching (aka windowing) at the computation level.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/SetPropertiesAction.java
```diff
@@ -0,0 +1,92 @@
...
+        return false;
+    }
+
+    @Override
+    protected void processBatch(ComputationContext context, List<String> entries) {
+        TransactionHelper.runInTransaction(() -> {
+            for (String docId : entries) {
+                BulkCommand command = BulkCommandHelper.getBulkCommandJson(commandAsBytes);
+                Map<String, Serializable> properties = command.getParam(PROPERTIES_PARAMS);
+                try (CloseableCoreSession session = CoreInstance.openCoreSession(command.getRepository(),
```
bdelbosc: why are you opening a session for each document to edit ?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkActionDescriptor.java
```diff
@@ -24,17 +25,16 @@
 import org.nuxeo.common.xmap.annotation.XObject;
 
 /**
- * The Bulk Operation Service configuration.
- *
```
covolution: No description at all?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkComponent.java
```diff
@@ -34,16 +33,15 @@
  *
  * @since 10.2
  */
-/*
- * BulkAdminService is implemented by the component and not as a service like BulkService because StreamBulkScroller
- * needs bulk configuration during its initialization. Its initialization happens during StreamService's start step
```
covolution: I liked a brief description

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/io/BulkJsonWriter.java
```diff
@@ -36,22 +42,18 @@
...
-        jg.writeStringField("creation", String.valueOf(entity.getCreationDate().toString()));
-        jg.writeStringField("repository", entity.getCommand().getRepository());
-        jg.writeStringField("operation", entity.getCommand().getOperation());
-        jg.writeStringField("username", entity.getCommand().getUsername());
-        jg.writeStringField("query", entity.getCommand().getQuery());
-        if (entity.getScrolledDocumentCount() != null) {
-            jg.writeNumberField("scrolledDocumentCount", entity.getScrolledDocumentCount().longValue());
+        jg.writeStringField(BULK_ID, entity.getId());
+        jg.writeStringField(BULK_STATE, entity.getState().toString());
+        jg.writeStringField(BULK_CREATION, String.valueOf(entity.getCreationInstant().toString()));
```
covolution: String.value of on a toString() ?

covolution:
:+1:

covolution:
Are you planning to add monitorability/metrics/reporting?

funshodavid:
This is planned with https://jira.nuxeo.com/browse/NXP-25099

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/SetPropertiesAction.java
funshodavid: Because it's a mistake :)

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkComponent.java
kevinleturc: This comment was moved to `getAdapter` method for clarity.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkActionDescriptor.java
funshodavid: Not sure this is valuable, as it only contains the action name.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkService.java
```diff
@@ -18,16 +18,26 @@
...
+ * API to manage Bulk Action Framework.
  *
  * @since 10.2
  */
 public interface BulkService {
 
-    BulkStatus runOperation(BulkCommand command);
+    /**
+     * Submits the input {@link BulkCommand} to the Bulk Action Framework system and get the initial status containing
+     * the bulk action id.
```
bdelbosc: We need stricter definition of things:
- terms like "input" and "system" are just noise
- a service that manages a framework is weird, may be keep BAF at documentation level and stick to bulk action at code level ?
- are you running an action or a command ?
- submit may be better name than runAction because of the asynchronous processing

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkService.java
```diff
@@ -18,16 +18,26 @@
...
+     * @return the initial bulk status
+     */
+    BulkStatus runAction(BulkCommand command);
 
-    BulkStatus getStatus(UUID bulkOperationId);
+    /**
+     * Returns the bulk action status of given bulk action id.
+     *
+     * @return the bulk action status of given bulk action id
+     */
```
bdelbosc: duplicate comment

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
```diff
@@ -59,96 +60,76 @@
...
+                                                    .getBytes();
 
             // store the bulk command and status in the key/value store
             KeyValueStore keyValueStore = getKvStore();
-            keyValueStore.put(bulkOperationId + STATE, status.getState().toString());
-            keyValueStore.put(bulkOperationId + CREATION_DATE, status.getCreationDate().toString());
-            keyValueStore.put(bulkOperationId + COMMAND, commandAsBytes);
+            keyValueStore.put(bulkActionId + STATE, status.getState().toString());
+            keyValueStore.put(bulkActionId + CREATION_DATE, status.getCreationInstant().toString());
+            keyValueStore.put(bulkActionId + COMMAND, commandAsBytes);
```
bdelbosc: I though we only need to store the state in the kv store not the command itself

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkService.java
kevinleturc: Agree with you for naming.
I will remove input and system, and remove framework from our code.

For "running an action or a command", indeed it's not clear, here some description: we submit a command, then an action (`SetPropertiesAction` for instance) does the work on the scrolled set.

Submit is good for method name. With that, we submit a command, then we scroll DB, then an action does computation. I don't think there's ambiguity anymore, is it ?

Remaining weird thing is places where we talk about bulk action id, which is in fact the bulk id (the one present in bulk status). We should rename bulkActionId to bulkId.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
kevinleturc: We store it in order to get it for getting the status. Also, we send this command in each record in the system. It's not really a good thing right ? Should we get it directly from K/V ? Don't know if it's better than sending it in record.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkStatus.java
```diff
@@ -35,69 +34,76 @@
...
      */
-    public void setUUID(UUID uuid) {
-        this.uuid = uuid;
+    public void setId(String id) {
+        this.id = id;
     }
 
     /**
-     * Gets bulk operation command.
+     * Gets bulk command.
```
bdelbosc: weird to store the command inside its status

kevinleturc: Do you think it's useless ? We think it could be useful for UI to be able to retrieve the command in the same call

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkStatus.java
```diff
@@ -106,38 +112,38 @@ public void setState(State state) {
...
+    public void setCreationInstant(Instant creationInstant) {
+        this.creationInstant = creationInstant;
     }
 
     /**
-     * Gets number of scrolled document
+     * Gets number of element
      *
-     * @return the number of scrolled document
+     * @return the number of element
```
bdelbosc: the number of elements touched by the operation or processed ?

kevinleturc: It's the number touched by the operation/action.
Processed metrics will be done with [NXP-25099](https://jira.nuxeo.com/browse/NXP-25099)

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/StreamBulkScroller.java
```diff
@@ -120,10 +121,11 @@ public void processRecord(ComputationContext context, String inputStreamName, Re
...
 
         protected void processRecord(ComputationContext context, Record record) {
-            KeyValueStore kvStore = getKeyValueStore();
+            KeyValueStore kvStore = Framework.getService(KeyValueService.class).getKeyValueStore(BULK_KV_STORE_NAME);
             try {
-                BulkCommand command = getBulkCommandJson(record.data);
-                if (!kvStore.compareAndSet(record.key + STATE, SCHEDULED.toString(), BUILDING.toString())) {
+                String bulkActionId = record.getKey();
+                BulkCommand command = BulkCommandHelper.getBulkCommandJson(record.getData());
+                if (!kvStore.compareAndSet(bulkActionId + STATE, SCHEDULED.toString(), BUILDING.toString())) {
```
bdelbosc: This should not happen, it will mean that 2 consumers of the same group read the same partition. But this check can be useful for handling cancellation, anyway if you skip the record you need to call context.askForCompletion

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkStatus.java
```diff
@@ -106,38 +112,38 @@ public void setState(State state) {
...
     /**
-     * Gets bulk operation creation date. This corresponds to the bulk command submission date.
+     * Gets bulk action creation instant. This corresponds to the bulk command submission instant.
      *
-     * @return the creation date
+     * @return the creation instant
      */
-    public ZonedDateTime getCreationDate() {
-        return creationDate;
+    public Instant getCreationInstant() {
```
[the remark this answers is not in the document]
kevinleturc: Indeed...
Here we just want to use `java.time` API in order to correctly handle timezone, etc...

On the other hand, we can customise serialisation of this instant to a millisecond when we send it to K/V, wdyt ?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/StreamBulkScroller.java
```diff
@@ -135,50 +137,45 @@ protected void processRecord(ComputationContext context, Record record) {
...
-                        context.askForCheckpoint();
                         // next batch
                         scroll = session.scroll(scroll.getScrollId());
                         TransactionHelper.commitOrRollbackTransaction();
                         TransactionHelper.startTransaction();
                     }
-                    kvStore.put(record.key + STATE, COMPLETED.toString());
-                    kvStore.put(record.key + SCROLLED_DOCUMENT_COUNT, documentCount);
+
+                    while (!documentIds.isEmpty()) {
```
bdelbosc: not sure to understand why you need to loop here

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/StreamBulkScroller.java
```diff
@@ -135,50 +137,45 @@ protected void processRecord(ComputationContext context, Record record) {
...
-                        // send these ids as keys to the appropriate stream
-                        // key will be bulkOperationId/docId
-                        // value/data is a BulkCommand serialized as JSON
-                        docIds.forEach(docId -> context.produceRecord(command.getOperation(), record.key + '/' + docId,
-                                record.data));
+                        documentIds.addAll(docIds);
+                        if (documentIds.size() >= recordBatchSize) {
+                            // send these ids as keys to the appropriate stream
+                            // key will be bulkActionId/docIds
+                            // value/data is a BulkCommand serialized as JSON
```
bdelbosc: oh the key is the list of ids and the value the bulk command.
the key is limited is not supposed to contain complex or long data, the main usage of the key is to shard records, in your case a simple roundrobin sharding is enough the key can be "%s:%s", actionId, count.
the value which is the record shoudl contains the action id and the list of doc ids.
let's disccuss about this.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/StreamBulkScroller.java
kevinleturc: Will keep the compareAndSet and add `context.askForCompletion()`.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/StreamBulkScroller.java
kevinleturc: We don't indeed.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkStatus.java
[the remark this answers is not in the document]
bdelbosc: Why not it depends on the usage of this method. Internally it is better to store only UTC timestamp.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/StreamBulkScroller.java
bdelbosc: count was supposed to be incremented for each batch, the result will be the same as using a randomUUID without overhead

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
kevinleturc: Choice made is to keep command in K/V, computation will ask for it at beginning of processing.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkStatus.java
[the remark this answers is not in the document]
kevinleturc: Times are stored in K/V as milliseconds, `Instant` is used for application purpose in `BulkStatus` only when getting status from upper services/REST.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkCommand.java
```diff
@@ -63,7 +69,7 @@ public String getRepository() {
     }
 
     /**
-     * @param query the query used to get the document set concerned by the operation
+     * @param query the query used to get the document set concerned by the action
```
bdelbosc: Why not having a normal builder pattern that include a build() method ? How do you ensure that mandatory field like actionId or query are always set ?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkCommand.java
```diff
@@ -74,17 +80,36 @@ public String getQuery() {
...
+    public BulkCommand withAction(String action) {
+        this.action = action;
+        return this;
+    }
+
+    public String getAction() {
+        return action;
+    }
+
+    public BulkCommand withParams(Map<String, Serializable> params) {
```
bdelbosc: This will cause pb on avro conversion, serializable is a java codec non interroperable, 
also if these params are going to be contributed with XML it might be better to start with a map of String

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkCommands.java
```diff
@@ -0,0 +1,72 @@
...
+import org.nuxeo.ecm.core.api.NuxeoException;
+import org.nuxeo.ecm.core.io.registry.MarshallerHelper;
+import org.nuxeo.ecm.core.io.registry.context.RenderingContext;
+import org.nuxeo.runtime.api.Framework;
+import org.nuxeo.runtime.kv.KeyValueService;
+import org.nuxeo.runtime.kv.KeyValueStore;
+
+/**
+ * Helper class for bulk commands.
+ *
```
bdelbosc: Helper is a bad smell, when you need to encode decode the object just use:
        Codec<BulkCommand> codec = service.getCodec("avro", BulkCommand.class);
	byte[] data = codec.encode(myCommand);
the codec can also be used for the KVstore to store the command and its status

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkRecords.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.List;
+
+import org.nuxeo.lib.stream.computation.Record;
+
+/**
+ * Helper class to handle {@link Record} in bulk services.
+ *
+ * @since 10.2
+ */
+public class BulkRecords {
```
bdelbosc: Helper again, with a DTO you don't need this
Command myCommand = codec.decode(record.getData())
myCommand.getId()
myCommand.getDocIds()
Should be the same for Bulk and Counter object

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkService.java
```diff
@@ -18,16 +18,23 @@
...
  */
 public interface BulkService {
 
-    BulkStatus runOperation(BulkCommand command);
+    /**
+     * Submits the {@link BulkCommand} for computation and get the initial status containing the bulk id.
+     *
+     * @param command the command to submit
+     * @return the initial bulk status
+     */
```
bdelbosc: Submits a {@link BulkCommand} that will be processed asynchronously.
@return a unique   bulk command identifier

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkService.java
```diff
@@ -18,16 +18,23 @@
...
+    String submit(BulkCommand command);
 
-    BulkStatus getStatus(UUID bulkOperationId);
+    /**
+     * @return the bulk action status of given bulk id
```
bdelbosc: Returns the command status corresponding to the command identifier.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkStatus.java
```diff
@@ -35,69 +34,76 @@
...
     public enum State {
-        SCHEDULED, BUILDING, COMPLETED
+        /** The {@link BulkCommand} has been submitted to the system. */
+        SCHEDULED,
+
+        /** System is currently scrolling the database. */
+        BUILDING,
+
+        /** System has finished to scroll. */
+        COMPLETED
```
bdelbosc: Don't we need an additional state to denote that scrolling is in progress ? SCHEDULED -> SCROLLING_RUNNING -> RUNNING -> COMPLETED

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkStatus.java
bdelbosc: might be easier to have 2 keys in the KV store one for the Command and one for the Status

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/StreamBulkScroller.java
```diff
@@ -66,52 +64,63 @@
 
     public static final String SCROLL_KEEP_ALIVE_SECONDS_OPT = "scrollKeepAlive";
 
+    public static final String BUCKET_SIZE_OPT = "bucketSize";
```
bdelbosc: Why defining a processor only for the scroller, this should be a bulk action processor that defines the following topology 
(s is for stream, c for computation)
command (s) -> Scroller (c) -> actions ids (s)
counter (s) -> Counter (c)
So you have at least 2 computations a Scroller and a Counter

Also if we want that computations don't write directly to the KV store in order to support a KV failure, we could add a KVWriter computation responsible to update the command status:
command (s) -> Scroller (c) -> actions ids (s)
Scroller (c) -> status (s)
counter (s) -> Counter (c) -> status (s)
status (s) -> KVWriter

funshodavid: The addition of the counter is coming in a separate task: https://jira.nuxeo.com/browse/NXP-25099

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkCommand.java
kevinleturc: We though it will be heavier to have to call build() each time.
Validation is done in service, and currently we just require `repository`. I'll add others.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkStatus.java
kevinleturc: Because `SCROLLING_RUNNING` and `RUNNING` could reflect state of bulk at the same moment. I think we should use two states, one for scroller and another for actions.
We were wondering with @funshodavid, how to detect `RUNNING`, bulk actions could send to status stream an event to set this status (and eventually begin date) if it's not yet set. But this could mean a lot of requests to K/V, wdyt ?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkCommands.java
kevinleturc: Indeed, we'll done it in https://jira.nuxeo.com/browse/NXP-25301.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkRecords.java
kevinleturc: Indeed, we'll done it in https://jira.nuxeo.com/browse/NXP-25301.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkCommand.java
kevinleturc: Converted to Map<String, String>.

covolution: +1 for Map<String, String>.  I have been using this approach for the various descriptor configurations with an interface. https://github.com/nuxeo/ai-core/blob/master/nuxeo-pipes-core/src/main/java/org/nuxeo/runtime/stream/pipes/streams/Initializable.java

PR #2080, opened 2018-06-21
Feature nxp 24400 impl streamworkmanager cancel

pierre-gautier:


nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/StreamWorkManager.java
```diff
@@ -212,9 +190,9 @@ public int getApplicationStartedOrder() {
     @Override
     public void start(ComponentContext context) {
         init();
-        storeState = Boolean.parseBoolean(Framework.getService(ConfigurationService.class).getProperty(STORESTATE_KEY));
-        stateTTL = Long.parseLong(
-                Framework.getService(ConfigurationService.class).getProperty(STATETTL_KEY, STATETTL_DEFAULT_VALUE));
+        ConfigurationService configuration = Framework.getService(ConfigurationService.class);
+        storeState = Boolean.parseBoolean(configuration.getProperty(STORESTATE_KEY));
```
funshodavid: You can use `storeState = configuration.isBooleanPropertyTrue(STORESTATE_KEY)` here

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/StreamWorkManager.java
```diff
@@ -492,6 +480,10 @@ protected boolean awaitCompletionOnQueue(String queueId, long duration, TimeUnit
         return false;
     }
 
+    /**
+     * @deprecated since 10.2 beacause unused
```
funshodavid: Typo

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkManagerImpl.java
```diff
@@ -450,11 +453,15 @@ protected boolean shutdownExecutors(Collection<WorkThreadPoolExecutor> list, lon
...
                 return false;
             }
-            timeout -= unit.convert(System.currentTimeMillis() - t0, TimeUnit.MILLISECONDS);
+            timeout -= System.currentTimeMillis() - t0;
         }
         return true;
     }
 
+    /**
+     * @deprecated since 10.2 beacause unused
```
funshodavid: Typo

pierre-gautier:
currently updating copyright and headers

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/StreamWorkManager.java
```diff
@@ -410,6 +388,16 @@ public boolean shutdown(long timeout, TimeUnit timeUnit) {
         shutdownInProgress = true;
         try {
             boolean ret = streamProcessor.stop(Duration.ofMillis(timeUnit.toMillis(timeout)));
+            // sleep for a given amount of time for works to have time to persist their state and stop properly
```
bdelbosc: simply change set the timeout duration of the streamProcessor.stop to max(timeout, shutdownDelay)

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/StreamWorkManager.java
```diff
@@ -410,6 +388,16 @@ public boolean shutdown(long timeout, TimeUnit timeUnit) {
...
+            long shutDowndelay = Long.parseLong(Framework.getService(ConfigurationService.class)
```
bdelbosc: shutDowndelay or shutdownDelay

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkComputation.java
```diff
@@ -75,27 +73,40 @@ public WorkComputation(String name) {
...
         try {
             if (work.isIdempotent() && workIds.contains(work.getId())) {
                 log.warn("Duplicate work id: " + work.getId() + " skipping");
             } else {
                 boolean storeState = Boolean.parseBoolean(
                         Framework.getService(ConfigurationService.class).getProperty(STORESTATE_KEY));
                 if (storeState) {
-                    StreamWorkManager.setState(work.getId(), Work.State.RUNNING.toString(), stateTTL);
+                    if (WorkStateHelper.getState(work.getId()) != Work.State.SCHEDULED) {
+                        // job has been canceled, saving and returning
```
bdelbosc: log.warn instead of comment, this is an important information

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkManagerImpl.java
```diff
@@ -678,17 +691,21 @@ public void shutdownAndSuspend() throws InterruptedException {
...
-                long shutDowndelay = Long.parseLong(Framework.getService(ConfigurationService.class)
-                                                             .getProperty(SHUTDOWN_DELAY_KEY, "0"));
-                if (shutDowndelay > 0) {
-                    Thread.sleep(shutDowndelay);
+                if (hasWorkRunning) {
+                    // sleep for a given amount of time for works to have time to persist their state and stop properly
+                    long shutdownDelay = Long.parseLong(Framework.getService(ConfigurationService.class)
+                                                                 .getProperty(SHUTDOWN_DELAY_MS_KEY, "0"));
+                    if (shutdownDelay > 0) {
+                        Thread.sleep(shutdownDelay);
```
efge: I didn't review the code that originally added this, but we could be a bit nicer: in a loop sleep 100ms and re-check if there are still running works, until there are no more or the complete timeout has expired. This allows for fast-stop once all works are finished.

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkStateHelper.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    protected static final String KV_NAME = "workManager";
+
+    protected static final String STATE_SUFFIX = ":state";
+
+    protected static final String CANCELED = "canceled";
+
+    protected static Work.State getState(String workId) {
+        KeyValueStore kvStore = Framework.getService(KeyValueService.class).getKeyValueStore(KV_NAME);
```
efge: This could be extracted into a `getKeyValueStore()` static method, as it's used 3 times in the class.

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkStateHelper.java
```diff
@@ -0,0 +1,63 @@
...
+        String stringState = kvStore.getString(workId + STATE_SUFFIX);
```
efge: `workId + STATE_SUFFIX` should also be extracted to a `getStateKey(String workId)` method (or similar name)

nuxeo-core/nuxeo-core-event/src/test/java/org/nuxeo/ecm/core/work/WorkManagerTest.java
```diff
@@ -142,6 +157,10 @@ public void assertDiff(long scheduled, long running, long completed, long cancel
     @Inject
     public FeaturesRunner runner;
 
+    protected int getDuration() {
```
efge: Please name it `getDurationMillis` to be clear about the time unit.

nuxeo-runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/computation/Computation.java
```diff
@@ -68,4 +68,15 @@ default void destroy() {
      * @return computation's metadata.
      */
     ComputationMetadata metadata();
+
+    /**
+     * A hook to inform that computation will be soon destroyed. It gives a way for long processing to cooperate to a
+     * quick shutdown. <br>
+     * This method is not invoked from the computation thread, it should only set some volatile flag and returns
+     * immediately.
+     */
```
efge: Add `@since 10.2`

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkManagerImpl.java
```diff
@@ -450,11 +453,15 @@ protected boolean shutdownExecutors(Collection<WorkThreadPoolExecutor> list, lon
             if (!executor.awaitTermination(timeout, TimeUnit.MILLISECONDS)) {
                 return false;
             }
-            timeout -= unit.convert(System.currentTimeMillis() - t0, TimeUnit.MILLISECONDS);
+            timeout -= System.currentTimeMillis() - t0;
```
efge: This is old code (modulo the unit bug you fixed) but we should also check we don't go into negative values to be robust. If timeout becomes <= 0 we should return false as well.

pierre-gautier: The code called, awaitTermintation(timeout, unit), handles it already, I believe it's OK to let this part as is since it would need to be tested twice and would get uglier.

funshodavid:
:+1:

nuxeo-core/nuxeo-core-event/src/test/java/org/nuxeo/ecm/core/work/StreamWorkManagerTest.java
```diff
@@ -89,54 +94,56 @@ public void testWorkManagerConfigDisableAllAfterStart() throws Exception {
...
     public void testWorkNonIdempotent() throws InterruptedException {
         MetricsTracker tracker = new MetricsTracker();
-        SleepWork work = new SleepWork(1000, false);
+        int duration = getDurationMillis() * 2;
+        SleepWork work = new SleepWork(duration);
         work.setIdempotent(false);
         assertFalse(work.isIdempotent());
         service.schedule(work);
-        assertTrue(service.awaitCompletion(5, TimeUnit.SECONDS));
+        assertTrue(service.awaitCompletion(duration + 100, TimeUnit.MILLISECONDS));
```
bdelbosc: 100ms of margin is too low, a CI slave can run dozen of jobs in concurrence and pause for seconds.
you can use long timeout here, the timeout will be reached only on error case.

nuxeo-core/nuxeo-core-event/src/test/java/org/nuxeo/ecm/core/work/StreamWorkManagerTest.java
```diff
@@ -145,7 +152,7 @@ public void testWorkIdempotentConcurrent() throws InterruptedException {
         service.schedule(work2);
         // we don't know if work1 and work2 are executed on the same thread
         // but we assume that the max duration is work1 + work2 because there is only one invocation of each
-        assertTrue(service.awaitCompletion(2500, TimeUnit.MILLISECONDS));
+        assertTrue(service.awaitCompletion(getDurationMillis() * 3, TimeUnit.MILLISECONDS));
```
bdelbosc: in theory it takes 2s, if it takes 3s it does not work because it means that work1 is executed twice.
The timeout must be lower than 3s try something like "getDurationMillis() * 3 - 200"

nuxeo-core/nuxeo-core-event/src/test/java/org/nuxeo/ecm/core/work/WorkManagerTest.java
```diff
@@ -180,17 +199,16 @@ public void testWorkManagerConfig() throws Exception {
     public void testWorkManagerWork() throws Exception {
         MetricsTracker tracker = new MetricsTracker();
         tracker.assertDiff(0, 0, 0, 0);
-        int duration = 3000; // ms
-        SleepWork work = new SleepWork(duration, false);
+        SleepWork work = new SleepWork(getDurationMillis());
```
bdelbosc: Previously it was 3s now you set it to 200ms, why ?
I am afraid this will create randomness depending on CI load.

nuxeo-core/nuxeo-core-event/src/test/java/org/nuxeo/ecm/core/work/WorkManagerTest.java
```diff
@@ -258,12 +275,9 @@ public void itCanFireWorkFailureEvent() throws Exception {
...
         int initSyncEvntCount = WorkFailureEventListener.getCount();
 
-        int duration = 2000; // 2s
-
-        SleepAndFailWork work = new SleepAndFailWork(0, false);
-        service.schedule(work);
+        service.schedule(new SleepAndFailWork(0));
 
-        assertTrue(service.awaitCompletion(duration * 3, TimeUnit.MILLISECONDS));
+        assertTrue(service.awaitCompletion(getDurationMillis(), TimeUnit.MILLISECONDS));
```
bdelbosc: hmm you change the code and the unit test, hard to say if you break anything

bdelbosc:
Why creating a separate review commit instead of fixing your previous commits ?

nuxeo-core/nuxeo-core-event/src/test/java/org/nuxeo/ecm/core/work/WorkManagerTest.java
pierre-gautier: Because these tests were taking quite long and I try to reduce execution time of tests I work on, 30s each time and the 3 hours we wait on every TP may one day get shorter.
Took it back to 3s.

pierre-gautier: Still in this test, I can see no reason to wait that long.

pierre-gautier:
Because reviews fixes concern various commit on the same file, and even commit that has already been merged.
I'll squash some of them to make things simpler.

pierre-gautier:
up @bdelbosc @efge @funshodavid @kevinleturc

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkStateHelper.java
```diff
@@ -0,0 +1,68 @@
...
+
+    protected static boolean isCanceled(String workId) {
+        return CANCELED.equals(getKeyValueStore().getString(getStateKey(workId)));
+    }
+
+    protected static void setCanceled(String workId) {
+        setState(workId, CANCELED, 0);
+    }
+
+    protected static void setState(String workId, String state, long ttl) {
```
kevinleturc: Does it make sense to have an API which take a `Work.State` ? `String` is only used here for `setCanceled`.

pierre-gautier:
👍

kevinleturc:
👍

PR #2179, opened 2018-07-24
NXP-25494: Create Folderish renditions with File doctype

ron1:


ron1:
Close since already committed from different PR.

PR #2184, opened 2018-07-24
NXP-25468: Add multi tenancy index

akervern:


kevinleturc:
👍 
we should do a T&P to check this change, we can use merge directly option after reviewing this PR.

nuxeo-services/nuxeo-platform-directory/nuxeo-platform-directory-mongodb/src/main/java/org/nuxeo/directory/mongodb/MongoDBDirectory.java
```diff
@@ -125,6 +126,9 @@ protected void initialize(MongoDBSession session) {
         if (dropCollection) {
             collection.drop();
         }
+        if (isMultiTenant()) {
+            collection.createIndex(Indexes.hashed(TENANT_ID_FIELD));
```
efge: This is going to be incorrect if the schema has a prefix.

akervern: Actually, after digging, it seems the resolution is more complex and relying on the prefixed field won't work here.
Issue is the `MongoDBSession` is not expecting any prefix and use the field name directly.

See usage of `org.nuxeo.ecm.directory.BaseSession#TENANT_ID_FIELD`
 - Entry creation: https://github.com/nuxeo/nuxeo/blob/master/nuxeo-services/nuxeo-platform-directory/nuxeo-platform-directory-mongodb/src/main/java/org/nuxeo/directory/mongodb/MongoDBSession.java#L152-L153
 - Query filter: https://github.com/nuxeo/nuxeo/blob/master/nuxeo-services/nuxeo-platform-directory/nuxeo-platform-directory-mongodb/src/main/java/org/nuxeo/directory/mongodb/MongoDBSession.java#L313

So, it deserves more work before relying on the prefixed field. I can create a ticket for the core team if you want

efge: Ok so TENANTI_ID_FIELD is treated specially, so maybe your patch is enough after all...

PR #2236, opened 2018-08-16
NXP-25187: fix permissions exception on CreateUser

mcedica:


nuxeo-services/login/nuxeo-platform-login-openid/src/main/java/org/nuxeo/ecm/platform/oauth2/openid/auth/UserResolver.java
```diff
@@ -54,12 +54,12 @@ protected DocumentModel createNuxeoUser(String nuxeoLogin) {
...
-
-            userDoc = userManager.getBareUserModel();
-            userDoc.setPropertyValue(userManager.getUserIdField(), nuxeoLogin);
-
-            userDoc = userManager.createUser(userDoc);
+            userDoc = Framework.doPrivileged(() -> {
+                UserManager userManager = Framework.getService(UserManager.class);
+                DocumentModel user = userManager.getBareUserModel();
+                user.setPropertyValue(userManager.getUserIdField(), nuxeoLogin);
+                return userManager.createUser(user);
```
efge: We tend to try to limit the scope of what goes inside a `doPrivileged` block, for better auditability. Here I think only this last line need to be inside the block.

ataillefer:
:+1:

dmetzler:
👍

PR #2295, opened 2018-08-28
NXP-25187: fix user mapper creation privilege

vpasquier:
This issue is similar to the one Mariana found in NXP-25187 when using the abstract mapper: it means for instance than the user mapper configuration we have for Amazon cannot work without this fix.

It's not a requirement for NBM, it was just part of our investigation on a use case.

Tell me if you want a dedicated ticket - unfortunately I won't have time to do the tests for openid<->usermapper.

### DIFF
--- a/nuxeo-services/nuxeo-usermapper/src/main/java/org/nuxeo/usermapper/extension/AbstractUserMapper.java
+++ b/nuxeo-services/nuxeo-usermapper/src/main/java/org/nuxeo/usermapper/extension/AbstractUserMapper.java
@@ -31,6 +31,7 @@ import org.nuxeo.ecm.core.api.CoreSession;
 import org.nuxeo.ecm.core.api.DataModel;
 import org.nuxeo.ecm.core.api.DocumentModel;
 import org.nuxeo.ecm.core.api.DocumentModelList;
+import org.nuxeo.ecm.core.api.NuxeoException;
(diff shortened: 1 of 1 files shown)


nuxeo-services/nuxeo-usermapper/src/main/java/org/nuxeo/usermapper/extension/AbstractUserMapper.java
```diff
@@ -151,7 +152,14 @@ protected DocumentModel createPrincipal(Map<String, Serializable> attributes) {
...
         userModel.getDataModel(userManager.getUserSchemaName()).setMap((Map) attributes);
-        return userManager.createUser(userModel);
+        DocumentModel userDoc;
+        try {
+            userDoc = Framework.doPrivileged(() -> userManager.createUser(userModel));
+        } catch (NuxeoException e) {
+            log.error("Error while creating user in UserManager", e);
+            return null;
+        }
+        return userDoc;
```
efge: You can put the `return` inside the `try`:
 
         try {
             return Framework.doPrivileged(() -> userManager.createUser(userModel));
         } ...

efge:
I would like a more detailed description of why this is needed please (inside the previous ticket if you want). What breaks and when if we don't change the code?
I'm asking because I'm wary of having other code that calls this method and expects security checks to be done by it. If we always do a `doPrivileged` then it would be a security hole (if we have such code).

PR #2312, opened 2018-09-04
Document VCS Tests

jcarsique:
Explain the logic and the related variables and properties.


jcarsique:
Git cleanup before merge: branch renamed to NXP-25739-doc-vcstest, commit title accordingly amended

PR #2325, opened 2018-09-06
Feature nxp 25378 use baf in trash

pierre-gautier:
As 6 T&P's have timed out in a row, I create to PR manually to start review before it passes


nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/AbstractBulkAction.java
```diff
@@ -137,22 +132,31 @@ public void processRecord(ComputationContext context, String inputStreamName, Re
...
 
         @Override
         public void destroy() {
-            getLog().debug(
-                    String.format("Destroy computation: %s, pending entries: %d", metadata.name(), documentIds.size()));
+            getLog().debug(String.format("Destroy: %s, pending entries: %d", metadata.name(), documentIds.size()));
+        }
+
+        /**
+         * Define if the computation does produce count records.<br>
```
covolution: `Define if the computation produces count records.`

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public static class RemoveProxyComputation extends AbstractBulkComputation {
+
+        public RemoveProxyComputation(String name, int batchSize, int batchThresholdMs) {
+            super(name, 1, 1, batchSize, batchThresholdMs);
+        }
+
+        @Override
+        protected void compute(CoreSession session, List<String> ids, Map<String, Serializable> properties) {
+            Collection<DocumentRef> refs = ids.stream().map(IdRef::new).collect(Collectors.toList());
```
covolution: I am wondering if a standard `foreach` would be simpler than multiple streams?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    protected Computation createComputation(int batchSize, int batchThresholdMs) {
+        return new RemoveProxyComputation(getActionName(), batchSize, batchThresholdMs);
+    }
+
+    public static class RemoveProxyComputation extends AbstractBulkComputation {
+
+        public RemoveProxyComputation(String name, int batchSize, int batchThresholdMs) {
+            super(name, 1, 1, batchSize, batchThresholdMs);
```
covolution: Just an observation. This computation takes 1 input stream and the `AbstractBulkComputation` writes to 1 output stream, which is the counter.  Is that right?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/SetSystemPropertiesAction.java
```diff
@@ -45,15 +56,34 @@ protected Computation createComputation(int batchSize, int batchThresholdMs) {
...
+            Map<String, Serializable> actualProperties = new HashMap<>(properties);
+            String eventId = (String) actualProperties.remove(NOTIFY);
+            Collection<DocumentRef> refs = ids.stream().map(IdRef::new).collect(Collectors.toList());
+            ids.forEach(id -> actualProperties.forEach((k, v) -> session.setDocumentSystemProp(new IdRef(id), k, v)));
             session.save();
+            if (eventId != null) {
+                EventService eventService = Framework.getService(EventService.class);
+                DocumentModelList docs = session.getDocuments(refs.toArray(new DocumentRef[refs.size()]));
+                docs.forEach(d -> {
+                    DocumentEventContext ctx = new DocumentEventContext(session, session.getPrincipal(), d);
```
covolution: You could extract this to a `raiseEvent` method.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/DocumentSetRepositoryInit.java
```diff
@@ -28,15 +28,45 @@
  */
 public class DocumentSetRepositoryInit extends DefaultRepositoryInit {
 
-    private static final int SIZE = 10;
+    public static final String ROOT = "/default-domain/workspaces/test";
+
+    public static final int SIZE = 3;
+
+    public static int CREATED_NON_PROXY = 0;
```
covolution: If these are variables (not constants) then they shouldn't be UPPER_CASE_WITH_UNDERSCORES.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestRemoveProxyAction.java
```diff
@@ -0,0 +1,82 @@
...
+    @Inject
+    public CoreSession session;
+
+    @Test
+    public void testRemoveProxy() throws Exception {
+
+        DocumentModel model = session.getDocument(new PathRef("/default-domain/workspaces/test"));
+        String nxql = String.format("SELECT * from Document where ecm:ancestorId='%s'", model.getId());
+
+        assertEquals(CREATED_PROXY, session.query(nxql + " and ecm:isProxy=1").size());
```
covolution: Ah, so you are using static variable from another class in the test with a static import.  Perhaps just add a comment to mention they are from `DocumentSetRepositoryInit` or don't use the static import.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/PropertyTrashService.java
```diff
@@ -90,13 +99,13 @@ protected void doTrashDocument(DocumentModel doc) {
...
         for (DocumentModel doc : docs) {
-            docRefs.addAll(doUntrashDocument(doc));
+            docRefs.addAll(doUntrashDocument(doc, true));
         }
         docs.stream().map(DocumentModel::getCoreSession).findFirst().ifPresent(CoreSession::save);
         return docRefs;
     }
 
-    protected Set<DocumentRef> doUntrashDocument(DocumentModel doc) {
+    protected Set<DocumentRef> doUntrashDocument(DocumentModel doc, boolean processChildren) {
```
covolution: This isn't getting backported so I suppose its fine to change the method signature.  I might have changed the existing method to call this method with default value for `processChildren`.  That way, existing code the uses it wouldn't break  (it is a `protected` method though, so unlikely to be used elsewhere).

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/PropertyTrashService.java
```diff
@@ -114,20 +123,49 @@ protected void doTrashDocument(DocumentModel doc) {
...
             DocumentModel parent = session.getDocument(parentRef);
-            parent.putContextData(SKIP_CHILDREN_PROCESSING_KEY, Boolean.TRUE);
-            Set<DocumentRef> ancestorDocRefs = doUntrashDocument(parent);
+            Set<DocumentRef> ancestorDocRefs = doUntrashDocument(parent, false);
             docRefs.addAll(ancestorDocRefs);
         }
         return docRefs;
     }
 
+    protected void setTrashPropertyToChildren(DocumentModel model, Boolean value) {
```
covolution: javadoc

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/PropertyTrashService.java
```diff
@@ -114,20 +123,49 @@ protected void doTrashDocument(DocumentModel doc) {
...
+                currentTrashValue, model.getId());
+        service.submit(new BulkCommand().withRepository(session.getRepositoryName())
+                                        .withUsername(session.getPrincipal().getName())
+                                        .withQuery(nxql)
+                                        .withAction(SetSystemPropertiesAction.ACTION_NAME)
+                                        .withParam(SYSPROP_IS_TRASHED, value)
+                                        .withParam(NOTIFY, eventId));
+    }
+
+    protected void removeProxies(DocumentModel model) {
```
covolution: javadoc

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestRemoveProxyAction.java
pierre-gautier: I'm basically against basic import, but that's how they are used in nuxeo, so says @efge

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/PropertyTrashService.java
pierre-gautier: Agree, but that's not backported and 10.1 protected method, I'd rather do this than duplicate a method for nothing

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
```diff
@@ -0,0 +1,70 @@
...
+            Collection<DocumentRef> proxies = docs.stream()
+                                                  .filter(DocumentModel::isProxy)
+                                                  .map(DocumentModel::getRef)
+                                                  .collect(Collectors.toList());
+            session.removeDocuments(proxies.toArray(new DocumentRef[0]));
+            session.save();
+        }
+
+    }
+}
```
funshodavid: Maybe not for this one, but we should start putting actions in their associated module (here `nuxeo-core`) IMO, and not list all actions in the `nuxeo-core-bulk` module

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
pierre-gautier: Hum the point is using the 2 methods taking a list, so it could not be done in only one forearch.
First one would only be more lines, for the second one it would be equivalent but since we began with streams let's finish with them.
wdyt ?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
pierre-gautier: yes, the 1 ouput is count, but that's a good question, should we compute this value in accordance of doesCount ?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/SetSystemPropertiesAction.java
pierre-gautier: do you believe I should ? I'm not really kind of only-called-once methods

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
pierre-gautier: Do you mean that we should move all implementations to nuxeo-core ?
I don't have a strong opinion, but I would'nt mind having implementations requiring only nuxeo-core-api in the bulk module
@kevinleturc @efge  ?

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/PropertyTrashService.java
pierre-gautier: it's not required as method is protected, do you think it's needed or simple enough ?

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/PropertyTrashService.java
pierre-gautier: it's not required as method is protected, do you think it's needed or simple enough ?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/SetSystemPropertiesAction.java
covolution: it will reduce the cognitive complexity, but up to you.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/PropertyTrashService.java
covolution: I thought protected methods needed javadoc.

covolution:
:+1:

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
kevinleturc: Agree with @pierre-gautier

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/SetSystemPropertiesAction.java
```diff
@@ -45,15 +56,34 @@ protected Computation createComputation(int batchSize, int batchThresholdMs) {
...
+                    ctx.setProperties(new HashMap<>(d.getContextData()));
```
kevinleturc: No need to give `contextData`, they're empty as you just fetched them.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestSetPropertiesAction.java
```diff
@@ -80,9 +86,11 @@ public void testSetProperties() throws Exception {
         BulkStatus status = service.getStatus(commandId);
         assertNotNull(status);
         assertEquals(COMPLETED, status.getState());
-        assertEquals(10, status.getProcessed());
+        assertEquals(SIZE * 2, status.getProcessed());
```
kevinleturc: You should use one of other static variables because it's hard to understand why we want `SIZE * 2` if we're not already aware of the code.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestSetSystemPropertiesAction.java
```diff
@@ -52,33 +61,57 @@
     @Inject
     public CoreSession session;
 
+    @Inject
+    public EventService events;
```
kevinleturc: Can you rename it to `eventService` ?

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestSetSystemPropertiesAction.java
```diff
@@ -52,33 +61,57 @@
...
+
     @Inject
     public TransactionalFeature txFeature;
 
+    private static class TestListener extends EventListenerDescriptor {
+
+        private int count;
+
+        private TestListener() {
+            events = new HashSet<>(Arrays.asList(DOCUMENT_TRASHED));
```
kevinleturc: Use `Collections#singletonList` instead.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/trash/test/TestTrashBackwardMechanism.java
```diff
@@ -120,12 +116,10 @@ public void testMechanismTriggersOneListener() {
 
     public static class BulkListenersFilter implements LogCaptureFeature.Filter {
 
-        protected final Set<String> bulkListenerNames = Stream.of(BulkLifeCycleChangeListener.class,
-                BulkTrashedStateChangeListener.class).map(Class::getName).collect(Collectors.toSet());
-
         @Override
         public boolean accept(LoggingEvent event) {
-            return Level.DEBUG.equals(event.getLevel()) && bulkListenerNames.contains(event.getLoggerName());
+            return Level.DEBUG.equals(event.getLevel())
```
kevinleturc: Test principle seems to have change regarding the comment on the test method and the test method name.
This test was there to ensure we don't loop twice and also to prevent issues on MySQL/MariaDB where transaction level could generate issues depending on which listener was called first.

nuxeo-features/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/test/java/org/nuxeo/elasticsearch/test/TestTreeIndexing.java
```diff
@@ -398,6 +403,7 @@ public void shouldStoreOnlyEffectiveACEs() throws Exception {
     }
 
     @Test
+    @Deploy("org.nuxeo.elasticsearch.core.test:dummy-bulk-login-config.xml")
```
kevinleturc: You can use the one present in `nuxeo-core-bulk`test-jar.

nuxeo-features/nuxeo-platform-userworkspace/nuxeo-platform-userworkspace-core/src/test/java/org/nuxeo/ecm/platform/userworkspace/core/tests/TestUserWorkspace.java
```diff
@@ -359,12 +364,14 @@ public void testCanRetrieveUserWorkspaceWithTrashedDomain() {
                                         .collect(Collectors.toList());
         TrashService trashService = Framework.getService(TrashService.class);
         trashService.trashDocuments(docs);
+        bulkService.await(Duration.ofSeconds(30));
```
kevinleturc: You ask `CoreFeature#waitForAsyncCompletion` instead to avoid having `BulkService` here.

nuxeo-runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/computation/log/ComputationRunner.java
```diff
@@ -321,6 +321,8 @@ protected void checkpointIfNecessary() {
                 checkpoint();
                 completed = true;
                 inCheckpointRecords = inRecords;
+            } catch (Exception e) {
+                log.error(e, e);
```
kevinleturc: Is this intended ? It could hide problems as the current runner won't be stopped.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/AbstractBulkAction.java
```diff
@@ -137,22 +132,31 @@ public void processRecord(ComputationContext context, String inputStreamName, Re
...
-            getLog().debug(
-                    String.format("Destroy computation: %s, pending entries: %d", metadata.name(), documentIds.size()));
+            getLog().debug(String.format("Destroy: %s, pending entries: %d", metadata.name(), documentIds.size()));
+        }
+
+        /**
+         * Define if the computation produces count records.<br>
+         * Intermediate computations don't want to do it.
+         */
+        protected boolean doesCount() {
```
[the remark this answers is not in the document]
covolution: I asked for this because I have intermediate computations.  At the moment, I don't need "another abstract computation".  I already have an implementation based on `AbstractBatchComputation`.  I am just blocked waiting for this.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/trash/test/TestTrashBackwardMechanism.java
pierre-gautier: Actually, with MySQL the tests fails without the Filter and works fine with, what's the problem ?

nuxeo-runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/computation/log/ComputationRunner.java
pierre-gautier: Bad move actually, thanks !

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/AbstractBulkAction.java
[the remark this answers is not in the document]
covolution: I don't prefer to wait! ;)

nuxeo-features/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/test/java/org/nuxeo/elasticsearch/test/TestTreeIndexing.java
pierre-gautier: Actually, the one in nuxeo-core-bulk test jar is already loaded but override by another module, it has to be deployed there, at the last moment, with another component id

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/AbstractBulkAction.java
```diff
@@ -77,10 +77,7 @@ protected AbstractBulkAction(String name) {
     public Topology getTopology(Map<String, String> options) {
         int size = getOptionAsInteger(options, BATCH_SIZE_OPT, DEFAULT_BATCH_SIZE);
         int timer = getOptionAsInteger(options, BATCH_THRESHOLD_MS_OPT, DEFAULT_BATCH_THRESHOLD_MS);
-        return Topology
-                       .builder()
-                       .addComputation(() -> createComputation(size, timer), getIOs())
-                       .build();
+        return Topology.builder().addComputation(() -> createComputation(size, timer), getIOs()).build();
```
bdelbosc: Watch out the naming: timer/threshold, getIOs/getStreamMapping

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
```diff
@@ -0,0 +1,73 @@
...
+
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentRef;
+import org.nuxeo.ecm.core.api.IdRef;
+import org.nuxeo.ecm.core.api.IterableQueryResult;
+import org.nuxeo.ecm.core.query.sql.NXQL;
+import org.nuxeo.lib.stream.computation.Computation;
+
+/**
+ * @since 10.3
```
bdelbosc: Describe the action plz so I understand why proxy deserves a specific action instead of being handled by a generic action

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
bdelbosc: if this is the convention used in a bulk computation, it should be stated somewhere

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
bdelbosc: I think yes action should be defined where they make sense in their context, but for now, it is ok to keep everything in the same module we will see that later in the polish epic.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/SetSystemPropertiesAction.java
bdelbosc: I agree, simply returns when eventId is null and extract a method will make thing readable.
Small methods are always good: they express intention and make profiling dead simple when using flame graph.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/SetSystemPropertiesAction.java
```diff
@@ -45,15 +56,33 @@ protected Computation createComputation(int batchSize, int batchThresholdMs) {
...
+
         public SetSystemPropertyComputation(String name, int batchSize, int batchThresholdMs) {
             super(name, 1, 1, batchSize, batchThresholdMs);
         }
 
         @Override
         protected void compute(CoreSession session, List<String> ids, Map<String, Serializable> properties) {
-            ids.forEach(id -> properties.forEach((k, v) -> session.setDocumentSystemProp(new IdRef(id), k, v)));
+            Map<String, Serializable> actualProperties = new HashMap<>(properties);
+            String eventId = (String) actualProperties.remove(NOTIFY);
```
bdelbosc: Why all this code to get a flag which is not event an eventId ?

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/DocumentSetRepositoryInit.java
```diff
@@ -28,15 +28,47 @@
  */
 public class DocumentSetRepositoryInit extends DefaultRepositoryInit {
 
-    private static final int SIZE = 10;
+    public static final String ROOT = "/default-domain/workspaces/test";
+
+    public static final int SIZE = 3;
```
bdelbosc: can't you state how much document or proxy you are going to create instead of size and level?

nuxeo-features/nuxeo-platform-userworkspace/nuxeo-platform-userworkspace-core/src/test/java/org/nuxeo/ecm/platform/userworkspace/core/tests/TestUserWorkspace.java
```diff
@@ -359,12 +359,14 @@ public void testCanRetrieveUserWorkspaceWithTrashedDomain() {
                                         .collect(Collectors.toList());
         TrashService trashService = Framework.getService(TrashService.class);
         trashService.trashDocuments(docs);
+        coreFeature.waitForAsyncCompletion();
```
bdelbosc: Does it mean that trashDocuments turned to async, or tests was buggy ?

pierre-gautier: It did turn async, children of documents trashed are handled by BAF now

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/SetSystemPropertiesAction.java
pierre-gautier: It's because we have to remove and get NOTIFY to iterate on all other and the provided map is unmodifiable
We could also jut get and put a filter in the stream line 70
Which one do you prefer ?

bdelbosc: ok, I see, I don't think it is going to be a pb so no change required.

nuxeo-features/nuxeo-platform-userworkspace/nuxeo-platform-userworkspace-core/src/test/java/org/nuxeo/ecm/platform/userworkspace/core/tests/TestUserWorkspace.java
bdelbosc: then org.nuxeo.ecm.core.api.trash.TrashService#trashDocuments java doc should be updated accordingly like (at least remove the "Placeless documents are deleted immediately.")

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/actions/RemoveProxyAction.java
pierre-gautier: It's part of the topology required for 'bafing' property trash service.
It's the most generic I could do.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
```diff
@@ -120,26 +120,15 @@ public KeyValueStore getKvStore() {
 
     @Override
     public boolean await(Duration duration) throws InterruptedException {
```
efge: `duration` is not used anywhere anymore. I don't think this is right.

PR #2326, opened 2018-09-06
NXP-25075: Add an AWS configuration service

loopingz:


covolution:
Did you see the comment I had added to [NXP-25596]( https://jira.nuxeo.com/browse/NXP-25596?focusedCommentId=353646&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-353646)? The AWS SDK already provides a provider chain.   Have you considered using a Provider Chain rather than this approach?

loopingz:
@covolution No i did not see your comment, I like your approach the fact to use nuxeo.s3.... for AI is not really logical. So we should aim for nuxeo.aws. instead.
If you look to the code, yes I know that AWS provide the chain, the issue was that the Nuxeo code is not using it by default, also with your implementation if AWS change the default chain we will not leverage this one.

covolution:
I tried to re-use AWS code as much as possible.  I also did a RegionProviderChain. https://github.com/nuxeo/ai-aws/blob/master/ai-aws-core/src/main/java/org/nuxeo/aws/credentials/NuxeoRegionProviderChain.java

nuxeo-runtime/nuxeo-runtime-aws/src/main/java/org/nuxeo/runtime/aws/AWSCredentialsDescriptor.java
```diff
@@ -0,0 +1,50 @@
...
+import com.amazonaws.auth.AWSCredentials;
+import org.apache.commons.lang3.StringUtils;
+import org.nuxeo.common.xmap.annotation.XNode;
+import org.nuxeo.common.xmap.annotation.XObject;
+
+
+@XObject("configuration")
+public class AWSCredentialsDescriptor implements AWSCredentials {
+
+    @XNode("accessKeyId")
```
pierre-gautier: Can you please implement Descriptor ?

nuxeo-runtime/nuxeo-runtime-aws/src/main/java/org/nuxeo/runtime/aws/NuxeoAWSCredentialsProviderImpl.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public void registerContribution(Object contribution,
+                                     String extensionPoint, ComponentInstance contributor) {
+        if ("credentials".equals(extensionPoint)) {
+            credentials = (AWSCredentialsDescriptor) contribution;
+        }
+    }
+
```
pierre-gautier: Can you please comply to new way of writing a Component ?
You can see there an example there : https://github.com/nuxeo/nuxeo/blob/fa7eb83350647d0bcc95727704e837bfb9a0a0bd/nuxeo-runtime/nuxeo-runtime-stream/src/main/java/org/nuxeo/runtime/avro/AvroComponent.java

nuxeo-runtime/nuxeo-runtime-aws/src/main/java/org/nuxeo/runtime/aws/NuxeoAWSCredentialsProvider.java
```diff
@@ -0,0 +1,59 @@
...
+            if (credentials != null) {
+                return credentials;
+            }
+        }
+        return DEFAULT.getCredentials();
+    }
+
+    @Override
+    public void refresh() {
+        DEFAULT.refresh();
```
covolution: Hmm, we could be calling `refresh` on a provider that has never been used.  Is there a test to check that doesn't error?

nuxeo-runtime/nuxeo-runtime-aws/src/main/java/org/nuxeo/runtime/aws/AWSConfigurationServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    public static final String XP_CONFIGURATION = "configuration";
+
+    protected AWSConfigurationDescriptor getDescriptor() {
+        return getDescriptor(XP_CONFIGURATION, Descriptor.UNIQUE_DESCRIPTOR_ID);
+    }
+
+    @Override
+    protected String getName() {
+        return "org.nuxeo.runtime.aws.AWSConfigurationService";
+    }
```
pierre-gautier: don't you want to merge it after https://github.com/nuxeo/nuxeo/pull/2371 and remove getName() ?

nuxeo-runtime/nuxeo-runtime-aws/src/main/java/org/nuxeo/runtime/aws/AWSConfigurationDescriptor.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @XNode("sessionToken")
+    protected String sessionToken;
+
+    @XNode("region")
+    protected String region;
+
+    @Override
+    public String getId() {
+        return UNIQUE_DESCRIPTOR_ID;
```
covolution: Returns an empty String?

nuxeo-runtime/nuxeo-runtime-aws/src/test/resources/log4j.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0"?>
```
pierre-gautier: is it log4j2 ready ?

pierre-gautier:
👍

nuxeo-runtime/nuxeo-runtime-aws/src/test/java/org/nuxeo/runtime/aws/TestAWSConfigurationService.java
```diff
@@ -0,0 +1,158 @@
...
+import com.amazonaws.auth.AWSSessionCredentials;
+import com.amazonaws.auth.EnvironmentVariableCredentialsProvider;
+import com.amazonaws.regions.AwsEnvVarOverrideRegionProvider;
+
+@RunWith(FeaturesRunner.class)
+@Features(RuntimeFeature.class)
+@Deploy("org.nuxeo.runtime.aws")
+public class TestAWSConfigurationService {
+
+    protected static final String AWS_ACCESS_KEY_ID_SYSTEM_PROPERTY = "aws.accessKeyId";
```
covolution: These are constants in `SDKGlobalConfiguration`

covolution:
:+1:

nuxeo-runtime/nuxeo-runtime-aws/src/main/java/org/nuxeo/runtime/aws/NuxeoAWSCredentialsProvider.java
efge: From my reading of the default providers' code it seems supported and harmless. Providers don't need to have been called to be fully operational in terms of internal datastructures.

nuxeo-runtime/nuxeo-runtime-aws/src/main/java/org/nuxeo/runtime/aws/AWSConfigurationDescriptor.java
efge: Yes that's the convention for XML config where there's no "id" to separate several configs.

efge:
And even though we document that nuxeo.conf properties should be used, the idea is not to read these properties from Java, but let them be expanded into XML config then read the XML config.

PR #2337, opened 2018-09-11
Use exec to launch command in Unix systems

rlabolle:
Launch commands using `/bin/sh -c 'exec ...'` instead of simply `/bin/sh -c '...'`.

This fix an error when using nuxeo with systemd: "Supervising process <pid from nuxeo.pid> which is not our child. We'll most likely not notice when it exits."

### DIFF
--- a/nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
+++ b/nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
@@ -871,6 +871,7 @@ public abstract class NuxeoLauncher {
     private List<String> getUnixCommand(List<String> roughCommand) {
         ArrayList<String> osCommand = new ArrayList<>();
         String linearizedCommand = new String();
+        linearizedCommand += "exec";
         for (String commandToken : roughCommand) {
             if (StringUtils.isBlank(commandToken)) {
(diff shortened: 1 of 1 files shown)


rlabolle:
I'm launching nuxeo via systemd with something close to that configuration:
https://doc.nuxeo.com/nxdoc/configuring-the-nuxeo-platform-as-a-daemon-with-systemd/

With an additional line like that:
```
PIDFile=/var/run/nuxeo/nuxeo.pid
```
So systemd is aware of the main PID of nuxeo and is able to restart it if it crash.

To be able to monitor the process efficiently, it must be a direct child of the init process.

Currently, we have this kind of process tree:
```
systemd,1
  └─sh,88149 -c...
     └─java,88150 (nuxeo main process)
```
And what we really want here is:
```
systemd,1
  └─java,88150 (nuxeo main process)
```

rlabolle:
Additional precision: this code is part of nuxeo-launcher project which is behind nuxeoctl command, so it concern the main nuxeo process launched by `nuxeoctl start`

efge:
Thanks for the info. I'll create a ticket and take this patch.
Do you have any changes to the doc that you think are worth changing? Is that `PIDFile` line something that is required/recommended?

efge:
Integrated in branch 9.10 with commit 33fcf1dc8a297e672d0d098a89bd5a95c61d4d60

PR #2387, opened 2018-09-28
Feature nxp 25269 multibranch

jcarsique:
Fix parameters expansion and rename params in the same time to avoid collision with already used namings.


efge:
👍

PR #2389, opened 2018-09-28
NXP-25887: Dynamic digest in PORTAL_AUTH

fmunch:
Allows the use of other (more robust) digests in PORTAL_AUTH while still falling back to MD5.

### DIFF
--- a/nuxeo-services/login/nuxeo-platform-login-portal-sso/src/main/java/org/nuxeo/ecm/platform/ui/web/auth/portal/PortalAuthenticator.java
+++ b/nuxeo-services/login/nuxeo-platform-login-portal-sso/src/main/java/org/nuxeo/ecm/platform/ui/web/auth/portal/PortalAuthenticator.java
@@ -15,6 +15,7 @@
  *
  * Contributors:
  *     Nuxeo - initial API and implementation
+ *     Florent Munch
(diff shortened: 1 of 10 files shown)


efge:
Thanks for the contribution.

I would like to discuss the use cas though. Do you really need the digest to be controlled by the client? Do you have use cases where different clients will use different digests?

If not, I'd prefer if the digest algorithm (better name than just "digest" btw) was just made configurable server-side through the configuration service. It minimizes decisions taken based on client input, which is always better.

fmunch:
I can use a `nuxeo.auth.portal.digest.algorithm` property instead if you want. I don't see a use case where we would _need_ different algorithms.

efge:
That would work yes. I've opened [NXP-25887](https://jira.nuxeo.com/browse/NXP-25887) to track this, if you could change the commit message as well.

Just to be clear, I'm talking about using `ConfigurationService` to get the value, not `Framework.getProperty`.

fmunch:
The digest algorithm is now a property.

efge:
See #2416 for the followup pull request with a few details changed in the tests.

efge:
@fmunch note that after reflection I changed the configuration mechanism, see the new PR.

PR #2390, opened 2018-09-30
NXP-25780: expose a configuration property for EditableModel

tmartins:
It will allow an alternative management of a missing row in the
EditableModel object.


efge:
👍

PR #2421, opened 2018-10-11
Fix nxp 25633 bulk add status completed time

bdelbosc:


nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/io/BulkConstants.java
```diff
@@ -21,31 +21,37 @@
 /**
  * Bulk constants for serialization.
  *
- * @since 10.2
+ * @since 10.3
  */
 public class BulkConstants {
 
-    // BulkStatus part
+    private BulkConstants() {
```
covolution: Why has the private method moved to the top?

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
```diff
@@ -76,15 +83,12 @@
     @AvroEncode(using = InstantAsLongEncoding.class)
     protected Instant scrollEndTime;
 
-    @Nullable
-    protected Long processed;
+    @AvroEncode(using = InstantAsLongEncoding.class)
```
covolution: I don't really understand the how the code works but I assume it does!  Here it looks like you are serialising instants as longs with avro?  But elsewhere you have Instants in json, which are serialized as ISO-8601 strings.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/io/BulkConstants.java
bdelbosc: yeah weird will fix it

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
bdelbosc: personally I would have prefer normal timestamp everywhere, but it makes sense to have instant encoded as timestamp with a long instead a string representation that is much more bigger.

PR #2432, opened 2018-10-15
NXP-25353: Comments on WebUI

nmpcunha:
https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-ncunha/19/

https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-ncunha/21/


nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentJsonWriter.java
```diff
@@ -47,13 +55,22 @@
...
+    protected CommentManager commentManager;
+
     public CommentJsonWriter() {
         super(COMMENT_ENTITY_TYPE, Comment.class);
     }
 
     @Override
     protected void writeEntityBody(Comment entity, JsonGenerator jg) throws IOException {
         writeCommentEntity(entity, jg);
+        boolean includeRepliesSummary = ctx.getFetched(COMMENT_ENTITY_TYPE).contains(FETCH_REPLIES_SUMMARY);
```
guirenard: I am not sure why the `writeCommentEntity` is `static`, it shouldn't IMO. And the `includeRepliesSummary` fetch thing would go inside.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-rest-api/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/comment/CommentAdapter.java
```diff
@@ -79,6 +89,10 @@ public Comment getExternalComment(@PathParam("entityId") String entityId) {
     @PUT
     @Path("{commentId}")
     public Comment updateComment(@PathParam("commentId") String commentId, Comment comment) {
+        if (!getContext().getCoreSession().exists(new IdRef(commentId))) {
+            throw new DocumentNotFoundException("The comment " + commentId + " does not exist.");
+        }
+
```
guirenard: Flo will probably tell you to remove this extra line :D

guirenard: The service is actually checking that the document exists but throw an `IllegalArgumentException` why doing this extra check?  If for a good reason, then the service should throw `DocumentNotFoundException` instead

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/resources/OSGI-INF/comment-pageprovider-contrib.xml
```diff
@@ -15,6 +15,7 @@
       <pattern>
         SELECT * FROM Comment WHERE comment:parentId = ?
       </pattern>
+      <sort column="comment:creationDate" ascending="false"/>
```
guirenard: To double-check, the service has
```
        GET_COMMENTS_FOR_DOC_PAGEPROVIDER_NAME, singletonList(new SortInfo("dc:created", true)), pageSize,
                currentPageIndex, props, documentId);
```
so this must be ineffective. What's the difference between `dc:created` and `comment:creationDate` btw?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-rest-api/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/comment/CommentAdapter.java
nmpcunha: I agree with you but I think @funshodavid may have a strong argument for that.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/resources/OSGI-INF/comment-pageprovider-contrib.xml
nmpcunha: That singletonList was removed in order to follow the contrib.

Your question is very relevant, since, currently, we need to pass the comment:creationDate as an argument and it is not validated. This means that we can create comments with past dates or, even, in the future. I raised this question for Team Core, but still without any reply.

Anyway, since we are relying on comment schema fields to represent comments, the usage of comment:creationDate as sort column was only for coherence purposes.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-rest-api/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/comment/CommentAdapter.java
efge: Strictly speaking `DocumentNotFoundException` is an implementation detail of how comments are stored. They may not be stored as documents so it doesn't really make sense to throw that.
@funshodavid plans on making changes to the comment service itself to throw a `NuxeoException` with a 400 status code instead of an `IllegalArgumentException`. Would a 400 be ok for the client part @nmpcunha?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/resources/OSGI-INF/comment-pageprovider-contrib.xml
efge: I don't remember seeing the question about `comment:creationDate`. Could you re-ask in #team-core please?

nmpcunha: > I don't remember seeing the question about `comment:creationDate`. Could you re-ask in #team-core please?

Done!

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentJsonWriter.java
nmpcunha: Other classes are relying on this method (e.g. AnnotationJsonWriter) and, for this reason, I won't change it for now, but would be interesting to have some feedback from Team Core (@funshodavid)

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-api/src/main/java/org/nuxeo/ecm/platform/comment/api/CommentImpl.java
```diff
@@ -147,4 +148,26 @@ public String getEntity() {
     public void setEntity(String entity) {
         this.entity = entity;
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
kevinleturc: Can you use commons-lang for such thing ?
For instance:
```
return EqualsBuilder.reflectionEquals(this, o);
```

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-api/src/main/java/org/nuxeo/ecm/platform/comment/api/CommentManager.java
```diff
@@ -186,6 +188,17 @@
...
+    /**
+     * Gets all comments for a document.
+     *
+     * @param session the core session
+     * @param documentId the document id
+     * @param sortAscending whether to sort ascending or descending
+     * @return the list of comments, or an empty list if no comment is found
+     * @since 10.3
+     */
+    List<Comment> getComments(CoreSession session, String documentId, boolean sortAscending);
```
kevinleturc: What the purpose of `sortAscending` ? Comments should already by sorted in one way, if you need them to be reverse sorted can't you do it client side ?
My concern is this service already has a lot of API and I don't see this one as needed.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentJsonWriter.java
```diff
@@ -78,4 +95,13 @@ protected static void writeCommentEntity(Comment entity, JsonGenerator jg) throw
             jg.writeStringField(EXTERNAL_ENTITY, ((ExternalEntity) entity).getEntity());
         }
     }
+
+    protected void writeRepliesSummary(Comment entity, JsonGenerator jg) throws IOException {
+        CoreSession session = ctx.getSession(null).getSession();
+        PartialList<Comment> comments = commentManager.getComments(session, entity.getId(), 1L, 0L, false);
```
kevinleturc: Not sure if it's the right place to do it. Do you find another place where we did it ?
Let's discuss it with @efge and/or @troger

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/PropertyCommentManager.java
```diff
@@ -132,10 +135,10 @@ public void deleteComment(DocumentModel docModel, DocumentModel comment) {
     public DocumentModel getThreadForComment(DocumentModel comment) {
         CoreSession session = comment.getCoreSession();
         DocumentModel parent = session.getDocument(new IdRef((String) comment.getPropertyValue(COMMENT_PARENT_ID)));
-        while (COMMENT_DOC_TYPE.equals(parent.getType())) {
-            comment = getThreadForComment(parent);
+        if (!COMMENT_DOC_TYPE.equals(parent.getType())) {
```
kevinleturc: Can you reverse condition ? just for clarity.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/AbstractTestCommentManager.java
```diff
@@ -104,8 +104,10 @@ public void testCreateComment() {
         try {
             commentManager.createComment(session, comment);
             fail("Creating a comment should have failed !");
-        } catch (IllegalArgumentException e) {
+        } catch (NuxeoException e) {
             // ok
+            assertEquals(404, e.getStatusCode());
+            assertNotNull(e.getMessage());
```
kevinleturc: Is it possible to assert the message against a string ?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/impl/CommentJsonWriterTest.java
```diff
@@ -0,0 +1,226 @@
...
+    @Test
+    public void shouldWriteDefaultPropertiesWhenNoFetchersAreProvided() throws IOException {
+        JsonAssert json = jsonAssert(comment);
+        verify(commentManager, never()).getComments(any(CoreSession.class), anyString(), eq(1L), eq(0L), eq(false));
+        assertCommentProperties(json);
+    }
+
+    @Test
+    public void shouldWriteCompleteRepliesSummaryWhenRepliesFetcherIsProvided() throws IOException {
+        PartialList<Comment> returnedComments = new PartialList<>(Lists.reverse(replies), replies.size());
```
kevinleturc: Can you use `Collections#reverse` instead ? We're trying to not use Guava.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/impl/TestPropertyCommentManager.java
```diff
@@ -0,0 +1,399 @@
...
+import org.nuxeo.ecm.platform.comment.api.CommentImpl;
+import org.nuxeo.runtime.test.runner.Deploy;
+
+/**
+ * @since 10.3
+ */
+@Deploy("org.nuxeo.ecm.platform.query.api")
+public class TestPropertyCommentManager extends AbstractTestCommentManager {
+
+    @Test(expected = NuxeoException.class)
```
kevinleturc: Testing exception case is really good, but `NuxeoException` can be thrown in many places and so this assertion is not strong enough. Furthermore don't you add/rework such tests in `AbstractTestCommentManager` ?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/resources/OSGI-INF/comment-pageprovider-contrib.xml
kevinleturc: > To double-check, the service has
> 
> ```
>         GET_COMMENTS_FOR_DOC_PAGEPROVIDER_NAME, singletonList(new SortInfo("dc:created", true)), pageSize,
>                 currentPageIndex, props, documentId);
> ```
> so this must be ineffective. What's the difference between `dc:created` and `comment:creationDate` btw?

This is legacy from former service.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-api/src/main/java/org/nuxeo/ecm/platform/comment/api/CommentManager.java
nmpcunha: JSF UI and WebUI use different sort orders (ascending and descending, respectively)

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/impl/TestPropertyCommentManager.java
nmpcunha: Yes, you can check the details of the exception (error code and message).

The other class (AbstractTestCommentManager) also has unit tests, but I've changed the minimum possible there in order to reduce the scope of this change and avoid compatibility issues with other CommentManager implementations.
Anyway, I would be happy if we can rework that in order to have separated tests per scenario and, probably, per implementation. 👍

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-api/src/main/java/org/nuxeo/ecm/platform/comment/api/CommentManager.java
```diff
@@ -186,6 +187,17 @@
      */
     List<Comment> getComments(CoreSession session, String documentId);
 
+    /**
+     * Gets all comments for a document.
+     *
+     * @param session the core session
+     * @param documentId the document id
+     * @param sortAscending whether to sort ascending or descending
```
efge: Please explain what's the sort criterion (I assume it's the comment date but it should be made explicit)

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-api/src/main/java/org/nuxeo/ecm/platform/comment/api/AnnotationService.java
```diff
@@ -44,37 +46,40 @@
...
-    Annotation getAnnotation(CoreSession session, String annotationId);
+    Annotation getAnnotation(CoreSession session, String annotationId) throws NuxeoException;
 
     /**
      * Gets all annotations for a document.
      *
      * @param session the core session
      * @param documentId the annotated document id
      * @param xpath the annotated blob xpath in the document
+     * @throws NuxeoException if no annotation was found with the given id
```
efge: Why not just return an empty list? After all there are no comments. Also the "no annotation was found with the given id" part is wrong, as it's the annotations about a given document.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-api/src/main/java/org/nuxeo/ecm/platform/comment/api/AnnotationService.java
```diff
@@ -44,37 +46,40 @@
...
-    void updateAnnotation(CoreSession session, String annotationId, Annotation annotation);
+    void updateAnnotation(CoreSession session, String annotationId, Annotation annotation) throws NuxeoException;
 
     /**
      * Deletes an annotation for a document.
      *
      * @param session the core session
      * @param annotationId the annotation document model id
-     * @throws IllegalArgumentException if no annotation was found with the given id
+     * @throws NuxeoException if no annotation was found with the given id
```
efge: It shouldn't throw at all if nothing was deleted.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-api/src/main/java/org/nuxeo/ecm/platform/comment/api/AnnotationService.java
```diff
@@ -44,37 +46,40 @@
      *
      * @param session the core session
      * @param annotationId the annotation document model id
+     * @throws NuxeoException if no annotation was found with the given id
      * @return the annotation, or {@code null} if the annotation with the given id was not found
      */
-    Annotation getAnnotation(CoreSession session, String annotationId);
+    Annotation getAnnotation(CoreSession session, String annotationId) throws NuxeoException;
```
efge: `NuxeoException` is a runtime and generic exception, there's no point in mentioning it as thrown. I don't like that the contract of the method rely on a generic exception. If you really want to specify something thrown, with application-specific meaning, then it should be a dedicated subclass of `NuxeoException`, like a new `AnnotationNotFoundException`.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-rest-api/src/test/java/org/nuxeo/ecm/restapi/server/jaxrs/comment/CommentAdapterTest.java
```diff
@@ -110,6 +110,14 @@ public void testCreateComment() throws IOException {
         }
     }
 
+    @Test
+    public void testGetCommentsForUnexistingDocument() {
```
efge: Unexisting -> Nonexistent everywhere.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-api/src/main/java/org/nuxeo/ecm/platform/comment/api/AnnotationService.java
nmpcunha: Mixed feelings about that... Assuming that annotations are similar to comments, we want to return a 404 when we try to delete a comment for a non existing document.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentJsonWriter.java
kevinleturc: Just found strange to call a service from a writer. Is it something we want ?

guirenard:
:+1:

guirenard:
:+1:

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentJsonWriter.java
funshodavid: Indeed, since annotation is an extension of comment it shares common code.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/PropertyCommentManager.java
```diff
@@ -132,8 +135,8 @@ public void deleteComment(DocumentModel docModel, DocumentModel comment) {
     public DocumentModel getThreadForComment(DocumentModel comment) {
         CoreSession session = comment.getCoreSession();
         DocumentModel parent = session.getDocument(new IdRef((String) comment.getPropertyValue(COMMENT_PARENT_ID)));
-        while (COMMENT_DOC_TYPE.equals(parent.getType())) {
-            comment = getThreadForComment(parent);
+        if (COMMENT_DOC_TYPE.equals(parent.getType())) {
+            return getThreadForComment(parent);
```
funshodavid: Looks like you're fixing something that as wrong, could you explain why ?

funshodavid:
:+1:

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/PropertyCommentManager.java
nmpcunha: It was getting to an infinite cycle with scenarios of several nested replies.
I added a unit test that you can easily change and see that situation ;)

PR #2529, opened 2018-11-13
NXP-25809: filter null values for files:files properties in resolver

funshodavid:


kevinleturc:
👍

efge:
I don't like the fact that we hardcode `file` / `files`. Customers or even Nuxeo could have other schemas that contain blobs. Can't we be more generic?

funshodavid:
T&P : https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-fdavid-master/504/

efge:
👍

PR #2588, opened 2018-11-27
Fix nxp 26248 stream sh monitor

bdelbosc:
T&P ok (failing on platform-3d due to npm event-stream, now fixed in master)


nuxeo-runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/tools/command/LatencyMonitorComputation.java
```diff
@@ -0,0 +1,131 @@
...
+
+    protected final String host;
+
+    protected final int port;
+
+    protected final boolean udp;
+
+    protected GraphiteSender graphite;
+
+    protected String basePrefix;
```
efge: Can be `final` too

kevinleturc:
👍

PR #2595, opened 2018-11-28
Fix nxp 26246 bulk status with error

bdelbosc:


kevinleturc:
Shouldn't we have a list of error messages ?

kevinleturc:
👍

bdelbosc:
> Shouldn't we have a list of error messages ?
maybe but it need to be limited and I am not sure if it is useful to have the 5 firsts messages if you have thousands of errors, the intent of the message is more to give a short feedback to the user than to be able to investigate problem.

PR #2608, opened 2018-11-30
NXP-26357: ignore random failing test

pierre-gautier:


nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestSetPropertiesAction.java
```diff
@@ -168,6 +169,7 @@ public void testSetPropertiesFailures() throws Exception {
     }
 
     @Test
+    @Ignore
```
jcarsique: Use `@Ignore("NXP-26357")` if `@RandomBug.Repeat(issue="NXP-26357")` does not fit better.

https://doc.nuxeo.com/corg/unit-testing/
https://doc.nuxeo.com/corg/random-tests/

PR #2628, opened 2018-12-05
Fix login screen translation

rlabolle:
Login screen translation is currently being broken by LoginScreenConfig extensions without a <supportedLocales> section.

Indeed, the merge() function consider an empty array of supportedLocales as a which to remove any item previously configured.
This commit fix this behaviour.
I also added a commit to avoid adding "null" to the returned list of supportedLocales when the default local is not defined.


atchertchian:
thanks!

atchertchian:
thanks!

atchertchian:
Actually already fixed by https://jira.nuxeo.com/browse/NXP-23786, see d1e260f6e09eeaa59e0c1d7e26d87c5429063ba6, many thanks for the PR!

PR #2674, opened 2018-12-28
fix-NXP-25187 when using UserMapper

cesaralves:
NXP-25187: fix permissions exception on CreateUser


efge:
Thanks. Would you mind changing the commit message to just `NXP-25187: fix permissions exception on CreateUser`?

efge:
Thanks, this has been merged for 10.10 with #2694

PR #2693, opened 2019-01-07
NXP-26497: nuxeoctl.bat search for tools.jar only before java 9

pierre-gautier:


nuxeo-distribution/nuxeo-server-tomcat/src/main/resources/tomcat/bin/nuxeoctl.bat
```diff
@@ -207,6 +202,14 @@ if "%JAVA_VERSION%" lss "%REQUIRED_JAVA_VERSION%" (
   timeout /t 30
   goto END
 )
+set JAVA_VERSION_TOOLS=9
+if "%JAVA_VERSION%" geq "%JAVA_VERSION_TOOLS%" (
```
jcarsique: greater or equal than; didn't you want to do the opposite lesser than?

pierre-gautier: Yes, but I fall onto an issue when realizing it, the comparison is made on strings, even on java required version.
It implies if required is 9 and actual is 11, it will not work.
I'm adapting the script

pierre-gautier:
up

pabgrall:
👍 
There remains the fact that when requiring JDK 8+, we also implicitly accept the usage of JDK  9 and JDK 10 although we do not support them, AFAIK.
Should this be the purpose of another NXP?

pierre-gautier:
@pabgrall As discussed this has always been the case and is not related to the bug that has to be fixed.

It could be the purpose of an another ticket but I don't see the point forbidding java versions that opensource/free users may want to use

jcarsique:
:+1:

PR #2711, opened 2019-01-10
NXP-23943: do not raise exception from check_branch_to_release()

jcarsique:
The check_branch_to_release() method is not compliant with the Package maintenance branch naming.
Waiting for NXP-21008 and a better solution for NXP-23943, this commit makes the check not blocker:
it logs a [WARN] instead of raising an exception.


kevinleturc:
👍

PR #2722, opened 2019-01-14
NXP-26642: fix NPE on concurrency

efge:


nuxeo-core/nuxeo-core-cache/src/main/java/org/nuxeo/ecm/core/transientstore/keyvalueblob/KeyValueBlobTransientStore.java
```diff
@@ -383,7 +381,14 @@ public void putParameters(String key, Map<String, Serializable> parameters) {
             }
         }
         // get values
-        return parameters.stream().collect(Collectors.toMap(identity(), p -> getParameter(key, p)));
+        Map<String, Serializable> map = new HashMap<>();
```
covolution: I am not keen on using a type as a variable name, i.e. `map`, but its not a big deal.  Perhaps `values` would have been more appropriate?

efge: Yeah, I would have used `parameters` but it was already taken by the list so I took the easy path.

PR #2732, opened 2019-01-17
NXP-26607: enable long processing

bdelbosc:
Increasing the poll interval so long computation don't block the processing.
Use 1h limit and poll only 2 records.


kevinleturc:
:+1:

efge:
👍

PR #2773, opened 2019-01-31
NXP-26637: introduce a specific handler for user and group aggregate …

guirenard:
…widget to be able to resolve prefixed ids

### DIFF
--- a/nuxeo-jsf/nuxeo-platform-contentview-jsf/src/main/java/org/nuxeo/ecm/platform/contentview/jsf/facelets/plugins/SelectManyCheckboxUserAndGroupAggregateWidgetTypeHandler.java
+++ b/nuxeo-jsf/nuxeo-platform-contentview-jsf/src/main/java/org/nuxeo/ecm/platform/contentview/jsf/facelets/plugins/SelectManyCheckboxUserAndGroupAggregateWidgetTypeHandler.java
@@ -1,0 +1,49 @@
+/*
+ * (C) Copyright 2019 Nuxeo SA (http://nuxeo.com/) and others.
+ *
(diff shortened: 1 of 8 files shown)


guirenard:
T&P will follow if 1st review is ok

atchertchian:
:+1: LGTM!

A few comments:
- For UIUserAndGroupSelectItems, it seems like a lot of code is duplicated, i assume you do not really have a choice (just in case there is another way)
- the fact that the select item labels are hardcoded removes a configuration possibility (for people who who'd like to customize and show the user email for instance) but maybe it's overkill to refine this -- note we could use the methods that have been added for this on some templates, see https://github.com/nuxeo/nuxeo/blob/release-10.10/nuxeo-jsf/nuxeo-platform-webapp-base/src/main/resources/web/nuxeo.war/widgets/user_group_display_template.xhtml#L39 but also https://github.com/nuxeo/nuxeo/blob/release-10.10/nuxeo-jsf/nuxeo-platform-webapp-base/src/main/resources/web/nuxeo.war/widgets/user_group_display_template.xhtml#L43, as well as https://github.com/nuxeo/nuxeo/blob/release-10.10/nuxeo-jsf/nuxeo-platform-webapp-base/src/main/resources/web/nuxeo.war/widgets/user_group_display_template.xhtml#L75 for groups. These could be exposed as "default expressions" on the widget type configuration if you agree
- as previously discussed, this widget could be added to the layout-demo showcase (for reference and functional tests)

atchertchian:
(above changes validated by T&P)

guirenard:
:+1:

troger:
:+1:

PR #2777, opened 2019-02-04
Nxp 24314 disable wms from properrty

pierre-gautier:
https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-pierregautier/233/

### DIFF
--- a/nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkManagerImpl.java
+++ b/nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkManagerImpl.java
@@ -104,6 +104,11 @@ public class WorkManagerImpl extends DefaultComponent implements WorkManager {
      */
     public static final String SHUTDOWN_DELAY_MS_KEY = "nuxeo.work.shutdown.delay.ms";
 
+    /**
+     * @since 11.1
+     */
+    public static final String WORKMANAGER_PROCESSING_DISABLE = "nuxeo.work.processing.disable";
+
     protected final MetricRegistry registry = SharedMetricRegistries.getOrCreate(MetricsService.class.getName());
 
     // used synchronized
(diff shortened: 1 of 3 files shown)


nuxeo-core/nuxeo-core-cache/src/main/java/org/nuxeo/ecm/core/cache/CacheServiceImpl.java
```diff
@@ -189,7 +189,7 @@ public int getApplicationStartedOrder() {
     public void start(ComponentContext context) {
         super.start(context);
         ClusterService clusterService = Framework.getService(ClusterService.class);
-        if (clusterService.isEnabled()) {
+        if (clusterService != null && clusterService.isEnabled()) {
```
efge: Please use NXP-25499 for this fix.
But the fix should rather be to add the proper `@Deploy("org.nuxeo.runtime.cluster")` to the test that fails (until we have cleaned up all our features to have one feature per test service).

pierre-gautier: Ok for the NXP-25499
For the fix, I do not agree, it's a potential NPE if clusterService is not deployed tests or not, and apparently it's not a blocking case

efge: ClusterService is a basic runtime service that is required by pubsub and therefore for invalidations of caches, directories and repositories. It should always be deployed before the cache service.

efge: If we had proper dependencies between services we would make this explicit with injection or requirements...

kevinleturc:
👍

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkManagerImpl.java
```diff
@@ -104,6 +104,11 @@
      */
     public static final String SHUTDOWN_DELAY_MS_KEY = "nuxeo.work.shutdown.delay.ms";
 
+    /**
+     * @since 11.1
+     */
+    public static final String WORKMANAGER_PROCESSING_DISABLE = "nuxeo.work.processing.disable";
```
funshodavid: `disabled` instead of `disable` ?

pierre-gautier:
👍

PR #2782, opened 2019-02-06
NXP-26760: add user to BulkCommand.Builder and check it's not null

pierre-gautier:
https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-pierregautier3/54/

### DIFF
--- a/nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkCommand.java
+++ b/nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkCommand.java
@@ -47,7 +47,6 @@ public class BulkCommand implements Serializable {
 
     protected String query;
 
-    @Nullable
     protected String username;
 
     @Nullable
@@ -159,7 +158,32 @@ public class BulkCommand implements Serializable {
          *
          * @param action the registered bulk action name
          * @param nxqlQuery the query that represent the document set to apply the action
+         * @param user the user with whose rights the computation will be executed
+         * @since 11.1
          */
+        public Builder(String action, String nxqlQuery, String username) {
+            if (isEmpty(action)) {
+                throw new IllegalArgumentException("Action cannot be empty");
+            }
+            this.action = action;
+            if (isEmpty(nxqlQuery)) {
+                throw new IllegalArgumentException("Query cannot be empty");
+            }
+            this.query = nxqlQuery;
+            if (isEmpty(username)) {
+                throw new IllegalArgumentException("Username cannot be empty");
+            }
+            this.username = username;
+        }
+
+        /**
+         * BulkCommand builder
+         *
+         * @param action the registered bulk action name
+         * @param nxqlQuery the query that represent the document set to apply the action
+         * @deprecated since 11.1, use {@link #Builder(String, String, String)} constructor with username instead
+         */
+        @Deprecated
         public Builder(String action, String nxqlQuery) {
             if (isEmpty(action)) {
                 throw new IllegalArgumentException("Action cannot be empty");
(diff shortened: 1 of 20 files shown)


covolution:
Is this a breaking change? ie. does this add a new mandatory field to the builder?  Is so, it will break the ai code.

pierre-gautier:
I'll patch AI modules as well

kevinleturc:
To follow what's @covolution said, we should probably deprecate former constructor and `user` method as the code is now in a LTS. @efge what do you think about it ? is it a part of platform under backward compatibility policy ?

troger:
As we are going on continuous delivery, almost everything of the platform should be backward compatible IMO (if possible).

covolution:
If the validation logic was in the `build()` method rather than the constructor then, for AI, it wouldn't be a breaking change.  The code currently sets the `user` just not in the constructor.

pierre-gautier:
It's set then, I'll keep the former constructor and deprecate it

pierre-gautier:
updated

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkCommand.java
```diff
@@ -159,7 +158,31 @@ public void setRepository(String repository) {
          *
          * @param action the registered bulk action name
          * @param nxqlQuery the query that represent the document set to apply the action
+         * @param user the user with whose rights the computation will be executed
          */
```
funshodavid: missing `@since 11.1`

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkCommand.java
```diff
@@ -159,7 +158,31 @@ public void setRepository(String repository) {
...
+            }
+            this.username = username;
+        }
+
+        /**
+         * BulkCommand builder
+         *
+         * @param action the registered bulk action name
+         * @param nxqlQuery the query that represent the document set to apply the action
+         * @deprecated 11.1 use constructor with username instead
```
funshodavid: `@deprecated 11.1 use constructor with username instead` -> `@deprecated since 11.1`, use {@link #Builder(String, String, String)} constructor with username instead.

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkCommand.java
```diff
@@ -181,7 +204,10 @@ public Builder repository(String name) {
 
         /**
          * User running the bulk action
+         *
+         * @deprecated 11.1 use constructor with username instead
```
funshodavid: `@deprecated 11.1 use constructor with username instead` -> `@deprecated since 11.1`, use {@link #Builder(String, String, String)} constructor with username instead.

nuxeo-features/nuxeo-automation/nuxeo-automation-features/src/main/java/org/nuxeo/ecm/automation/core/operations/services/bulk/BulkRunAction.java
```diff
@@ -37,6 +37,7 @@
 import org.nuxeo.ecm.core.bulk.BulkService;
 import org.nuxeo.ecm.core.bulk.io.BulkParameters;
 import org.nuxeo.ecm.core.bulk.message.BulkCommand;
+
```
funshodavid: unnecessary empty line

funshodavid:
:+1:

efge:
👍

PR #2818, opened 2019-02-21
NXP-26137: Handle multiple packages to upload

alexistimic:
Handle multiple packages to upload


jcarsique:
:+1: 
could we enrich a test to cover that?

scripts/release_mp.py
```diff
@@ -292,11 +292,13 @@ def upload(self, url, marketplace, dryrun=False, owner=None):
         """ Upload the given Marketplace package and update the config file."""
         uploaded = [url + ":"]
         mp_to_upload = self.mp_config.get(marketplace, "mp_to_upload")
-        for pkg in glob.glob(mp_to_upload):
-            if os.path.isfile(pkg):
-                retcode = self.upload_file(url, pkg, dryrun=dryrun, owner=owner)
-                if retcode == 0:
-                    uploaded.append(os.path.realpath(pkg))
+        mp_to_upload = mp_to_upload.replace(" ", "")
```
jcarsique: Why? Could you add a test to highlight the need to strip? 
If needed, why replace versus strip?
https://www.programiz.com/python-programming/methods/string/strip

scripts/tests/test_release_mp.py
```diff
@@ -0,0 +1,55 @@
...
+        assert release.get_packages_list() == ["nuxeo-ai"]
+
+        # Use fake values to prevent bad substitution and ease testing
+        release.mp_config.set("DEFAULT", "branch", "master")
+        release.mp_config.set("DEFAULT", "is_final", "True")
+        release.mp_config.set("DEFAULT", "other_versions", "")
+
+        # Set several packages to upload
+        packages = ["*/target/nuxeo-ai*.zip", "*/*/target/nuxeo-*.zip"]
+        release.mp_config.set("nuxeo-ai", "mp_to_upload", ",".join(packages))
```
jcarsique: `packages = "*/target/nuxeo-ai*.zip, */*/target/nuxeo-*.zip"` for instance (see above remark)

scripts/release_mp.py
BoboTiG: We need to `replace` because we could have a mix of coma + space or just coma:
```
*/target/a-*.zip, */*/target/b-*.zip,*/target/c-*.zip
```
This is only to be less strict and allow both syntaxes.

scripts/tests/test_release_mp.py
BoboTiG:
```suggestion
        release.mp_config.set("nuxeo-ai", "mp_to_upload", ", ".join(packages))
```

scripts/release_mp.py
BoboTiG: We can remove that line and use `strip()` in the next-next line:
```python
for pkg in glob.glob(mp.stip()):
```

scripts/tests/test_release_mp.py
alexistimic: :+1:

jcarsique:
:+1:

PR #2821, opened 2019-02-21
Added all dublin core properties as constants

kant-nxo:


efge:
The commit contains changes unrelated to dublincore constants.
Also please create a NXP ticket describing issue and the reason for the changes, and include the NXP-xyz number at the start of the commit message.

kant-nxo:
Bad commit in a hurry.
NXP: https://jira.nuxeo.com/browse/NXP-26855 ; the component might be wrong.

kant-nxo:
Fixed, forced pushed.

nuxeo-services/nuxeo-platform-dublincore/src/main/java/org/nuxeo/ecm/platform/dublincore/constants/DublinCoreConstants.java
```diff
@@ -32,28 +33,72 @@ private DublinCoreConstants() {
 
     public static final String DUBLINCORE_SCHEMA_NAME = "dublincore";
 
+    public static final String DUBLINCORE_CONTRIBUTORS_PROPERTY = "dc:contributors";
```
efge: Please add JavaDoc `/** @since 11.1 */` on all new constants.

nuxeo-services/nuxeo-platform-dublincore/src/main/java/org/nuxeo/ecm/platform/dublincore/constants/DublinCoreConstants.java
```diff
@@ -32,28 +33,72 @@ private DublinCoreConstants() {
...
+
+    public static final String DUBLINCORE_PUBLISHER_PROPERTY = "dc:publisher";
+
+    public static final String DUBLINCORE_CONTRIBUTORS = "contributors";
+
+    public static final String DUBLINCORE_CREATED_DATE = "created";
 
     public static final String DUBLINCORE_CREATOR = "creator";
 
+    public static final String DUBLINCORE_DESCRIPTION = "description";
```
efge: Usage of the "bare" properties (without prefix) is really not a good idea, people should always use full property names like "dc:description". So I'd rather not have these new constants for things we don't want in the code.

kant-nxo: I agree with it, just tried to be consistent with the existing

nuxeo-services/nuxeo-platform-dublincore/src/main/java/org/nuxeo/ecm/platform/dublincore/constants/DublinCoreConstants.java
kant-nxo: ok np

nuxeo-services/nuxeo-platform-dublincore/src/main/java/org/nuxeo/ecm/platform/dublincore/constants/DublinCoreConstants.java
kant-nxo: In some cases (but no idea about the reason), you could use doc.getProperty(DUBLINCORE_SCHEMA_NAME, DUBLINCORE_CREATOR)

efge: Yes that's the "old style". We're trying to discourage its usage. Although there hasn't been a task force to clean up our code yet.

kant-nxo: ok done

efge:
Please squash the two commits.

efge:
👍

nuxeo-services/nuxeo-platform-dublincore/src/main/java/org/nuxeo/ecm/platform/dublincore/constants/DublinCoreConstants.java
nmpcunha: It was added for coherency purposes with the existing ones.
Should we remove them or just mark as @deprecated, @efge?

nmpcunha:
Despite the question added to the thread, 👍

nuxeo-services/nuxeo-platform-dublincore/src/main/java/org/nuxeo/ecm/platform/dublincore/constants/DublinCoreConstants.java
efge: If we deprecate them we should first cleanup our own codebase, so I think it's a separate (but related) issue.

efge:
👍

PR #2861, opened 2019-03-06
Nxp 26777 baf idempotency support

pierre-gautier:


nuxeo-runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/computation/FilteringPolicy.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.nuxeo.lib.stream.computation;
+
+/**
+ * @since 11.1
+ */
+public interface FilteringPolicy {
+
+    default boolean shouldSkip(String key, ComputationContext context) {
```
kevinleturc: We shouldn't have `default` on this method as it's the only contract to implement for this interface.

nuxeo-runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/log/internals/InMemoryLogOffsetStorage.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.nuxeo.lib.stream.log.LogOffsetStorage;
+import org.nuxeo.lib.stream.log.LogOffset;
+
+/**
+ * @since 11.1
+ */
+public class InMemoryLogOffsetStorage implements LogOffsetStorage {
+
+    protected Map<String, LogOffset> map = new HashMap<>();
```
kevinleturc: We probably need a `ConcurrentHashMap` here?

nuxeo-runtime/nuxeo-stream/src/test/java/org/nuxeo/lib/stream/tests/computation/TestStreamProcessor.java
```diff
@@ -28,7 +28,6 @@
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collections;
-
```
kevinleturc: Should be kept.

nuxeo-runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/computation/FilteringPolicy.java
pierre-gautier: Would you prefer a concrete class with default implementation ?

PR #2883, opened 2019-03-26
NXP-27075: set Eclipse old 1.8 compliance

jcarsique:
Upgrade to org.apache.maven.plugins:maven-eclipse-plugin:2.10
Set org.eclipse.jdt.core.prefs 1.8 compiler compliance


pierre-gautier:
since we don't want to follow this path, I believe this PR can be closed
@jcarsique

PR #2893, opened 2019-03-28
NXP-27103 / NXP-27109: key value blob transient store blobprovider default

efge:
https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/656/

### DIFF
--- a/nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobManager.java
+++ b/nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobManager.java
@@ -67,8 +67,23 @@ public interface BlobManager {
      * @param id the blob provider id or namespace
      * @return the blob provider
      * @since 10.10
+     * @deprecated since 11.1, use {@link #getBlobProviderWithNamespace(String, String)} instead
      */
-    BlobProvider getBlobProviderWithNamespace(String id);
+    @Deprecated
+    default BlobProvider getBlobProviderWithNamespace(String id) {
+        return getBlobProviderWithNamespace(id, "default");
+    }
+
+    /**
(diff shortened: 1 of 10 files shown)


kevinleturc:
👍

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobManager.java
```diff
@@ -67,8 +67,23 @@
      * @param id the blob provider id or namespace
      * @return the blob provider
      * @since 10.10
+     * @deprecated since 11.1, use {@link #getBlobProviderWithNamespace(String, String)} instead
      */
-    BlobProvider getBlobProviderWithNamespace(String id);
+    @Deprecated
+    default BlobProvider getBlobProviderWithNamespace(String id) {
+        return getBlobProviderWithNamespace(id, "default");
```
troger: Why not using (moving) the `DEFAULT_ID` constant?

troger:
:+1:

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobManager.java
efge: Well if we define a constant it would be only used by the deprecated method... I'd rather leave the constants about defaults in whatever is using the API (`KeyValueBlobTransientStore.CONFIG_DEFAULT_BLOB_PROVIDER_DEFAULT`).

troger: Makes sense. So maybe we can deprecate `BlobManagerComponent#DEFAULT_ID` 😬

PR #2900, opened 2019-04-01
NXP-27075: adapt poms to java 11 unique visibily constraint

pierre-gautier:
- remove xml-apis
- remove jaxen
- bump xom
- replace jta
- patch webdav-jaxrs
- exclude affinity
- exclude jsr173_api

bonus : remove NuxeoNature maven-eclipse configuration


pierre-gautier:
Early PR for review, T&P is failing for now

efge:
Commit message has "jwxrs" instead of "jaxrs"

nuxeo-runtime/nuxeo-stream/pom.xml
```diff
@@ -18,6 +18,16 @@
       <groupId>net.openhft</groupId>
       <artifactId>chronicle-queue</artifactId>
     </dependency>
+    <dependency>
+      <groupId>net.openhft</groupId>
+      <artifactId>chronicle-threads</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>net.openhft</groupId>
+          <artifactId>affinity</artifactId>
```
efge: Please add a comment explaining why we exclude it.  Also it's probably better to put this exclusion in the toplevel POM.

nuxeo-services/nuxeo-scim-server/pom.xml
```diff
@@ -113,12 +113,24 @@
       <groupId>info.simplecloud</groupId>
       <artifactId>scimcore</artifactId>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>javax.xml.bind</groupId>
+          <artifactId>jsr173_api</artifactId>
```
efge: Same comment as for `affinity`: explain why and move to toplevel POM.

nuxeo-runtime/nuxeo-stream/pom.xml
pierre-gautier: Explained.
In the toplevel POM only chronicle BOM is declared, it seems more relevant excluded down there

pierre-gautier:
updated to review

nuxeo-runtime/nuxeo-stream/pom.xml
pierre-gautier: it appears maven eclipse:eclipse does not support the case where the bom is imported and a dependency from it is declared alongside, I moved it back to stream module since the chronicle dependency will not be used elsewhere

ataillefer:
After update, I still get
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-eclipse-plugin:2.9:eclipse (default-cli) on project nuxeo-common: Execution default-cli of goal org.apache.maven.plugins:maven-eclipse-plugin:2.9:eclipse failed: For artifact {net.openhft:chronicle-threads:null:jar}: The version cannot be empty.
```
when running `mvn eclipse:eclipse` on nuxeo-commons, as we had noticed together.

ataillefer:
Which is now fixed by https://github.com/nuxeo/nuxeo/pull/2900#discussion_r271657644.

ataillefer:
As seen together, there are some issues with the `xercesImpl` lib that exposes the `org.w3c.dom` package via `org/w3c/dom/html/HTMLDOMImplementation.class`.

Thus the error message
```
The package org.w3c.dom is accessible from more than one module: <unnamed>, java.xml
```
in `DomHelper` for instance.

Strangely, it is not reported in the Package Explorer nor in the Problems view.

pierre-gautier:
close PR as the content has too much changed, a TP will create another fresh one

PR #2934, opened 2019-04-05
NXP-26168 Swagger Specification: convertAdapter.ftl

dbnuxeo:
Swagger Specification: convertAdapter.ftl definition contains multiple errors

The convertAdapter.ftl specification contains missing references and a duplicate entry.

'blobXPath' parameter is not defined (lines: 22, 54)
Line 54 path should begin with '/id/' instead of '/path/'. Entry is duplicate of line 6.

Parameter and file names follow the convention for the doc & paramtypes directories.


ataillefer:
:+1:

troger:
:+1:

PR #2938, opened 2019-04-05
NXP-26166: invalid parameter reference in path

dbnuxeo:
The definition in 'paramtypes/username.ftl' defines 'userName'. 'oauth2.ftl' path references are broken, they should be changed from 'username' to 'userName'. Other endpoints use this paramtype - oauth2.ftl is the only template that incorrectly uses this parameter.


ataillefer:
:+1:

troger:
:+1:

PR #2975, opened 2019-04-15
NXP-27234: cleanup, format & good practices

jcarsique:
Cleanup and review for consistency between pipelines
    
    Move up warningsPublisher() in the thread stack
    Remove catch Throwable set for debug purpose (if useful, then document and add everywhere)
    Use Blue Ocean URLs
    Add Groovy shebang
    Line return at 120 for readability sh main command and its parameters
    Use Shell "-e" (exit on error) mode


kevinleturc:
👍

alexistimic:
:+1:

PR #3060, opened 2019-05-16
NXP-27386: fix Redis pubsub subscriber reentrancy to Redis

efge:
https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/690/


troger:
I'm wondering, looking at the diff, for each `onMessage` we are using a thread from the pool. Wouldn't that be more efficient to have a separate thread for the subscribe part but still use the same thread for the publishing? I would think we are doing less subscribe than publish.

Not sure it's possible or better...

bdelbosc:
:+1:

bdelbosc:
@troger there will be only one thread created used for the onMessage, the thread will be cached for 60s

troger:
:+1:

ataillefer:
:+1:

PR #3079, opened 2019-05-23
NXP-27110 adding NOT LIKE and NOT ILIKE operators to NXQL builder

andreinechaev:


kevinleturc:
Could you run a T&P on all branches?

andreinechaev:
sure

PR #3111, opened 2019-06-04
Improve Nuxeo Platform Rendering to add Map and List in templateParam to populate the Templating Context

bjalon:
@aescaffre : Here is the contribution, I told you... This evolution was needed by Clockworkx to add any complexe templateParam structure. They financed this evolution.

In this pull request, we added the possibility to contribute a complexe structure as templateParam.
For instance :

```
<fields xmlns:dt="http://www.nuxeo.org/DocumentTemplate">
  <dt:field name="account" type="Map">
    <dt:field name="owner" type="Map">
      <dt:field name="firstname" type="String" value="Benjamin"/>
      <dt:field name="lastname" type="String" value="JALON"/>
    </dt:field>
    <dt:field name="type" type="String" value="checking"/>
    <dt:field name="operations" type="list">
      <dt:field name="op1" type="Map">
        <dt:field name="date" type="Date" value="2017-07-14 13:14:15.678"/>
        <dt:field name="value" type="String" value="$ 200.20"/>
        <dt:field name="isPositive" type="Boolean" value="true"/>
      </dt:field>
      <dt:field name="op2" type="Map">
        <dt:field name="date" type="Date" value="2017-07-15 13:10:15.678"/>
        <dt:field name="value" type="String" value="-$ 40.27"/>
        <dt:field name="isPositive" type="Boolean" value="false"/>
      </dt:field>
    </dt:field>
  </dt:field>
</fields>
```

And so the following ftl template filled :

```
Hello ${account.owner.firstname} ${account.owner.lastname},
Here are the operations of the account ${account.number} for this month:
            DATE          |   VALUE  | IS POSITIVE
 <#list account.operations as operation>
 ${operation.date?datetime} | ${operation.value} | ${operation.isPositive?string("yes", "no")}
 </#list>
```

I did a COMMAND + SHIFT + F on the nuxeo-template-rendering directory on the 41bd2941fd4acd1190d6eea014e5ca574727158d changeset if this is too much.

If I'm right I didn't change any behaviour after this contribution. I tried to keep my commit clear. In  detail, here is what I did :
* I refactor a lot AbstractBindingResolver
* I deprecated XMLSerializer static class and move the implementation into the XMLSerializer contribution to a new extension point. I hesitated to move this logic into a converter, but I finally choose to create a specific contribution
* I added a lot of test over the code I modified
* I refactored also TemplateType

Evolution are essentially AbstractBindingResolver / XMLSerializer / TemplateType :
* In AbstractBindingResolver, I manage the case when a TemplateType is a List or Map and I created the corresponding Object that will be added to the BindingResolver implementation (Freemarker, XDocReport, ...)
* In XML Serializer, I manage List and Map to bind it to a new both TemplateType List and Map

That's it. Hoper you'll love this contribution that I did with a lot of heart :) Have fun!

Benjamin


efge:
Hi @bjalon 

Could you please cleanup your 24 commits to squash together related things? As it is, it's nearly impossible to review. Also you should avoid formatting commits, as it makes things that much harder to comprehend.

bjalon:
@efge : I squashed my commits and reword to be a bit more explicit. Can you now review this evolution ? I'm ready to cleanup more :)

Just to summarize a bit this commit :
* First I add more tests as part of the template rendering code that will be changed for my evolution  was not unit tested. And I wanted to be sure that behavior will not change with this evolution.
* I added a SerializerService that will expose Serializers that can serialize and deserialize from JSON or XML to TemplateInput Objects
  * Serialization is used during the template rendering context generation and in the automation operation
  * I plan to add the JSON implementation later as we will need for our customer project. Maybe the JSON serializer should not be there or should return an UnimplementedException. Tell me... 
  * The implementation of the XML is a simple copy of the code XMLSerializer static class. Refactoring is done in the next commits
  * The XMLSerializer static class now used the XMLSerializer contribution of the SerializerService.
  * I deprecated XMLSerializer static class
* Then I changed all XMLSerializer static class usage by the XMLSerializer contribution of the SerializerService
* Then I refactored the doDeserialization method of the XMLSerializer contribution of the SerializerService (no behavior modification) that will help me during the evolution
* I added the map and list management in the doDeserialization method of the XMLSerializer contribution of the SerializerService
* Then I refactored the doSerialization method of the XMLSerializer contribution of the SerializerService (no behavior modification) that will help me during the evolution
* I added the map and list management in the doSerialization method of the XMLSerializer contribution of the SerializerService
* Then I refactored the AbstractBindingResolver that get the TemplateInput Object collected to generate the Objects in the Template context (without behavior modification)
* Then I added the Map and List management
* Finally I add an unexpected regression. I decided to let it appears as this behavior is a bit strange. I can squash it but you can also preserve me :)

Benjamin

bjalon:
@efge Do you want me to reword commits to add "NXP-27613" as prefix ?

bjalon:
Prefix added.

addons/nuxeo-template-rendering/nuxeo-template-rendering-api/src/main/java/org/nuxeo/template/api/TemplateInput.java
```diff
@@ -21,7 +21,9 @@
 package org.nuxeo.template.api;
 
 import java.io.Serializable;
-import java.util.Date;
+import java.util.*;
```
efge: Please don't use any wildcard import

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/XMLSerializer.java
```diff
@@ -20,143 +20,38 @@
...
 
 /**
- * {@link TemplateInput} parameters are stored in the {@link DocumentModel} as a single String Property via XML
- * Serialization. This class contains the Serialization/Deserialization logic.
+ * Should use now Framework.getService(SerializerService.class).getSerializer("xml")
  *
  * @author Tiry (tdelprat@nuxeo.com)
+ * @author bjalon (bjalon@qastia.com)
  */
+@Deprecated(since = "11.1")
```
efge: Also add a `@deprecated` tag in the Javadoc: `@deprecated since 11.1, use XYZ instead`

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/XMLSerializer.java
```diff
@@ -20,143 +20,38 @@
...
-    public static final Namespace ns = new Namespace(XML_NAMESPACE_PREFIX, XML_NAMESPACE);
-
-    public static final QName fieldsTag = DocumentFactory.getInstance().createQName("templateParams", ns);
-
-    public static final QName fieldTag = DocumentFactory.getInstance().createQName("field", ns);
-
-    public static final String DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS";
+    public static Serializer getXmlSerializer() {
+        if (xmlSerializer == null) {
+            xmlSerializer = Framework.getService(SerializerService.class).getSerializer("xml");
```
efge: Don't store the result of `Framework.getService` in a field. Do the lookup every time (same issue in other classes too)

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/adapters/doc/TemplateBinding.java
```diff
@@ -18,14 +18,16 @@
...
+import org.nuxeo.template.api.TemplateInput;
+import org.nuxeo.template.serializer.executors.Serializer;
+import org.nuxeo.template.serializer.service.SerializerService;
+
 import java.io.Serializable;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import org.nuxeo.template.XMLSerializer;
```
efge: In our formatting style `org` is after `java`, like it was originally in the file.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/automation/RenderWithTemplateOperation.java
```diff
@@ -67,6 +67,18 @@
     @Param(name = "templateData", required = false)
     protected String templateData = null;
 
+    @Param(name = "serializer", required = false)
+    protected String serializerName = null;
```
efge: `= null` is already the default, no need to write it.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/processors/AbstractBindingResolver.java
```diff
@@ -70,106 +72,181 @@ protected DocumentObjectWrapper getWrapper() {
...
+            } catch (ValueNotFound e) {
+                log.warn("Unable to handle binding for param " + param.getName(), e);
+            } catch (NoValueToAddInContext e) {
+                log.debug("Skip param to add " + param.getName(), e);
+            }
+        }
+    }
+
+
+    private Object extractValueFromParam(TemplateBasedDocument templateBasedDocument, TemplateInput param) throws ValueNotFound, NoValueToAddInContext {
```
efge: Don't use double blank lines.
Don't use `private` methods as they prevent subclassing, use `protected`.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/JSONSerializer.java
```diff
@@ -0,0 +1,19 @@
+package org.nuxeo.template.serializer.executors;
```
efge: Missing copyright.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,207 @@
+package org.nuxeo.template.serializer.executors;
```
efge: Missing copyright. If this file (or another) is derived from code copied from a Nuxeo-copyrighted file, then please use the Nuxeo copyright.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public Serializer getSerializer(String name) {
+        SerializerContribution contrib = serializers.get(name);
+
+        if (contrib == null) {
+            contrib = serializers.get(DEFAULT_SERIALIZER_NAME);
+            if (contrib == null) {
+                throw new UnknownSerializer(name);
```
kevinleturc: Could you use a `NuxeoException` instead?

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/test/java/org/nuxeo/template/TestTemplateRenderingService.java
```diff
@@ -0,0 +1,273 @@
...
+
+    protected static final String TEMPLATE_NAME = "mytestTemplate";
+
+    protected static final String WEBVIEW_RENDITION = "webView";
+
+    @Inject
+    protected CoreSession session;
+
+    @Inject
+    TemplateProcessorService tps;
```
kevinleturc: Could you make fields `protected`?

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/test/java/org/nuxeo/template/TestTemplateRenderingService.java
```diff
@@ -0,0 +1,273 @@
...
+        Blob result = templateBase.renderWithTemplate(TEMPLATE_NAME);
+        assertNotNull(result);
+        assertEquals("Hello from MyTemplateBase !", result.getString());
+    }
+
+    @Test
+    public void whenTemplateWithGivenInputTemplate_String_shouldRenderBlobAsIt() throws IOException {
+        TemplateSourceDocument templateSrc = createTemplateSourceDoc("We are introducing ${myStringInCtxt} !", WEBVIEW_RENDITION);
+        TemplateBasedDocument templateBase = createTemplateBasedDoc(templateSrc.getAdaptedDoc());
+        List<TemplateInput> params = Arrays.asList(
```
kevinleturc: Could you use `List#of` or `Collections#singletonList`?

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/JSONSerializer.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.dom4j.DocumentException;
+import org.nuxeo.template.api.TemplateInput;
+
+import java.util.List;
+
+public class JSONSerializer implements Serializer {
```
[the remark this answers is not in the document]
bjalon: Sorry, for the delay I didn't receive any notification about your comments, and thanks for your feedback. As I explained in the documentation attached to this pull request we plan to implement a JSONSerilizer. So Yes, I will remove it.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,207 @@
...
+                Map<String, TemplateInput> listValue = new HashMap<>();
+                for (Element childNode : node.elements()) {
+                    TemplateInput childParam = extractTemplateInputFromXMLNode(childNode);
+                    if (childNode != null) {
+                        listValue.put(childParam.getName(), childParam);
+                    }
+                }
+                paramValue = listValue;
+                break;
+            case MapValue:
```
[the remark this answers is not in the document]
bjalon: You right, sorry I didn't see that during the squash git keep both case. The first one is the wrong one. I'm gonna remove it. Do you prefer to make the modification in the history or create one commit with all fixes or one new commit per modification requested ?

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,207 @@
...
+            case ListValue:
+                LinkedHashMap<String, TemplateInput> listValue = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
bjalon: No because order is important for the map. If the templateData is defined like that :
```
<map>
  <entry key="key1" value="value1"/>
  <entry key="key3" value="value3"/>
  <entry key="key2" value="value2"/>
</map>
```
or like that :
```
<list>
  <entry key="key1" value="value1"/>
  <entry key="key3" value="value3"/>
  <entry key="key2" value="value2"/>
</list>
```
User would like that when he will iterate through the list or the map the order will be the same.
(Think that case above was not the right one - see comment above).

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,207 @@
...
+                break;
+            case DocumentProperty:
+            case PictureProperty:
+            case Content:
+                field.addAttribute("source", param.getSource());
+                break;
+        }
+    }
+
+    private class TemplateInputBadFormat extends Throwable {
```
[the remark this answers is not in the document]
bjalon: Are you sure that I change that... This exception is totally internal to the class and really specific to runtime inside this class... And if I follow the craftsmanship approach, it is a better to create your own exception that is more expressive. But if you prefer I can do that.

bjalon:
@efge @kevinleturc : Do you prefer to have modification in :
* one commit per comment
* one big commit with all fixes
* rewrite history ?

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/XMLSerializer.java
bjalon: Hi,

I don't know how to modify the javadoc. Other point when you say "XYZ", you mean 11.1.0 ? because if I'm looking for most of the depreciation for the 11.X.X you added, you put 11.1 (example [MimetypeIconUpdater.java](https://github.com/nuxeo/nuxeo/blob/457a1b98ae32d1ced0a4de0009b979214da69b52/nuxeo-services/nuxeo-platform-filemanager-core-listener/src/main/java/org/nuxeo/ecm/platform/filemanager/core/listener/MimetypeIconUpdater.java#L65)).

bjalon: do I put for removal ?

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerServiceImpl.java
bjalon: Ok I extends NuxeoException. I don't like to have generic exception. If you are agree...

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/test/java/org/nuxeo/template/TestTemplateRenderingService.java
bjalon: done

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/test/java/org/nuxeo/template/TestTemplateRenderingService.java
bjalon: done

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerServiceImpl.java
bjalon: I extended NuxeoExcpetion. => done

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    private static final String SERIALIZER_EP_NAME = "serializers";
+    private static final String DEFAULT_SERIALIZER_NAME = "default";
+
+    private Map<String, SerializerContribution> serializers = new HashMap<>();
+
+    @Override
+    public void registerContribution(Object contribution, String extensionPoint, ComponentInstance contributor) {
+        if (SERIALIZER_EP_NAME.equals(extensionPoint)) {
+            SerializerContribution sc = (SerializerContribution) contribution;
+            serializers.put(sc.getName(), sc);
```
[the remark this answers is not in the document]
bjalon: Done

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerContribution.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public String getName() {
+        return name == null || name.trim().isEmpty() ? "default" : name;
+    }
+
+    public Serializer getImplementation() {
+        Serializer obj;
+        try {
+            obj = implementationClass.getDeclaredConstructor().newInstance();
+        } catch (NoSuchMethodException e) {
```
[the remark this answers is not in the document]
bjalon: Oups => forgotten... Let me one minute.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerContribution.java
```diff
@@ -0,0 +1,46 @@
...
+public class SerializerContribution {
+
+    @XNode("@class")
+    public Class<Serializer> implementationClass;
+
+    @XNode("@name")
+    public String name;
+
+    public String getName() {
+        return name == null || name.trim().isEmpty() ? "default" : name;
```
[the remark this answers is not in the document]
bjalon: Done

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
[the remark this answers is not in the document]
bjalon: @kevinleturc ? Are you agree with my comment ?

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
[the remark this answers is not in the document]
bjalon: @kevinleturc  ? Are you agree with my comment ?

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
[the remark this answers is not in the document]
bjalon: Fixed

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/adapters/doc/TemplateBinding.java
bjalon: I embedded this fix in a commit of the history. But I dit a [specific commit for other classes that needed an reordered imports](https://github.com/clockworx-com/nuxeo/commit/957bbefb671863d5c2052f9542526a73d18616c9).

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerContribution.java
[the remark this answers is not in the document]
bjalon: Can I instead just extends the thrown exception to NuxeoException instead. I prefer to keep more expressive exception. What do you think ? I did the in this way... Tell me if you really want a generic NuxeoException instead...

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerServiceImpl.java
bjalon: If you are agree I prefer to keep my specific exception and extends it as NuxeoException... What do you think ? I did the modification in this way...

bjalon:
@efge : I fixed all your comments.
@kevinleturc : I fixed all your comments except 
* 2 comments around exception : because I'm not really agree with that I would like to have a conversation before to fix in your way.
* 1 comment on the usage of a LinkedHashMap. I explained why.
You can contact me with Skype (benjamin.jalon) or Hangout (bjalon@qastia.com) or any other ways :)

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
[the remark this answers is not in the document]
kevinleturc: I'm 👍

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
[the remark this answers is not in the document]
kevinleturc: I'm ok for exceptions with such internal usages. As long as the exception is not thrown outside of the class.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerContribution.java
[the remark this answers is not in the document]
kevinleturc: For this case, I would prefer to have a generic `NuxeoException` as the only benefit there is to have a un-checked exception, so having `NuxeoException` with the original exception as cause should be fine.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerServiceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+    private static final String DEFAULT_SERIALIZER_NAME = "default";
+
+    @Override
+    public Serializer getSerializer(String id) {
+        SerializerContribution contrib = getDescriptor(EXTENSION_POINT_NAME, id);
+
+        if (contrib == null) {
+            contrib = getDescriptor(EXTENSION_POINT_NAME, DEFAULT_SERIALIZER_NAME);
+            if (contrib == null) {
+                throw new UnknownSerializer(id);
```
kevinleturc: Here just a `NuxeoException` with the message who wrote in `UnknowSerializer` is enough.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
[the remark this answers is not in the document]
bjalon: Ok I keep the Exception like that here.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerContribution.java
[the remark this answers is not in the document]
bjalon: Ok

bjalon:
Should be ok, now.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/XMLSerializer.java
```diff
@@ -20,143 +20,33 @@
...
-
-    public static final Namespace ns = new Namespace(XML_NAMESPACE_PREFIX, XML_NAMESPACE);
-
-    public static final QName fieldsTag = DocumentFactory.getInstance().createQName("templateParams", ns);
-
-    public static final QName fieldTag = DocumentFactory.getInstance().createQName("field", ns);
-
-    public static final String DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS";
+    public static Serializer getXmlSerializer() {
+            return Framework.getService(SerializerService.class).getSerializer("xml");
```
efge: `Framework.getService(SerializerService.class).getSerializer("xml")` is used many times, so I think it's worth adding a helper method `getXmlSerializer` to the `SerializerService` itself (can be a default method in the interface). Also in `SerializerService` add a constant `String XML = "xml"` with Javadoc `The well-known name of the XML serializer`.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/adapters/doc/TemplateBasedDocumentAdapterImpl.java
```diff
@@ -68,6 +69,15 @@
 
     protected final TemplateBindings bindings;
 
+    protected Serializer serializer;
+
+    public Serializer getSerializer() {
```
efge: Call it `getXmlSerializer`, and remove the `protected Serializer serializer` field caching, this is a premature optimization. Same in other classes with this pattern.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/processors/AbstractBindingResolver.java
```diff
@@ -70,106 +78,179 @@ protected DocumentObjectWrapper getWrapper() {
...
-                    Serializable value = null;
-                    if (property != null) {
-                        value = property.getValue();
+                }
+                return resultList;
+            case Content:
+                ContentInputType contentInput = ContentInputType.getByValue(param.getSource());
+                if (BlobContent.equals(contentInput))
+                    return extractBlobContent(doc, param);
+                else if (HtmlPreview.equals(contentInput))
```
efge: Please add braces everywhere in `if` and `else` statements (also in several other places in this class).

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/processors/AbstractBindingResolver.java
```diff
@@ -70,106 +78,179 @@ protected DocumentObjectWrapper getWrapper() {
...
+                    return extractHTMLPreview(doc, param);
+                else {
+                    Serializable docPropertyValue = getDocPropertyValue(doc, propKey);
+                    if (docPropertyValue instanceof String)
+                        return handleHtmlField(param.getName(), (String) getDocPropertyValue(doc, propKey));
+                }
```
efge: You're probably missing a `break` here.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/processors/AbstractBindingResolver.java
```diff
@@ -70,106 +78,179 @@ protected DocumentObjectWrapper getWrapper() {
...
+                try {
+                    Serializable docPropertyValue = getDocPropertyValue(doc, propKey);
+                    if (isBlob(docPropertyValue)) {
+                        Blob blob = (Blob) getDocPropertyValue(doc, propKey);
+                        addDefaultMimetypeIfRequired(blob);
+                        return handlePictureField(param.getName(), blob);
                     }
+                } catch (ValueNotFound e) {
+                    return handlePictureField(param.getName(), null);
+                }
```
efge: Missing `break` too.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/processors/AbstractBindingResolver.java
```diff
@@ -70,106 +78,179 @@ protected DocumentObjectWrapper getWrapper() {
...
+                        return nuxeoWrapper.wrap(property);
+                    } catch (TemplateModelException e) {
+                        throw new ValueNotFound(e);
                     }
+                }
+        }
+    }
+
+    protected void addDefaultMimetypeIfRequired(Blob blob) {
+        if (blob.getMimeType() == null || "".equals(blob.getMimeType().trim())) {
```
efge: I know this is refactored code, but please take the opportunity to use `StringUtils.isBlank` here.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/processors/AbstractBindingResolver.java
```diff
@@ -70,106 +78,179 @@ protected DocumentObjectWrapper getWrapper() {
...
+    }
+
+    protected Object extractAutoLoop(TemplateInput param, Property property) {
+        // should do the same on all children properties ?
+        return handleLoop(param.getName(), property);
+    }
+
+    protected Object extractBlobContent(DocumentModel doc, TemplateInput param) throws ValueNotFound {
+        Object propValue = getDocPropertyValue(doc, param.getSource());
+        if (propValue != null && propValue instanceof Blob) {
```
efge: `propValue != null` is redundant because `propValue instanceof Blob` already checks for it.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/processors/AbstractBindingResolver.java
```diff
@@ -195,4 +276,17 @@ protected String getHtmlValue(BlobHolder bh) throws IOException {
         return "";
     }
 
+    protected class ValueNotFound extends Throwable {
```
efge: Should be `protected static class`. Please extend `NuxeoException` instead of `Throwable`. Same for `NoValueToAddInContext`.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/Serializer.java
```diff
@@ -0,0 +1,42 @@
...
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.template.api.TemplateInput;
+
+/**
+ * (was XMLSerializer)
+ * {@link TemplateInput} parameters are stored in the {@link DocumentModel} as a single String Property via XML
+ * Serialization. This class contains the Serialization/Deserialization logic.
+ *
+ * @author Tiry (tdelprat@nuxeo.com)
+ * @author bjalon (bjalon@qastia.com)
```
efge: Add `@since 11.1`

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/Serializer.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public interface Serializer {
+
+    List<TemplateInput> doDeserialization(String content) throws DocumentException;
+
+    String doSerialization(List<TemplateInput> content);
```
efge: Add Javadoc to the methods.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,219 @@
...
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.template.api.InputType;
+import org.nuxeo.template.api.TemplateInput;
+
+/**
+ * {@link TemplateInput} parameters are stored in the {@link DocumentModel} as a single String Property via XML
+ * Serialization. This class contains the Serialization/Deserialization logic.
+ *
+ * @author Tiry (tdelprat@nuxeo.com)
+ * @author bjalon (bjalon@qastia.com)
```
efge: `@since 11.1`

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,219 @@
...
+public class XMLSerializer implements Serializer {
+
+    public static final String XML_NAMESPACE = "http://www.nuxeo.org/DocumentTemplate";
+    public static final String XML_NAMESPACE_PREFIX = "nxdt";
+    public static final Namespace ns = new Namespace(XML_NAMESPACE_PREFIX, XML_NAMESPACE);
+    public static final QName fieldsTag = DocumentFactory.getInstance().createQName("templateParams", ns);
+    public static final QName fieldTag = DocumentFactory.getInstance().createQName("field", ns);
+    public static final String DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS";
+
+    protected static final Log log = LogFactory.getLog(XMLSerializer.class);
```
efge: The `log` is always `private`, we don't want any inheritance.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,219 @@
...
+                try {
+                    SimpleDateFormat dateFormat = new SimpleDateFormat(DATE_FORMAT);
+                    paramValue = dateFormat.parse((String) paramValue);
+                } catch (ParseException e) {
+                    throw new DocumentException(e);
+                }
+                break;
+            case MapValue:
+            case ListValue:
+                LinkedHashMap<String, TemplateInput> listValue = new LinkedHashMap<>();
```
efge: `Map<String, TemplateInput> listValue = new LinkedHashMap<>()`

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,219 @@
...
+            default:
+                paramValue = node.attributeValue("source");
+        }
+        return TemplateInput.factory(paramName, paramType, paramValue, paramDesc, isReadonly, isAutoloop);
+    }
+
+
+    protected Boolean getIsReadonlyFromXMLNode(Element elem) {
+        return elem.attribute("readonly") != null ?
+                Boolean.parseBoolean(elem.attribute("readonly").getValue()) :
```
efge: Put `elem.attribute("readonly")` in a local variable. Same below.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,219 @@
...
+        InputType type = null;
+        if (elem.attribute("type") != null) {
+            type = InputType.getByValue(elem.attribute("type").getValue());
+        }
+
+        if (type == null) {
+            return StringValue;
+        } else {
+            return type;
+        }
```
efge: `return type == null ? StringValue : type`

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,219 @@
...
+                doSerialization(field, param);
+            } catch (TemplateInputBadFormat e) {
+                log.error("Can't Serialize the following param: " + param);
+                root.remove(field);
+            }
+        }
+        return root.asXML();
+    }
+
+    private void doSerialization(Element field, TemplateInput param) throws TemplateInputBadFormat {
```
efge: No `private` methods please, they hamper subclassing.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
```diff
@@ -0,0 +1,219 @@
...
+            if (param.getStringValue() == null) {
+                log.warn(param.getName() + " is null");
+                throw new TemplateInputBadFormat();
+            } else
+                type = StringValue;
+        }
+
+        field.addAttribute("type", type.getValue());
+        if (param.isReadOnly())
+            field.addAttribute("readonly", "true");
```
efge: Always use braces for `if` and `else`.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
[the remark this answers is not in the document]
efge: Extend `NuxeoException` instead of `Throwable`. Class should be `static` and at least `protected`, for subclassing.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerContribution.java
```diff
@@ -0,0 +1,59 @@
...
+    public Serializer getImplementation() {
+        Serializer obj;
+        try {
+            obj = implementationClass.getDeclaredConstructor().newInstance();
+        } catch (NoSuchMethodException e) {
+            throw new NuxeoException("Serializer Contribution NoDefaultConstructorDefined: " + name, e);
+        } catch (IllegalAccessException e) {
+            throw new NuxeoException("Serializer Contribution DefaultConstructorHidden: " + name, e);
+        } catch (InstantiationException | InvocationTargetException e) {
+            throw new NuxeoException("Serializer Contribution ErrorOnInstanciation: " + name, e);
```
efge: Catch just `ReflectiveOperationException` instead of a bunch of different catches.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/SerializerServiceImpl.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ */
+
+package org.nuxeo.template.serializer.service;
+
+import org.nuxeo.ecm.core.api.NuxeoException;
+import org.nuxeo.runtime.model.DefaultComponent;
+import org.nuxeo.template.serializer.executors.Serializer;
+
+public class SerializerServiceImpl extends DefaultComponent implements SerializerService {
```
efge: Add a Javadoc, including `@since 11.1`

addons/nuxeo-template-rendering/nuxeo-template-rendering-xdocreport/src/test/java/org/nuxeo/ecm/platform/template/tests/TestFreemarkerVariableExractor.java
```diff
@@ -24,16 +24,29 @@
...
 import org.nuxeo.template.processors.xdocreport.XDocReportProcessor;
+import org.nuxeo.template.serializer.executors.Serializer;
+import org.nuxeo.template.serializer.service.SerializerService;
 
 public class TestFreemarkerVariableExractor extends NXRuntimeTestCase {
 
+
+    protected Serializer serializer;
+
+    public Serializer getService() {
```
efge: -> `getXmlSerializer`, and remove the field caching.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/XMLSerializer.java
bjalon: I created a specific commit for that (the last one) that will simplify your review.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/processors/AbstractBindingResolver.java
bjalon: @efge : We don't really need that as we always return something in it => if / if else / else
The last else return something... Or I can break and throw exception in the bottom of the method (that should never happens)... Tell me

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/processors/AbstractBindingResolver.java
bjalon: This is normal : if the value is not the blob we go through the default behaviour. That's not my choice, it was like that before me.

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/processors/AbstractBindingResolver.java
bjalon: Ok I put the static.

As I told to Kevin, this is internal exception never thrown outside. Do you think this is required

addons/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/executors/XMLSerializer.java
bjalon: Ok. It seems in many place protected is used... I fixed just in my place... :)

bjalon:
@efge : Hi Florent,
I fixed all the points except : 
* [the exception point](https://github.com/nuxeo/nuxeo/pull/3111#discussion_r301009069) : see my comment. I can put it as NuxeoException, if you want.
* the both breaks ([here](https://github.com/nuxeo/nuxeo/pull/3111#discussion_r300992520) and [there](https://github.com/nuxeo/nuxeo/pull/3111#discussion_r300991342). Tell me what do you think.

I also created a [specific commit for the helper that will simplify the review on that part](https://github.com/clockworx-com/nuxeo/commit/76fa351333925695f5517d095b5e167882cddb1a).

Waiting for your feedback.

troger:
Work on this PR continues in https://github.com/nuxeo/nuxeo/pull/3479.

PR #3161, opened 2019-06-20
NXP-27612: Add platform team as codeowners

kevinleturc:


jcarsique:
If you want to merge into master, then it's worth to create a dedicated nxp ticket that may be backported. Code owners are a GitHub feature, not only tied with pull Panda.

PR #3182, opened 2019-06-28
NXP-26305: add NXQL NOW() with optional added period

efge:
https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/703/

### DIFF
--- a/nuxeo-common/src/main/java/org/nuxeo/common/utils/PeriodAndDuration.java
+++ b/nuxeo-common/src/main/java/org/nuxeo/common/utils/PeriodAndDuration.java
@@ -1,0 +1,217 @@
+/*
+ * (C) Copyright 2019 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *       Florent Guillaume
+ */
+package org.nuxeo.common.utils;
+
+import java.time.Duration;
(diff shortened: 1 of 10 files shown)


nuxeo-common/src/main/java/org/nuxeo/common/utils/PeriodAndDuration.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /**
+     * Adds this {@link PeriodAndDuration} to the given date/time and returns the result.
+     *
+     * @param dateTime the date/time to add
+     * @return the date/time to which the period and duration have been added
+     */
+    public ZonedDateTime plus(ZonedDateTime dateTime) {
+        return dateTime.plus(period).plus(duration);
```
kevinleturc: Could be interesting to make this class implement `TemporalAmount`, as it we could use `ZonedDateTime#plus(PeriodAndDuration)`, but it requires more method to implement (4 methods).
Otherwise, it seems that `addTo` is preferred on a temporal amount. Wdyt?

efge: I initially thought `TemporalAmount` would be too complex to implement but indeed on second look it doesn't seem to be the case, so I'll take a stab. Noted for `addTo` as well in case I don't do it.

efge: Done, please re-review

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestSQLRepositoryQuery.java
```diff
@@ -2533,6 +2533,26 @@ public void testLTTimeWithMilliseconds() {
         assertEquals(1, docs.size());
     }
 
+    @Test
+    public void testQueryNow() throws Exception {
+        setupDocTest();
+        // sleep a second to be sure that the document is found
```
bdelbosc: You can use  waitForFulltextIndexing(); instead of nextTx + sleep

efge: But `waitForFulltextIndexing` has different semantics. Here the sleep is to wait for time to advance.

PR #3193, opened 2019-07-04
NXP-27691: Don't log garbage collection info to the console

ataillefer:


NourNuxeo:
Wouldn't it be relevant to do it as well for other java versions ? 
This improvement doesn't seem specific to java 11 to me ?

ataillefer:
You mean update the Java 8 part?
```
#JAVA_OPTS=$JAVA_OPTS -Xloggc:${nuxeo.log.dir}/gc.log -verbose:gc -XX:+PrintGCDetails -XX:+PrintGCTimeStamps
```
https://github.com/nuxeo/nuxeo/blob/master/nuxeo-distribution/nuxeo-server-tomcat/src/main/resources/tomcat/bin/nuxeo.conf#L27

Launching a 10.10 server with Java 8 and this line uncommented doesn't print the GC logs to the console. The `-verbose:gc` option behavior has changed from JDK 8 to 11.

PR #3195, opened 2019-07-04
Fix NXP-27575 pipe command on jx

kevinleturc:


nuxeo-core/nuxeo-platform-commandline-executor/src/main/java/org/nuxeo/ecm/platform/commandline/executor/service/executors/ShellExecutor.java
```diff
@@ -170,7 +170,9 @@ protected int getReturnCode(List<Process> processes) {
      * The streams are both closed when the copy is finished.
      *
      * @since 7.10
+     * @deprecated since 11.1, seems unused
      */
+    @Deprecated(since = "11.1")
```
ataillefer: Should be in the cleanup commit no, so I guess in https://github.com/nuxeo/nuxeo/pull/3198?

kevinleturc: In fact I can't put it in the cleanup commit, this method became unused after my improvement.

kevinleturc: But I didn't have it in #3198, error in cherry-pick, will fix it.

ataillefer: OK, so it should be in "NXP-27575: Use new pipeline API for piped process" no?

kevinleturc: Exactly, this is the case for #3198 now, will do some rebase after its merge to just have the two commits for JX in this PR ;)

PR #3210, opened 2019-07-10
NXP-27649 blob computation temp dir delete on exit

andreinechaev:


nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/action/computation/AbstractTransientBlobComputation.java
```diff
@@ -47,6 +47,7 @@ public void init(ComputationContext context) {
         super.init(context);
         try {
             temp = Files.createTempDirectory(metadata().name());
+            temp.toFile().deleteOnExit();
```
kevinleturc: It is probably better to use `Framework#trackFile`.

andreinechaev: thought about it but then what to use as ref? `context`? don't think the object will be alive needed time. 
The main complain was that temp folder wasn't cleaned up at server startup

kevinleturc: We use the file itself as marker.

andreinechaev: ok, thx

ataillefer:
Would it be possible to add a unit test?

andreinechaev:
> Would it be possible to add a unit test?

@ataillefer I don't see how we can test shutdown of jvm and cleanup of FS. open for suggestions.

ataillefer:
I would've imagined that at some point the blob is not supposed to be referenced anymore (before shutdown), thus garbage collected.
But I don't know when this happens and if'ts possible to simulate.
Maybe it's not doable!

kevinleturc:
It's probably possible to shutdown the computation runner which will free the computation and thus the Path. But this test will probably be random.

The marker I suggested may not be relevant, as `toFile` will return a new File, it will probably be garbage collected before the Path itself. Maybe interesting to use the Path as marker.

@andreinechaev can you squash your commits and change the marker? Next, I think a T&P could highlights issues if marker is not the right one.

kevinleturc:
T&P is ok: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-anechaev/12/

PR #3220, opened 2019-07-13
NXP-16751: Allows for readonly metadata mapping (do not write from doc to blob)

ThibArg:


addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/BinaryMetadataServiceImpl.java
```diff
@@ -252,16 +253,18 @@ public void handleUpdate(List<MetadataMappingDescriptor> mappingDescriptors, Doc
             if (blob != null) {
                 boolean isDirtyMapping = isDirtyMapping(mappingDescriptor, doc);
                 if (isDirtyMapping) {
-                    BlobManager blobManager = Framework.getService(BlobManager.class);
-                    BlobProvider blobProvider = blobManager.getBlobProvider(blob);
-                    // do not write metadata in blobs backed by extended blob providers (ex: Google Drive) or blobs from
-                    // providers that prevent user updates
-                    if (blobProvider != null && (!blobProvider.supportsUserUpdate() || blobProvider.getBinaryManager() == null)) {
-                        return;
+                    if(!mappingDescriptor.getReadOnly()) {
```
kevinleturc: Does it make sense to join the `isDirtyMapping` and `readOnly` conditions? As it we won't loose history and code will be less complex.

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
```diff
@@ -89,4 +93,8 @@ public String getBlobXPath() {
     public Boolean getIgnorePrefix() {
         return ignorePrefix;
     }
+
+    public Boolean getReadOnly() {
```
kevinleturc: What do you think about keeping the `Boolean` type for the field (which will allow merge for future usage) but rename this method `isReadOnly` and make it return `boolean`. For instance:
```
public boolean isReadOnly() {
    return Boolean.TRUE.equals(readOnly);
}
```

addons/nuxeo-binary-metadata/src/test/resources/binary-metadata-readonly-test.xml
```diff
@@ -0,0 +1,37 @@
...
+
+  <extension target="org.nuxeo.ecm.platform.actions.ActionService" point="filters">
+    <filter id="hasFileType">
+      <rule grant="true">
+        <type>File</type>
+      </rule>
+    </filter>
+  </extension>
+
+</component>
```
kevinleturc: Could you add a new line a the end?

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
```diff
@@ -45,6 +46,9 @@
 
     @XNode("@ignorePrefix")
     protected Boolean ignorePrefix = true;
+    
+    @XNode("@readOnly")
```
kevinleturc: Can you add the since below please?
```
/** @since 11.1 */
```

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/BinaryMetadataServiceImpl.java
ThibArg: I wanted to separate the `if` handling the mapping from the `else` that is about the blob.
But if it's better practice to merge them, I'm ok with that of course

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
ThibArg: I actually wanted to do that?. But then I noticed there already was this `getIgnorePrefix`, so I decided to keep it as `get` too, well, stay in "harmony". (I thought about renaming `getIgnorePrefix` to `ìgnorePrefix`, but thought I should not modify something that is not related to the ticket)

So, I'm ok with "is", want to make sure you're ok with having different styles in the class :-)

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/BinaryMetadataServiceImpl.java
kevinleturc: It depends on the case. Globally if we can join the ifs we do it.
Here, under the scenario below during an update:
- update some metadata which are under the binary mapping
- the mapping is read only
- update the blob which is under the binary mapping

Without the join, we do nothing whereas with it, we update the document with metadata coming from blob. So updated metadata could be erased. I think this is not what we want? But, on the other hand, if we just update the document's blob, these metadata will be updated (if we join the two conditions).

In the end, question is more: Do we want to keep binary metadata in sync with their blob when readOnly flag is true?

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
kevinleturc: Hey hey, you're totally right about  consistency. Here, what we would have done is to introduce `isIgnorePrefix` + deprecate `getIgnorePrefix` + replace everywhere former method name by the new one. I though it was too much to ask you for this contribution.

If you want to do it, it would be very cool ✌️. Otherwise, put the new name and we'll clean it one day 😉.

ThibArg: ok, good. No no, it's not too much, i'll do it, thanks :-)

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/BinaryMetadataServiceImpl.java
ThibArg: So, no actually. When readonly is true we want to:
-> Update the fields only if the blob changes
-> Do nothing to the blog if the fields change.

So ultimately, blob metadata and fields can end up being desynchronized (user bodies a field). But if the blob is modified, then all fields are reset.

I mean, we don't keep track of changes. We don't do this: "If field1 was modified by a user, we don't modify the field if a new blob is uploaded". This could be another feature, later :-)

kevinleturc: Sure. 👍

ThibArg: Also, while re-reviewing it, I did this also to better stick to the comment of the interface (`BinaryMetadataService#handleUpdate`}, so we clearly separate a change in the metadata from a change in the blob (and I forgot to update the comment of these methods to briefly talk about the readOnly flag.

So, ok to keep it with 2 `if`?

kevinleturc: It's ok 😉

kevinleturc:
T&P: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-kleturc-3/31/

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
```diff
@@ -45,6 +46,10 @@
 
     @XNode("@ignorePrefix")
     protected Boolean ignorePrefix = true;
+    
+    /** @since 11.1 */
+    @XNode("@readOnly")
+    protected Boolean readOnly = false;
```
efge: Using default values in the fields makes it impossible to have two contributions merging each other without redefining a field (I know there's no merge right now but it's to prepare for the future). You should just have `protected Boolean readOnly` (so default to `null`).

Same for `ignorePrefix` except that because you want the default to be true, make the getter be `return !Boolean.FALSE.equals(ignorePrefix)`. You can check that the end result is the same.

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/BinaryMetadataServiceImpl.java
```diff
@@ -252,16 +253,18 @@ public void handleUpdate(List<MetadataMappingDescriptor> mappingDescriptors, Doc
             if (blob != null) {
                 boolean isDirtyMapping = isDirtyMapping(mappingDescriptor, doc);
                 if (isDirtyMapping) {
-                    BlobManager blobManager = Framework.getService(BlobManager.class);
-                    BlobProvider blobProvider = blobManager.getBlobProvider(blob);
-                    // do not write metadata in blobs backed by extended blob providers (ex: Google Drive) or blobs from
-                    // providers that prevent user updates
-                    if (blobProvider != null && (!blobProvider.supportsUserUpdate() || blobProvider.getBinaryManager() == null)) {
-                        return;
+                    if(!mappingDescriptor.isReadOnly()) {
```
efge: Space before `if` and `(`

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/BinaryMetadataServiceImpl.java
```diff
@@ -252,16 +253,18 @@ public void handleUpdate(List<MetadataMappingDescriptor> mappingDescriptors, Doc
...
+                	    BlobManager blobManager = Framework.getService(BlobManager.class);
```
efge: Indent is wrong because of a tab here. Configure your editor to never use tabes.

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
```diff
@@ -45,6 +46,10 @@
 
     @XNode("@ignorePrefix")
     protected Boolean ignorePrefix = true;
```
efge: This must be just `protected Boolean ignorePrefix;`

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
efge: This must be just `protected Boolean readOnly;`

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
```diff
@@ -86,7 +91,19 @@ public String getBlobXPath() {
         return blobXPath;
     }
 
+    /**
+     * @deprecated since 11.1, use {@link #ignorePrefix()}
+     */
+    @Deprecated(since = "11.1")
     public Boolean getIgnorePrefix() {
-        return ignorePrefix;
+        return !Boolean.FALSE.equals(ignorePrefix);
```
efge: `getIgnorePrefix`, the deprecated one, should just `return ignorePrefix()` by delegating to the new one.

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
```diff
@@ -86,7 +91,19 @@ public String getBlobXPath() {
...
+    }
+
+    public boolean ignorePrefix() {
+    	return Boolean.TRUE.equals(ignorePrefix);
```
efge: The new `ignorePrefix()` should do `!Boolean.FALSE.equals(ignorePrefix)` in order to return `true` if no value has been defined, therefore "defaulting" to `true`.

ThibArg: 👍

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
ThibArg: 👍

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
ThibArg: 👍

addons/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/MetadataMappingDescriptor.java
ThibArg: 👍

efge:
In the future please remember to name your branches `fix-NXP-16751-...` or `feature-NXP-16751-...`, this is our convention and this is also recognized by various tools, including one that would have included a link to the JIRA ticket automatically at the start of this PR.

ThibArg:
I all very sorry, I did not merged to master 2 months ago (provided a quick patched 10.10 jar for a demo). So, as Florent explained to me, I did re-ran a T&P that was successful => I am "rebase-merging"

PR #3289, opened 2019-08-20
Fix nxp 27683 comments permissions not set correctly

RSalem07:
PR In Progress : https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-saouana/97/

### DIFF
--- a/nuxeo-features/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/main/java/org/nuxeo/elasticsearch/ElasticSearchFilter.java
+++ b/nuxeo-features/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/main/java/org/nuxeo/elasticsearch/ElasticSearchFilter.java
@@ -36,7 +36,9 @@ import org.nuxeo.elasticsearch.listener.ElasticSearchInlineListener;
  */
 public class ElasticSearchFilter implements Filter {
 
-    public static final String ES_SYNC_FLAG = "nx_es_sync";
+    public static final String ES_SYNC_FLAG = "nx-es-sync";
+
+    protected static final String ES_SYNC_FLAG_COMPAT = "nx_es_sync";
 
     @Override
     public void init(FilterConfig filterConfig) throws ServletException {
@@ -50,7 +52,9 @@ public class ElasticSearchFilter implements Filter {
             return;
         }
         final HttpServletRequest httpRequest = (HttpServletRequest) request;
-        final boolean esSync = Boolean.parseBoolean(httpRequest.getHeader(ES_SYNC_FLAG));
+        final boolean esSync = Boolean.parseBoolean(httpRequest.getHeader(ES_SYNC_FLAG))
+                // NXP-28075: keep compatibility with old header with underscores
+                || Boolean.parseBoolean(httpRequest.getHeader(ES_SYNC_FLAG_COMPAT));
         if (!esSync) {
             chain.doFilter(request, response);
             return;


nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/PropertyCommentManager.java
```diff
@@ -186,11 +180,14 @@ public Comment createComment(CoreSession session, Comment comment)
...
-        if (!CoreInstance.doPrivileged(session, s -> {return s.exists(docRef);})) {
+        if (!CoreInstance.doPrivileged(session, s -> {
+            return s.exists(docRef);
+        })) {
             throw new CommentNotFoundException("The document or comment " + comment.getParentId() + " does not exist.");
         }
         DocumentRef ancestorRef = CoreInstance.doPrivileged(session, s -> {
-            return getAncestorRef(s, s.getDocument(new IdRef(parentId)));
+            DocumentModel document = s.getDocument(new IdRef(parentId));
+            return getAncestorRef(s, document);
```
kevinleturc: Changes not needed.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/PropertyCommentManager.java
```diff
@@ -203,8 +200,9 @@ public Comment createComment(CoreSession session, Comment comment)
         }
 
         return CoreInstance.doPrivileged(session, s -> {
-            String path = getCommentContainerPath(s, parentId);
-            DocumentModel commentModel = s.createDocumentModel(path, COMMENT_NAME, COMMENT_DOC_TYPE);
+            DocumentModel documentToComment = s.getDocument(ancestorRef);
```
kevinleturc: `documentToComment` can lead to miscomprehension, for instance: for a reply what is the document to comment? The document whose we try to reply to or the top level document?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/impl/TestPropertyCommentManager.java
```diff
@@ -561,8 +563,18 @@ public void testCreateCommentAsRegularUser() {
             subComment.setText(text);
             subComment.setParentId(createdComment.getId());
 
-            Comment createdSubcomment = commentManager.createComment(johnSession, subComment);
-            assertEquals(createdComment.getId(), createdSubcomment.getParentId());
+            Comment createdSubComment = commentManager.createComment(johnSession, subComment);
+            assertEquals(createdComment.getId(), createdSubComment.getParentId());
+
+            DocumentModelList documents = johnSession.getDocuments(
```
kevinleturc: Not needed as you're retrieving them unitary just after.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/impl/TestPropertyCommentManager.java
```diff
@@ -625,13 +645,66 @@ public void testGetCommentAsRegularUser() {
...
+
+        try (CloseableCoreSession janeSession = CoreInstance.openCoreSession(doc.getRepositoryName(), "jane")) {
+            DocumentModelList documents = janeSession.getDocuments(
+                    new DocumentRef[] { new IdRef(comment.getId()), new IdRef(subComment.getId()) });
+            assertEquals(0, documents.size());
+        }
+    }
+
+    @Test
+    public void testQueryCommentAsRegularUser() {
```
kevinleturc: This test and this [one](https://github.com/nuxeo/nuxeo/pull/3262/files#diff-ce5dd04ad48c6df00884ab89cc9ae552R68) should be pretty the same and should be the abstract test class for the 2nd PR to merge.

RSalem07: yes, we were thinking to rework this test class because there is a lot duplicated code (create files, create the same comments...) but initially we  were thinking to add this work on this same PR with perhaps a separate commit

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/migrator/AbstractRepositoryMigrator.java
```diff
@@ -84,8 +95,17 @@ protected void migrateRepository(String repositoryName) {
 
     protected abstract String probeSession(CoreSession session);
 
+    /**
+     * @deprecated since 11.1  Use {@link #migrateSession(MigrationContext, String, CoreSession)} instead
+     */
+    @Deprecated(since = "11.1", forRemoval = true)
     protected abstract void migrateSession(CoreSession session);
```
kevinleturc: We should provide a default implementation here. This is weird to have to implement something deprecated.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/migrator/AbstractRepositoryMigrator.java
```diff
@@ -58,8 +58,19 @@ protected void reportProgress(String repositoryName, String message, long num, l
         reportProgress(String.format("[%s] %s", repositoryName, message), num, total);
     }
 
+    /**
+     * @deprecated since 11.1. Use {@link #migrateRepository(MigrationContext, String, String)} instead
+     */
+    @Deprecated(since = "11.1", forRemoval = true)
```
kevinleturc: There's also the field and its usage to deprecate. You will need to keep affectation in order to be backward compatible, as it you can use these deprecated field(s) in the deprecated method to provide backward compatibility.

nuxeo-features/nuxeo-automation/nuxeo-automation-features/src/test/java/org/nuxeo/ecm/automation/core/test/collections/CreateCollectionTest.java
```diff
@@ -84,29 +84,4 @@ public void testCreateCollectionWithPath() throws Exception {
         String collectionPath = testWorkspace.getPathAsString() + "/" + COLLECTION_NAME;
         assertTrue(session.exists(new PathRef(collectionPath)));
     }
-
-    @Test(expected=OperationException.class)
```
kevinleturc: `OperationException` is now an unused import.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,21 +73,67 @@
 
     protected static final int BATCH_SIZE = 50;
 
+    /**
+     * /**
```
kevinleturc: Something wrong there.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,21 +73,67 @@
...
+    protected boolean hasUnsecuredComments(CoreSession session) {
+        List<DocumentModel> folders = getCommentFolders(session);
+        return folders.stream().anyMatch((folder) -> session.hasChildren(folder.getRef()));
+    }
+
+    /**
+     * @since 11.1
+     */
+    protected List<DocumentModel> getCommentFolders(CoreSession session) {
+        return session.query(GET_HIDDEN_FOLDERS_QUERY, (doc) -> doc.getName().equals(COMMENTS_DIRECTORY)
```
kevinleturc: Here you're using post filtering API but this is not performant. On a very large repository, there could be several `HIDDEN_FOLDER_TYPE `. We need to find a more performant query to get desired comments.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,21 +73,67 @@
...
+                || session.getDocument(doc.getParentRef()).getName().equals(COMMENTS_DIRECTORY));
     }
 
     @Override
     protected void migrateSession(CoreSession session) {
```
kevinleturc: This API shouldn't exist in migrators, see my previous comments.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -139,6 +249,26 @@ protected void migrateComments(CoreSession session, RelationManager relationMana
...
+        comment.setACP(acp, true);
+    }
+
+    /**
+     * @since 11.1
+     */
+    protected void migrateCommentsFromPropertyToSecured(CoreSession session, PropertyCommentManager commentManager,
+            DocumentModel comment) {
+        comment.putContextData(DISABLE_NOTIFICATION_SERVICE, TRUE); // Remove notifications
+        session.move(comment.getRef(), commentManager.getAncestorRef(session, comment), null);
```
kevinleturc: I'm in favor to introduce `DocumentRef getAncestorRef(CoreSession, String)` in the interface as it you'll avoid to cast. Wdyt?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -128,6 +237,7 @@ protected void migrateComments(CoreSession session, RelationManager relationMana
         if (parent != null && comment != null) {
             comment.putContextData(DISABLE_NOTIFICATION_SERVICE, TRUE); // Remove notifications
             comment.setPropertyValue(COMMENT_PARENT_ID, parent.getId());
+            stripACLs(comment);
```
kevinleturc: I would say we don't touch previous migration.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/migrator/AbstractRepositoryMigrator.java
```diff
@@ -58,8 +58,19 @@ protected void reportProgress(String repositoryName, String message, long num, l
...
     protected void migrateRepository(String repositoryName) {
-        TransactionHelper.runInTransaction(() -> CoreInstance.doPrivileged(repositoryName, this::migrateSession));
+        TransactionHelper.runInTransaction(() -> CoreInstance.doPrivileged(repositoryName, (CoreSession session) -> migrateSession(session)));
+    }
+
+    /**
+     * @since 11.1
+     */
+    protected void migrateRepository(MigrationContext context, String step,  String repositoryName) {
```
kevinleturc: It may be better to keep order of `Migrator#run` method, wdyt?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -82,41 +142,90 @@ protected void migrateSession(CoreSession session) {
...
+        } else if (MIGRATION_STEP_PROPERTY_TO_SECURED.equals(step)) {
+            migrateSessionPropertyToSecured(session);
         }
     }
 
     @Override
     public String probeState() {
         List<String> repositoryNames = Framework.getService(RepositoryService.class).getRepositoryNames();
-        if (repositoryNames.stream().map(this::probeRepository).anyMatch(isEqual(MIGRATION_STATE_RELATION))) {
+        List<String> probes = repositoryNames.stream().map(this::probeRepository).collect(Collectors.toList());
```
kevinleturc: You can use a `Set` as you're only using `contains`.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -82,41 +142,90 @@ protected void migrateSession(CoreSession session) {
...
+            return MIGRATION_STATE_PROPERTY;
         }
-        return MIGRATION_STATE_PROPERTY;
+        return MIGRATION_STATE_SECURED;
     }
 
     @Override
     public void run(String step, MigrationContext migrationContext) {
-        if (!MIGRATION_STEP_RELATION_TO_PROPERTY.equals(step)) {
+        if (!List.of(MIGRATION_STEP_RELATION_TO_PROPERTY, MIGRATION_STEP_PROPERTY_TO_SECURED).contains(step)) {
```
kevinleturc: `Set` too.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/TestCommentsMigrator.java
```diff
@@ -241,33 +206,29 @@ public void testProbe() {
...
-        // just a relation-based comment, detected as not migrated
+        // Just a relation-based comment, detected as not migrated
         DocumentModel comment = session.createDocumentModel(null, "comment", COMMENT_DOC_TYPE);
         comment.setPropertyValue(COMMENT_AUTHOR, session.getPrincipal().getName());
         comment = relationCommentManager.createComment(file, comment);
         session.save();
         assertEquals(MIGRATION_STATE_RELATION, migrator.probeState());
 
-        // both a relation-based comment and a property-based comment, detected as not migrated
-        DocumentModel otherComment = session.createDocumentModel(null, "comment", COMMENT_DOC_TYPE);
```
kevinleturc: We should keep this block.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/TestCommentsMigrator.java
```diff
@@ -241,33 +206,29 @@ public void testProbe() {
 
         CommentServiceConfig config = newConfig();
         CommentManager relationCommentManager = new CommentManagerImpl(config);
-        CommentManager propertyCommentManager = new PropertyCommentManager();
 
         DocumentModel domain = session.createDocumentModel("/", "test-domain", "Domain");
         session.createDocument(domain);
-        DocumentModel file = session.createDocumentModel("/test-domain", "file1", "File");
+        DocumentModel file = session.createDocumentModel("/test-domain", "anotherFile", "File");
```
kevinleturc: This became complex to understand because you're ending with tests leveraging `@Before` setup and some tests don't.

I would rather keep the design of the test which is shared between all migrator tests.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/TestCommentsMigrator.java
```diff
@@ -279,7 +240,135 @@ public void testProbe() {
         List<Statement> statementList = graph.getStatements(commentRes, null, null);
         graph.remove(statementList);
 
+        // No more relation comments detected as 'Relation'
```
kevinleturc: detected as property (or something else), right?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/TestCommentsMigrator.java
```diff
@@ -279,7 +240,135 @@ public void testProbe() {
...
+    protected void checkCommentsForDocument(String docId) {
+        DocumentModelList dml = session.query(String.format("SELECT * FROM Document WHERE %s = '%s' AND %s = '%s' ",
+                NXQL.ECM_PARENTID, docId, NXQL.ECM_PRIMARYTYPE, COMMENT_DOC_TYPE));
+        assertEquals(NB_COMMENTS / 2, dml.size());
     }
 
+    /**
+     * @since 11.1
+     */
+    protected void migrateFromRelationToProperty(Migrator migrator) {
```
kevinleturc: Can you keep an order on methods such as migration steps order for instance.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/TestCommentsMigrator.java
```diff
@@ -279,7 +240,135 @@ public void testProbe() {
...
+        runMigration(() -> migrator.run(MIGRATION_STEP_RELATION_TO_PROPERTY, getMigrationContext(progressLines)));
+
+        CommentManager propertyCommentManager = new PropertyCommentManager();
+        List<Comment> commentsForFile1 = propertyCommentManager.getComments(session, firstFileToComment.getId());
+        List<Comment> commentsForFile2 = propertyCommentManager.getComments(session, secondFileToComment.getId());
+
+        assertEquals(NB_COMMENTS, commentsForFile1.size() + commentsForFile2.size());
+        for (Comment comment : commentsForFile1) {
+            assertEquals(firstFileToComment.getId(), comment.getParentId());
+            assertNull(session.getDocument(new IdRef(comment.getId())).getACP().getACL(LOCAL_ACL));
```
kevinleturc: `CoreSession#getACP(DocumentRef)` could be better here.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/TestCommentsMigrator.java
```diff
@@ -279,7 +240,135 @@ public void testProbe() {
...
+        }
+    }
+
+    /**
+     * @since 11.1
+     */
+    protected void migrateFromPropertyToSecured(Migrator migrator) {
+        List<String> progressLines = new ArrayList<>();
+
+        runMigration(() -> migrator.run(MIGRATION_STEP_PROPERTY_TO_SECURED, getMigrationContext(progressLines)));
```
kevinleturc: `progressLines` usage looks weird to me. Previous usage was less surprising as the anonymous was directly in the test but as it isn't the case anymore I would rather go to a static class, inside this class, holding this list.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/TestCommentsMigrator.java
```diff
@@ -279,7 +240,135 @@ public void testProbe() {
...
+            public boolean isShutdownRequested() {
+                return false;
+            }
+        };
+    }
+
+    /**
+     * @since 11.1
+     */
+    protected void mockUnsecuredPropertyComments() {
```
kevinleturc: This method makes me think that you need to have a working code after applying the HF without having to migrate to secured properties.

RSalem07:
T&P Ok on default configuration (H2 DB) : https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-saouana/99/

T&P on mongodb : https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-nalkotob/44/ some tests failed, because with our new commit (third one https://github.com/nuxeo/nuxeo/pull/3289/commits/01ec8acff2de0c4612447f88040c06ac15aa1bb5) we used projection, which  failed on mongo ` Projection not supported: COUNT(ecm:uuid)`

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/resources/OSGI-INF/comment-notification-contrib.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component name="org.nuxeo.ecm.platform.comment.service.notification">
```
kevinleturc: Where is this contrib used?

efge:
I'm wondering about proxies. Could you add a check that a comment on a proxy used to work, and still works with the new system?

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/migrator/AbstractRepositoryMigrator.java
RSalem07: yep, you right 👍 , we will implement it by calling the other method with `null` parameters

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: yes we agree, it's why we make a new commit which proposal is to use the scrolling, i don't know wdty, plus that the T&P on mongo failed, https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-nalkotob/44/ .

another way of we can't using the scroll solution is to get `Comments` folder in a separate query and then get all its child as below :

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: yes we agree it's better, the only problem that i see, it's as we discuss yesterday, is about using the old service (CommentManagerImpl) in 10.10, if a customer take the choice to continue using this old service, if he attempt to retrieve the ancestor it will failed because this method will try to get `comment:parentId` which is `null`, but we can override it and throw an exception

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/migrator/AbstractRepositoryMigrator.java
RSalem07: yep thx, we forgot  this one

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
kevinleturc: It should work, note that you can use a NXQL to get the children once you have all parent ids.

This will have better performance instead of fetching them one by one, furthermore with queryProjection you can just get what you need which it should be the id in order to move it.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
kevinleturc: If you use that, but if you use relations for old service you can get the top level commented document.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: after discussion with @NourNuxeo this add this modification is needed to avoid any security issue, with @troger we decided to add it, in a case where there no ACLs modification script as done by Thierry Martin under 10.10 and the next release, in the case where the client decide only to migrate from `Relation` to `Property`

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: when you said it should work, you are talking about the solution with the use of `while` loop... ?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
kevinleturc: Still, if the client wants something secured it has to run the next migration. Issue I see there is that we're going to have two different cases for the same migration state which could make the support analysis more difficult for almost nothing as we have a migration to fix it.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: when you said it should work, you are talking about the solution with the use of `while` loop... ? 

About the use of NXQL to get the children of given parent Ids: are you talking about using the `IN` operator in the NXQL ? If it's the case, we tried it at the start of the migration development but we didn't find a way to do the `SubQuery IN (select...)` and there is no way to do a subquery in NXQL as far as we know. On the other hand, we know that we can query in (parent1, parent2...) but we didn't find a pretty way to do it so far unless we use the StringBuilder to build the id list (make concatenations ....)

kevinleturc: The logic may work, the loop is to avoid due to performance.

Indeed there's no sub query in NXQL.

Once you have the parent folder id, you can query its direct children with:
```
SELECT * FROM Document WHERE ecm:parentId IN (...)
```
Or all children with:
```
SELECT * FROM Document WHERE ecm:ancestorId IN (...)
```

To build the list of ancestorIds you can use streams:
```java
parentIds.stream().map(id -> "'" + id + "'").collect(Collectors.joining(", "));
```
which will produces: _'id1', 'id2', ..., 'idn'_

Or:
```java
parentIds.steam().collect(Collectors.joining(", ", "('", "')");
```
which will produces: _('id1', 'id2', ..., 'idn')_

efge: Check that the list of `parentIds` is not empty before doing that though.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/TestCommentsMigrator.java
RSalem07: Before our new devs, the two methods (`testMigration` and `testMigrationThroughService`) create a relation comments in the runMigration method, the `testProbe` in other hand doesn't rely on the run migration but create his own data.  After refactoring, it leverages the ones in the before like the 2 others

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/TestCommentsMigrator.java
RSalem07: please see https://github.com/nuxeo/nuxeo/pull/3289#discussion_r318125639, now there no need to create a new comment, as they are created by the `@before`

kevinleturc: Ok for document creation, but the assertion of `probeState` for a repository containing both relation and property comments was lost (at least we don't test it explicitly).

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: We end up by using a NXQL with the `IN` operator : 
- To get all comments folders `Ids` (Comments under Domain + the Sub folders under these Comments) using session#queryProjection
- To get all comments documents, unfortunately we need the comment  document model to move it from the current path to the new one, 
` session.move(comment.getRef(), commentManager.getAncestorRef(session, comment), null);`  i don't know if we can get the AncestorIds (comment:ancestorIds) in projection our idea is to replace the call on `commentManager.getAncestorRef` and if not what is the best way to get this document model directly by using `session#query` or by getting the document unitary `session#getDocument` (then it will be use the session cache if its loaded before, but we can end up with memory issu if there is many comments)

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -139,6 +251,36 @@ protected void migrateComments(CoreSession session, RelationManager relationMana
...
+        comment.putContextData(DISABLE_NOTIFICATION_SERVICE, TRUE); // Remove notifications
+        session.move(comment.getRef(), commentManager.getAncestorRef(session, comment), null);
+        stripACLs(comment);
+    }
+
+    /**
+     * @since 11.1
+     */
+    protected String buildQuery(List<String> parentIds, String docType) {
+        String inClause = parentIds.stream().map(entry -> "'" + entry + "'").collect(Collectors.joining(", "));
```
efge: No need for `.map`, just do `.stream().collect(Collectors.joining("', '", "('", "')"))`

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -139,6 +251,36 @@ protected void migrateComments(CoreSession session, RelationManager relationMana
...
+
+        return String.format("SELECT %s FROM Document WHERE %s IN(%s) AND %s = '%s'", ECM_UUID, ECM_PARENTID, inClause,
```
efge: `IN` is not a function so we prefer to write `IN (%s)`

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment-api/src/main/java/org/nuxeo/ecm/platform/comment/api/CommentManager.java
```diff
@@ -322,4 +323,11 @@ void deleteExternalComment(CoreSession session, String entityId)
      */
     boolean hasFeature(Feature feature);
 
+    /**
+     * Gets the ancestor document ref for the given document.
+     * 
+     * @since 11.1
+     */
+    DocumentRef getAncestorRef(CoreSession session, DocumentModel documentModel);
```
efge: As you're making this public, it's worth explain what exactly is an ancestor in this context, because in general in Nuxeo an ancestor is the parent or the parent's parent or the parent's parent's parent etc. So there's no single ancestor, we usually talk about the set of ancestors for a document.

Or consider renaming the method to something else if it makes sense.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,21 +74,81 @@
...
+            List<String> timestampCommentFoldersIds = session.queryProjection(query, 0, 0)
+                                                             .stream()
+                                                             .map(entry -> (String) entry.get(ECM_UUID))
+                                                             .collect(Collectors.toList());
+
+            parentIds.addAll(rootCommentsFolderIds);
+            parentIds.addAll(timestampCommentFoldersIds);
+        }
+
+        return parentIds;
```
efge: The blank lines in this method don't bring anything, this is already very readable without them.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: We end up by using a NXQL with the `IN` operator : 
- To get all comments folders `Ids` (Comments under Domain + the Sub folders under these Comments) using session#queryProjection
- To get all comments documents, unfortunately we need the comment  document model to move it from the current path to the new one, 
` session.move(comment.getRef(), commentManager.getAncestorRef(session, comment), null);`  i don't know if we can get the AncestorIds (`comment:ancestorIds`) in projection our idea is to replace the call on `commentManager.getAncestorRef` and if not what is the best way to get this document model directly by using `session#query` or by getting the document unitary `session#getDocument` (then it will be use the session cache if its loaded before, but we can end up with memory issu if there is many comments) 

another question, we want to use a projection `count(ecm:uuid)`to check if there isa unsecured comment, but we think it will not be work on  [`mongoDB`](https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-nalkotob/44/ )

RSalem07:
We added a test to comment  a proxy document, the result is the comment document is related to the proxy one and not the target.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: We endup by creating a projection query that give us the `ecm:uuid` of the comment document and his top level `ancestor` (the file being commented) but unfortunately. We didn't put this code, because in our migration we need the document model for two propos :

RSalem07: We endup by creating a projection query that give us the ecm:uuid of the comment document and his top level ancestor (the file being commented) but unfortunately. We didn't put this code, because in our migration we need the document model for two propos :

RSalem07: We made the query with projection that give (`ecm:uuid` and `ancestor`), but our migration need the document model to : StripACL and disabling notification event.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/AbstractCommentManager.java
```diff
@@ -160,4 +162,38 @@ protected void setCommentPermissions(CoreSession session, DocumentModel document
...
     }
 
+    /**
+     * @since 11.1
+     */
+    protected DocumentModel getThreadForComment(CoreSession session, DocumentModel comment)
+            throws CommentSecurityException {
+
+        NuxeoPrincipal principal = session.getPrincipal();
+        return CoreInstance.doPrivileged(session, s -> {
```
efge: I would switch the names `s` and `session` because we tend to naturally want to use `session` in the lambda.

Same in the next method.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentManagerImpl.java
```diff
@@ -500,4 +497,21 @@ public boolean hasFeature(Feature feature) {
...
         }
     }
+
+    /**
+     * @since 11.1
+     */
+    @Override
+    public DocumentRef getAncestorRef(CoreSession session, DocumentModel documentModel) {
+        return CoreInstance.doPrivileged(session, s -> {
+            if (!documentModel.hasSchema(COMMENT_SCHEMA)) {
```
efge: `s` doesn't seem to be used at all in the lambda, so `doPriviledged` is useless. Or is this a mistake somewhere else?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,161 @@
 
     protected static final int BATCH_SIZE = 50;
 
+    /**
+     * @since 11.1.
+     */
+    public static final String GET_COMMENTS_FOLDERS_QUERY = String.format(
+            "SELECT %s FROM Document WHERE %s='%s' AND %s='%s'", ECM_UUID, ECM_NAME, COMMENTS_DIRECTORY,
```
efge: Cosmetic: usual SQL style is to add spaces around `=`.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,161 @@
...
+    /**
+     * @since 11.1
+     */
+    protected List<DocumentModel> getUnsecuredComments(CoreSession session) {
+        List<String> parentIds = getCommentFolders(session);
+        if (!parentIds.isEmpty()) {
+            return session.query(buildAllCommentsQuery(parentIds, COMMENT_DOC_TYPE));
+        }
+
+        return Collections.emptyList();
```
efge: I tend to prefer early-exit for exceptional or easy cases:
```
        if (parentIds.isEmpty()) {
            return Collections.emptyList();
        }
        return session.query(buildAllCommentsQuery(parentIds, COMMENT_DOC_TYPE));
```

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,161 @@
...
+        List<String> parentIds = new ArrayList<>();
+
+        List<String> rootCommentsFolderIds = session.queryProjection(GET_COMMENTS_FOLDERS_QUERY, 0, 0)
+                                                    .stream()
+                                                    .map(entry -> (String) entry.get(ECM_UUID))
+                                                    .collect(Collectors.toList());
+
+        if (!rootCommentsFolderIds.isEmpty()) {
+            String query = buildAllCommentsQuery(rootCommentsFolderIds, HIDDEN_FOLDER_TYPE);
+            List<String> timestampCommentFoldersIds = session.queryProjection(query, 0, 0)
```
efge: I don't get what "timestamp" means in the name.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,161 @@
...
+     * @since 11.1
+     */
+    protected void migrateSessionPropertyToSecured(CoreSession session, MigrationContext migrationContext) {
+        List<DocumentModel> comments = getUnsecuredComments(session);
+        PropertyCommentManager commentManager = (PropertyCommentManager) Framework.getService(CommentManager.class);
+        processBatched(migrationContext, BATCH_SIZE, comments,
+                comment -> migrateCommentsFromPropertyToSecured(session, commentManager, comment),
+                "Migrating comments from Property to Secured");
+
+        String queryCommentsRootFolder = String.format("SELECT %s FROM Document WHERE %s='%s' AND %s='%s'", ECM_UUID,
```
efge: Spaces around `=`. This quey could be a constant like `GET_COMMENTS_FOLDERS_QUERY`.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,161 @@
...
+                ECM_NAME, COMMENTS_DIRECTORY, ECM_PRIMARYTYPE, HIDDEN_FOLDER_TYPE);
+        session.query(queryCommentsRootFolder).forEach((folder -> session.removeDocument(folder.getRef())));
```
efge: Use the `removeDocuments` API that takes all the refs at once (and can optimize the removal in some cases — or could in the future).

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/impl/TestPropertyCommentManager.java
```diff
@@ -625,13 +641,66 @@ public void testGetCommentAsRegularUser() {
...
+            johnSession.save();
+
+            DocumentModelList documents = johnSession.query(query);
+            assertEquals(2, documents.size());
+
+            assertEquals(doc.getId(), johnSession.getDocument(documents.get(0).getParentRef()).getId());
+            assertEquals(doc.getId(), johnSession.getDocument(documents.get(1).getParentRef()).getId());
+        }
+
+        // FIXME understand Browse Permission
```
efge: What's the question?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
NourNuxeo: The comments are not under the "root" comment container but under subdirectories of that folder. their names are timestamps.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/impl/TestPropertyCommentManager.java
NourNuxeo: Explenations about the difference between read and browse permission are welcome. I feel it's still unclear for me

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: it's the same query :), i change it

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/impl/TestPropertyCommentManager.java
RSalem07: because if we give the `Browse` permission to `jane` as ` acl.add(new ACE("jane", SecurityConstants.BROWSE, true));` our test failed, we expected that this permission works differently then the `READ` one

`CoreInstance.openCoreSession(doc.getRepositoryName(), "jane")` -> will get the doc with Browse permission.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/migrator/AbstractRepositoryMigrator.java
```diff
@@ -37,13 +37,25 @@
 
     private static final Log log = LogFactory.getLog(AbstractRepositoryMigrator.class);
 
+    /**
+     * @deprecated since 11.1. Not used any more as a filed of this class, you should pass it as a parameter instead
```
kevinleturc: In fact, implementation is not always responsible of passing parameters as in some cases they are called by the `MigrationService`.

Here it's more:
```
    /*
     * @deprecated since 11.1, please use parameters when using or implementing:
     * <ul>
     *   <li>{@link #migrateRepository(String, MigrationContext, String)}</li>
     *   <li>{@link #migrateSession(String, MigrationContext, CoreSession)}</li>
     *   <li>{@link #checkShutdownRequested(MigrationContext)}</li>
     * ...
     * </ul>
     */
```
I didn't list every methods involved.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/AbstractCommentManager.java
```diff
@@ -160,4 +162,38 @@ protected void setCommentPermissions(CoreSession session, DocumentModel document
...
+    /**
+     * @since 11.1
+     */
+    protected DocumentModel getThreadForComment(CoreSession coreSession, DocumentModel comment)
+            throws CommentSecurityException {
+
+        NuxeoPrincipal principal = coreSession.getPrincipal();
+        return CoreInstance.doPrivileged(coreSession, session -> {
+            DocumentModel thread = comment;
+            DocumentModel parent = session.getDocument(new IdRef((String) thread.getPropertyValue(COMMENT_PARENT_ID)));
```
kevinleturc: This field doesn't exist for previous comment implementation.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/AbstractCommentManager.java
```diff
@@ -160,4 +162,38 @@ protected void setCommentPermissions(CoreSession session, DocumentModel document
...
+            if (!session.hasPermission(principal, ancestorRef, SecurityConstants.READ)) {
+                throw new CommentSecurityException("The user " + principal.getName()
+                        + " does not have access to the comments of document " + ancestorRef.reference());
+            }
+            return thread;
+        });
+    }
+
+    @Override
+    public DocumentRef getAncestorRef(CoreSession coreSession, DocumentModel documentModel) {
```
kevinleturc: This method doesn't need to be there as you're overriding it in `CommentManagerImpl`, probably the same for `getThreadForComment`.

Furthermore, public API should be with id, otherwise we will have same issues than previous API.

Getting the last elements of `comment:ancestorId` didn't work?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,166 @@
...
         CommentServiceConfig commentServiceConfig = commentComponent.getConfig();
         if (commentServiceConfig != null) {
             Graph graph = Framework.getService(RelationManager.class).getGraph(commentServiceConfig.graphName, session);
-            if (graph.getStatements().size() > 0) {
+            if (!graph.getStatements().isEmpty()) {
                 return MIGRATION_STATE_RELATION;
             }
         }
-        return MIGRATION_STATE_PROPERTY;
+        // If not in relation, check if there are still property comments under hidden Comments folder(s)
```
kevinleturc:
```suggestion
        // If not in relation, check if there are still comments under hidden Comments folder(s)
```
or 
```suggestion
        // If not in relation, check if there are still comments (with properties) under hidden Comments folder(s)
```

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,166 @@
...
+    protected List<String> getCommentFolders(CoreSession session) {
+        List<String> parentIds = new ArrayList<>();
+
+        List<String> rootCommentsFolderIds = session.queryProjection(GET_COMMENTS_FOLDERS_QUERY, 0, 0)
+                                                    .stream()
+                                                    .map(entry -> (String) entry.get(ECM_UUID))
+                                                    .collect(Collectors.toList());
+
+        if (!rootCommentsFolderIds.isEmpty()) {
+            String query = buildAllCommentsQuery(rootCommentsFolderIds, HIDDEN_FOLDER_TYPE);
```
kevinleturc: `buildAllCommentsQuery` is confusing, here you're retrieving children of `rootCommentsFolderIds` of `HIDDEN_FOLDER_TYPE` type, which has nothing to deal with comments, method is generic.

Furthermore, a comment explaining what you fetch and in which case would be welcomed.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,166 @@
...
+            return Collections.emptyList();
+        }
+        return session.query(buildAllCommentsQuery(parentIds, COMMENT_DOC_TYPE));
+    }
+
+    /**
+     * @since 11.1
+     */
+    protected boolean hasUnsecuredComments(CoreSession session) {
+        List<String> folderIds = getCommentFolders(session);
```
kevinleturc: This may be costly to probe the session. Here we just interested in the count whereas we fetch all comments containers. This is not so bad for `PropertyCommentManager` but it could be very costly for `CommentManagerImpl` which could have a container per comment at most (one container per month).

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,166 @@
...
+        return folderIds.stream().anyMatch(folderId -> session.hasChildren(new IdRef(folderId)));
```
kevinleturc: Do we need it? We just retrieved all ids, so they suppose to exist no?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,166 @@
...
+                                           .stream()
+                                           .map(m -> new IdRef((String) m.get(ECM_UUID)))
+                                           .toArray(IdRef[]::new);
+
+        session.removeDocuments(documentsToRemove);
+
+        reportProgress("Done Migrating from Property to Secured", comments.size(), comments.size());
+    }
+
+    protected void migrateSession(String step, MigrationContext migrationContext, CoreSession session) {
```
kevinleturc: Missing `@Override` + should be the first migrate method.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -139,6 +256,36 @@ protected void migrateComments(CoreSession session, RelationManager relationMana
 
         Graph graph = relationManager.getGraph(config.graphName, session);
         graph.remove(statement);
+
+    }
+
+    /**
+     * @since 11.1
+     */
+    protected void stripACLs(DocumentModel comment) {
```
kevinleturc: Method is not needed.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -59,64 +74,166 @@
...
+     * @since 11.1
+     */
+    protected void migrateSessionPropertyToSecured(CoreSession session, MigrationContext migrationContext) {
+        List<DocumentModel> comments = getUnsecuredComments(session);
+        PropertyCommentManager commentManager = (PropertyCommentManager) Framework.getService(CommentManager.class);
+        processBatched(migrationContext, BATCH_SIZE, comments,
+                comment -> migrateCommentsFromPropertyToSecured(session, commentManager, comment),
+                "Migrating comments from Property to Secured");
+
+        session.query(GET_COMMENTS_FOLDERS_QUERY).forEach((folder -> session.removeDocument(folder.getRef())));
```
kevinleturc: Seems un-useful.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
```diff
@@ -139,6 +256,36 @@ protected void migrateComments(CoreSession session, RelationManager relationMana
...
+        acp.removeACL(LOCAL_ACL);
+        comment.setACP(acp, true);
+    }
+
+    /**
+     * @since 11.1
+     */
+    protected void migrateCommentsFromPropertyToSecured(CoreSession session, PropertyCommentManager commentManager,
+            DocumentModel comment) {
+        comment.putContextData(DISABLE_NOTIFICATION_SERVICE, TRUE); // Remove notifications
```
kevinleturc: Not sure it's needed as after this line we're using the reference, the id or ACL.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/TestCommentsMigrator.java
```diff
@@ -156,20 +171,15 @@ public void testMigrationThroughService() {
...
 
-        testMigration(() -> {
+        // Launch the step relation to property
+        runMigration(() -> {
             migrationService.runStep(MIGRATION_ID, MIGRATION_STEP_RELATION_TO_PROPERTY);
 
-            // wait a bit for the migration to start
-            sleep(1000);
+            // Wait a bit for the migration to start
+            await().atLeast(1, SECONDS);
```
kevinleturc: Could be merged to the next one with help of `pollDelay`.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
efge: Ok this is worth explaining in a comment, as a reminder of the structure

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/AbstractCommentManager.java
RSalem07: as we added the new method `org.nuxeo.ecm.platform.comment.api.CommentManager#getAncestorRef` which need the`org.nuxeo.ecm.platform.comment.impl.AbstractCommentManager#getThreadForComment`. and you right this field will not exit in the `CommentManagerImpl`. What we propos is to override `getThreadForComment` in the `CommentManagerIm` which will call its own implementation.

fyi the `CommentManagerImpl` has now its implementation of `getAncestorRef`.

kevinleturc: Ok, but in the end, what the points to have these two methods in the abstract if it's to override them in one of the two implementations ?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/AbstractCommentManager.java
RSalem07: Getting the last elements of comment:ancestorId didn't work? ->> for the `CommentManagerImpl`  didn't work, it's why we provide the two implementations.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/AbstractCommentManager.java
RSalem07: yep, we taking into account, the default methods will be removed.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/AbstractCommentManager.java
kevinleturc: We're talking about implementation for PropertyCommentManager but on code present in the abstract. You should move it back to the correct class.
After that, does the `comment:ancestorId` way (proposed by @NourNuxeo or @troger) works for property comment manager?

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: yep, please see https://github.com/nuxeo/nuxeo/pull/3289#discussion_r318810777. comment explanation will be added :)

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: actually the `org.nuxeo.ecm.platform.comment.impl.CommentsMigrator#getCommentFolders` returns the folders **Ids** (we use projection query) and if one of this folder contains children we stop iterating.

if we take the example of one comment per month we will have 12 folders per year.  what is for you the maximum threshold that be critical.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: Yes, we took the hypothesis that we can have a customers which can have some comment hidden folders that have been emptied.

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: that for cleaning the Comments folders, which will be empty at the end of our migration.
and we take the proposal of efge to use `session.removeDocuments(documentsToRemove);`

nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/main/java/org/nuxeo/ecm/platform/comment/impl/CommentsMigrator.java
RSalem07: In `org.nuxeo.ecm.core.api.AbstractSession#move(org.nuxeo.ecm.core.api.DocumentRef, org.nuxeo.ecm.core.api.DocumentRef, java.lang.String)` will dispatch different events : `ABOUT_TO_MOVE` and `DOCUMENT_MOVED`, as it's a job migration we disable notification as its done on others cases.

RSalem07:
This PR is now closed, the changes are outdated. a new one will be created to validate the new devs.

PR #3307, opened 2019-08-27
NXP-27917: add a cache to avoid slow LoginContext.login

efge:
This is experimental for now


efge:
This isn't enough, there are many other code paths where `LoginContext.login` is called

PR #3322, opened 2019-09-05
10.10 hf13/revert nxp 27683 comments permissions not set correctly

NourNuxeo:
As discussed, this is for HF 14


efge:
Wrong base, please redo this PR.

PR #3357, opened 2019-09-11
feature-NXP-27514-build-content-platform-docker-image

ataillefer:


docker/scripts/install-packages.sh
```diff
@@ -0,0 +1,45 @@
...
+EOF
+  echo
+  $NUXEO_HOME/bin/nuxeoctl mp-install --accept yes --nodeps $packages
+else
+  echo 'Found no packages to install.'
+fi
+echo
+
+# Reset Connect URL
+sed -i "/$connectUrlProp/d" $NUXEO_CONF
```
kevinleturc: Shouldn't we add `^` and `$` to prevent removal of the property declared by user ?

ataillefer: In fact, the script is executed at build time on the raw nuxeo.conf from the distribution, so it should be safe.
Besides, the line is not necessarily at the end of the file since the `mp-install` command run in the meantime has added some properties.

ataillefer:
Closed in favovr of https://github.com/nuxeo/nuxeo/pull/3451.

PR #3426, opened 2019-10-02
NXP-27441: do not package and deploy in parallel

troger:


troger:
Will be part of https://github.com/nuxeo/nuxeo/pull/3432.

PR #3431, opened 2019-10-04
NXP-27805: make S3 digest algorithm configurable

efge:


addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3BinaryManager.java
```diff
@@ -418,6 +413,8 @@ protected static boolean isMissingKey(AmazonClientException e) {
         return false;
     }
 
+    /** @deprecated since 11.1, now unused */
+    @Deprecated
     public static boolean isMD5(String digest) {
         return MD5_RE.matcher(digest).matches();
```
tmartins: Why don't you fallback on the method isValidDigest? It would allow to remove the private constant MD5_RE.

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3BinaryManager.java
```diff
@@ -162,6 +155,8 @@ protected String getDefaultDigestAlgorithm() {
 
     public static final String DELIMITER = "/";
 
+    /** @deprecated since 11.1, now unused */
+    @Deprecated
     private static final Pattern MD5_RE = Pattern.compile("[0-9a-f]{32}");
```
tmartins: Maybe this **private** constant can be removed is the method **isMD5** is rewritten (cf comment below)

efge: See above, actually not with the way the code is written now.

PR #3444, opened 2019-10-07
NXP-27683: Add check to prevent NPE when comment:parentId is not defined

RSalem07:


RSalem07:
T&P is ok: https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-saouana-10.10/36/

PR #3453, opened 2019-10-10
NXP-28123: add a builder-platform Docker image

troger:


docker/builder-platform/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.nuxeo.docker</groupId>
+    <artifactId>nuxeo-docker-images</artifactId>
+    <version>11.1-SNAPSHOT</version>
+  </parent>
+
+  <groupId>org.nuxeo.docker</groupId>
```
ataillefer: Useless :smile:

docker/builder-platform/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# It provides:
+# - A Nuxeo server distribution with appropriate permissions in the /distrib directory.
+# - A set of pre-installed packages on the server distribution
+# - An install-packages.sh script to install Nuxeo packages.
+#
+# It is based on an OpenJDK image as it requires Java for the package installation.
+#
+# It must be used within a multi-stage build.
+# For instance, you can use the following Dockerfile sample to build an image containing a Nuxeo Platform distribution:
+# - In the NUXEO_HOME directory.
```
bdelbosc: Do you mean "Build from the NUXEO_HOME directory"?

ataillefer:
Yes @bdelbosc we all agree, I'll create a task for this purpose in the next sprint.

docker/builder-platform/Dockerfile
troger: Nop, I don't think so. The distribution is in `NUXEO_HOME` in the final image of the sample:
```
COPY --from=builder --chown=UID:GID /distrib NUXEO_HOME
```

PR #3489, opened 2019-10-21
NXP-27982: refactor to allow field unrelated DocumentValidationException (global) + forward/marhsall label key

guirenard:


guirenard:
T&P OK https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-grenard-maven3/313/

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/ConstraintViolation.java
```diff
@@ -208,4 +198,18 @@ public String toString() {
...
+        // test whether there's a specific translation for for this field and this constraint
+        // the expected key is label.schema.constraint.violation.[constraintName].[schemaName].[field].[subField]
+        List<String> pathTokens = new ArrayList<>();
+        pathTokens.add(constraint.getMessageKey());
+        pathTokens.add(schema.getName());
+        for (PathNode node : path) {
+            String name = node.getField().getName().getLocalName();
+            pathTokens.add(name);
+        }
+		return StringUtils.join(pathTokens, '.');
```
troger: Format issue 😬

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/GlobalViolation.java
```diff
@@ -0,0 +1,37 @@
+package org.nuxeo.ecm.core.api.validation;
```
troger: Missing copyright

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/ValidationViolation.java
```diff
@@ -0,0 +1,17 @@
+package org.nuxeo.ecm.core.api.validation;
```
troger: Copyright

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/io/marshallers/json/validation/DocumentValidationReportJsonWriterTest.java
```diff
@@ -66,8 +67,8 @@ public void test() throws Exception {
         json.has("has_error").isTrue();
         json.has("number").isEquals(2);
         json = json.has("violations").length(2);
-        json.has(0).properties(4).has("message").isText();
-        json.has(1).properties(4).has("message").isText();
+        json.has(0).properties(5).has("message").isText();
+        json.has(1).properties(5).has("message").isText();
```
troger: Worth testing we have also the `messageKey` field here?

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/DocumentValidationException.java
```diff
@@ -38,10 +43,21 @@
     private DocumentValidationReport report;
 
     public DocumentValidationException(DocumentValidationReport report) {
-        super();
+        super(HttpServletResponse.SC_PRECONDITION_FAILED);
         this.report = report;
     }
 
+    public DocumentValidationException(String message, int statusCode) {
```
troger: Is this constructor really used/needed? It seems we always use `SC_PRECONDITION_FAILED` as `statusCode`.

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/ConstraintViolation.java
```diff
@@ -208,4 +198,18 @@ public String toString() {
 
     }
 
+    @Override
+    public String getMessageKey() {
+        // test whether there's a specific translation for for this field and this constraint
```
kevinleturc:
```suggestion
        // test whether there's a specific translation for this field and this constraint
```

nuxeo-features/rest-api/nuxeo-rest-api-test/src/test/java/org/nuxeo/ecm/restapi/listeners/DummyGlobalValidationListener.java
```diff
@@ -0,0 +1,43 @@
+package org.nuxeo.ecm.restapi.listeners;
```
kevinleturc: You're missing copyright.

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/DocumentValidationException.java
```diff
@@ -38,10 +43,21 @@
     private DocumentValidationReport report;
 
     public DocumentValidationException(DocumentValidationReport report) {
-        super();
+        super(HttpServletResponse.SC_PRECONDITION_FAILED);
```
kevinleturc: Why changing `BAD_REQUEST` (400) to `PRECONDITION_FAILED` (412) ? It seems the second one is not adapted to our case, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412. It seems more adapted in cases such as `changeToken` has changed.

I think we're more in a `400` case, but if you want to distinguish this validation error leveraging the http status code, `422` seems more appropriate to what we do. Wdyt ?

guirenard: Indeed, there's a pending debate, I was hesitate with `422`

kevinleturc: Let's go for `422` then ;)

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/DocumentValidationException.java
troger: To come back to this comment... after thinking, I think a validation error is *always* `422` (SC_UNPROCESSABLE_ENTITY).
The entity is always readable, but the server refuses because of validation errors.

`411` (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/411) for instance has nothing to do with our validation errors.

efge: I agree that `SC_PRECONDITION_FAILED` doesn't seem to me the right error.
Note that CMIS uses `SC_CONFLICT` (409) for most errors having to do with constraints or duplicates. I think it could fit validation errors.

efge: FYI CMIS uses `SC_CONFLICT` for `CmisConstraintException`, `CmisContentAlreadyExistsException`, `CmisNameConstraintViolationException`, `CmisUpdateConflictException`, `CmisVersioningException. Cf https://github.com/nuxeo/nuxeo/blob/master/addons/nuxeo-chemistry/nuxeo-opencmis-bindings/src/main/java/org/nuxeo/ecm/core/opencmis/bindings/DefaultErrorExtractor.java#L119

efge: And we already use it for `ConcurrentUpdateException` and `DocumentExistsException`

guirenard: @efge  It is acknowledged that `SC_PRECONDITION_FAILED` is not the correct one. Here we are to return `SC_UNPROCESSABLE_ENTITY` (422) in case of validation exception thrown by default (see https://github.com/nuxeo/nuxeo/pull/3489#discussion_r338478093).

Here @troger is rather asking if we want to let the possibility to return something else than `422` in custom code.

efge: Ok I see what you mean. Yes you should let the constructor open to providing a different http code, from what you can see above I think it's entirely possible that we'll want custom code to return `409` or even something else.

troger: Still not sure a `DocumentValidationException` triggered by our `DocumentValidationService` could be a `409` or something else but well, we can still keep the constructor even if not used.

guirenard: It'd would rather be thrown by custom code by some custom code such an inline core listener on `aboutToCreate` or `beforeDocumentModification` event

troger: `DocumentValidationException` seems really tied to `ConstraintViolation` and so on our `DocumentValidationService` ... I would use `NuxeoException` or some other custom Exception for `aboutToCreate` or `beforeDocumentModification` 😛

efge: Indeed custom code can throw other exceptions. Let's not complicate for now, just use 422 always then and not make it configurable.

guirenard: Ok

PR #3560, opened 2019-11-21
NXP-24873: avoid css overriding by @media to fix 'login box backgroun…

ThierryCasanova:
…d color customization does not work'

(changed mobile device styles position in login.jsp)


bdelbosc:
For the next time :) 
The seven rules of a great Git commit message:
    Separate subject from body with a blank line.
    Limit the subject line to 50 characters. (this is hard)
    Capitalize the subject line.
    Do not end the subject line with a period.
    Use the imperative mood in the subject line.
    Wrap the body at 72 characters.
    Use the body to explain what and why vs. how.

PR #3580, opened 2019-11-28
8.10 hf/fix nxp 24873 login box background color customization does not work

ThierryCasanova:


ThierryCasanova:
@NourNuxeo : done on 8.10 : kept only one commit and simplified commit message

PR #3589, opened 2019-11-29
Fix nxp 27559 cq fix

bdelbosc:
Bump the Chronicle version and re-enable CQ unit tests to see if it helps.


bdelbosc:
This PR is to trigger JX no review needed at this stage.

PR #3606, opened 2019-12-10
NXP-27351: Fix timeout command in nuxeoctl.bat

BoboTiG:
`nuxeoctl.bat` used to use the `timeout.exe` file from Windows.
However, when Cygwin is installed, the executable used is ambiguous because `timeout.exe` exists in the "C:\Windows\System32" folder and in the Cygwin binaries one.

This causes `timeout.exe` to fail with the following error:
```batch
timeout: invalid time interval ‘/t’
Try 'timeout --help' for more information.
```

The chosen fix is to use the full path of the executable.

---
**EDIT**: first proposed fix.

The fix is to move to `ping`:
```batch
ping 127.0.0.1 -n 31 > NUL
```
Here, 31 is needed to wait 30 seconds as the first ping will not wait before being done. And as `ping` waits 1 second between attempts, we have the same expected behavior as before.

Using `ping` is reliable and a classical move in such situations.

### DIFF
--- a/nuxeo-distribution/nuxeo-server-tomcat/src/main/resources/tomcat/bin/nuxeoctl.bat
+++ b/nuxeo-distribution/nuxeo-server-tomcat/src/main/resources/tomcat/bin/nuxeoctl.bat
@@ -14,6 +14,7 @@ rem ## Lesser General Public License for more details.
 rem ##
 rem ## Contributors:
 rem ##     Julien Carsique
+rem ##     Mickael Schoentgen
 rem ##
 rem ## DOS script calling a multi-OS Nuxeo Java launcher
 rem ##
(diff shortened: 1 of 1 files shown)


nuxeo-distribution/nuxeo-server-tomcat/src/main/resources/tomcat/bin/nuxeoctl.bat
```diff
@@ -64,14 +65,14 @@ if exist "%NUXEO_CONF%" goto FOUND_NUXEO_CONF
 
 REM ***** All checks failed *****
 echo Could not find nuxeo.conf in the path, the environment or the registry
-timeout /t 30
+ping 127.0.0.1 -n 31 > NUL
```
efge: Please add a small comment explaining why this is done, for future maintainers.

BoboTiG: I did it on the 1st occurrence. LMK if it is OK like that.

PR #3617, opened 2019-12-18
fix-NXP-28421-download-filename-encoding

efge:
PR: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/738/ (test failure is unrelated)

### DIFF
--- a/nuxeo-common/src/main/java/org/nuxeo/common/utils/RFC2231.java
+++ b/nuxeo-common/src/main/java/org/nuxeo/common/utils/RFC2231.java
@@ -32,6 +32,14 @@ import static org.nuxeo.common.utils.UserAgentMatcher.isMSIE6or7;
  */
 public class RFC2231 {
 
+    // RFC 2045
+    private static final String MIME_SPECIALS = "()<>@,;:\\\"/[]?=\t ";
+
+    // RFC 2231
+    private static final String RFC2231_SPECIALS = "*'%" + MIME_SPECIALS;
+
+    private static final char[] HEX_DIGITS = "0123456789ABCDEF".toCharArray();
+
     // Utility class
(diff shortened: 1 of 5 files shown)


nuxeo-common/src/main/java/org/nuxeo/common/utils/RFC2231.java
```diff
@@ -59,6 +66,38 @@ public static void percentEscape(StringBuilder sb, String value) {
...
+                sb.append((char) c);
+            }
+        }
+        if (!encoded) {
+            // undo and use basic format
+            sb.setLength(originalLength);
+            sb.append('=');
+            sb.append(value);
+        }
+    }
```
ataillefer: Wouldn't it be clearer to use another StringBuilder `valueSb` for the possibly encoded value, then have something simple afterwards, like:
```
if (encoded) {
    sb.append("*=UTF-8''"); // no language
    sb.append(valueSb.toString());
} else {
    sb.append('=');
    sb.append(value);
}
```

efge: That was my initial code but it's an anti-pattern to have two StringBuilders and add one to the other. The `setLength` trick allows an easy "reset" to avoid allocating a new StringBuilder.

ataillefer: It may be an anti-pattern but I still find it more explicit when reading the code.

efge: I think it's a matter of habit... It's useful to be able to "go back" in a StringBuilder.

PR #3618, opened 2019-12-18
NXP-28437: S3 probe should not report failure when there are not S3 blob providers

efge:


addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/status/S3BinaryManagerStatusProbe.java
```diff
@@ -38,24 +38,15 @@
...
                 return ProbeStatus.newFailure("S3BinaryManager cannot access the configured bucket: " + id);
             }
-            someSuccess = true;
-        }
-        if (someSuccess) {
-            return ProbeStatus.newSuccess("S3BinaryManager can access the configured buckets");
-        } else {
-            return ProbeStatus.newFailure("No S3BinaryManager bucket configured");
         }
+        return ProbeStatus.newSuccess("S3BinaryManager can access the configured buckets");
```
troger: Isn't it weird to have "S3BinaryManager can access the configured buckets" as status message if there is no S3BinaryManager at all?

efge: Yes it's a bit weird but I didn't know if anything could be parsing these messages. But indeed given that we're changing the behavior if there's no binary manager we might as well cleanup the message. I'll do that.

efge: @troger now updated

PR #3619, opened 2019-12-18
NXP-28437: S3 probe should not report failure when there are not S3 blob providers (10.10)

efge:


efge:
@troger @kevinleturc ok with this last change that uses a different message when there is no blob provider configured?

troger:
Why not just using a `boolean`? 🤔

efge:
I wanted to include the number of blob providers in the message if it's > 0 but I didn't do it in the end. Yes a boolean could work, similar to what was there before actually. I'll do the change.

efge:
@troger @kevinleturc @ataillefer in the end I'm simplifying this particular patch for the bugfix, it will be slightly refactored anyway for NXP-28276.

PR #3632, opened 2019-12-24
NXP-28406: Handle duplicates with MongoDB

RSalem07:
Manually opened PR as Draft for feedbacks.

Perhaps i missed somethings, but there is some points that i observed when i did the developments, its why i opened as a draft (no T&P started) 

- My first idea was to create a dedicate test from MondoDBBackEnd like [here](https://github.com/nuxeo/nuxeo/blob/master/nuxeo-core/nuxeo-core-storage-sql/nuxeo-core-storage-sql-test/src/test/java/org/nuxeo/ecm/core/storage/sql/TestSQLBackend.java#L1105) in `org.nuxeo.ecm.core.storage.mongodb`. But we need the `CoreFeatue` (which exists under ` nuxeo-core-test` which needs  `org.nuxeo.ecm.core.storage.mongodb` to build `org.nuxeo.ecm.core.test.StorageConfiguration` and we find ourself with a cyclic maven dependency module.

- Actually in junit tests context if we need to add more configuration on `MongoDBRepositoryDescriptor` like `nativeId`, `sequenceBlockSize` and the new one `childNameUniqueConstraintEnabled` we have to modify [test-storage-repo-mongodb-contrib.xml](https://github.com/nuxeo/nuxeo/blob/master/nuxeo-core/nuxeo-core-test/src/main/resources/OSGI-INF/test-storage-repo-mongodb-contrib.xml) for whole mondodb tests, i didn't find a way to override it at demand, there is no merge of contribution in  [MongoDBRepositoryService](https://github.com/nuxeo/nuxeo/blob/master/nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepositoryService.java). Do we need to manage the merge ? should we be do it on this ticket or not ? if we choice to do it we should pay attention on [here](https://github.com/nuxeo/nuxeo/blob/master/nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java#L210)

- Perhaps another idea is to group the VCS (case Postgres, TestSQLBackend) and DBS (case mongo) the check on the unicity and the DB unique index, if its what we want. where (under which project) add theses tests 

T&P on H2 -> OK => https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-saouana/164/

T&P on MongoDB -> OK  => https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-saouana/170/

### DIFF
--- a/nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
+++ b/nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
@@ -1,0 +1,125 @@
+/*
+ * (C) Copyright 2019 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Salem Aouana
+ */
+
+package org.nuxeo.ecm.core.mongodb;
+
+import static javax.servlet.http.HttpServletResponse.SC_CONFLICT;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.nuxeo.runtime.mongodb.MongoDBFeature.DEFAULT_MONGODB_DBNAME;
+
+import javax.inject.Inject;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.nuxeo.ecm.core.api.CloseableCoreSession;
+import org.nuxeo.ecm.core.api.ConcurrentUpdateException;
+import org.nuxeo.ecm.core.api.CoreInstance;
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentExistsException;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.security.ACE;
+import org.nuxeo.ecm.core.api.security.ACL;
+import org.nuxeo.ecm.core.api.security.SecurityConstants;
+import org.nuxeo.ecm.core.api.security.impl.ACPImpl;
+import org.nuxeo.ecm.core.test.CoreFeature;
+import org.nuxeo.runtime.mongodb.MongoDBFeature;
+import org.nuxeo.runtime.test.runner.Deploy;
+import org.nuxeo.runtime.test.runner.Features;
+import org.nuxeo.runtime.test.runner.FeaturesRunner;
+
+/**
+ * @since 11.1
+ */
+@RunWith(FeaturesRunner.class)
+@Features({ MongoDBFeature.class, CoreFeature.class })
+@Deploy("org.nuxeo.ecm.core.mongodb.test:OSGI-INF/mongodb-disable-renaming-listener-test.xml")
+@Deploy("org.nuxeo.ecm.core.mongodb.test:OSGI-INF/mongodb-enable-unique-indices-contrib.xml")
+public class TestMongoDBIndices {
+
+    protected static final String DOCUMENT_NAME = "anyFile";
+
+    @Inject
+    protected CoreSession coreSession;
+
+    protected DocumentModel folder;
+
+    @Before
+    public void before() {
+        folder = coreSession.createDocumentModel("/", "Folder", "Folder");
+        folder = coreSession.createDocument(folder);
+        coreSession.save();
+        addPermissions(folder, "james");
+        addPermissions(folder, "bob");
+    }
+
+    @Test
+    public void shouldFailWhenCreatingExistingChildNameDocument() {
+        try (CloseableCoreSession bobSession = CoreInstance.openCoreSession(coreSession.getRepositoryName(), "bob")) {
+            createDocument(bobSession, folder, DOCUMENT_NAME);
+        }
+
+        try (CloseableCoreSession jamesSession = CoreInstance.openCoreSession(coreSession.getRepositoryName(),
+                "james")) {
+            createDocument(jamesSession, folder, DOCUMENT_NAME);
+            fail("should throw a ConcurrentUpdateException");
+        } catch (ConcurrentUpdateException cue) {
+            assertEquals(SC_CONFLICT, cue.getStatusCode());
+            assertTrue(
+                    cue.getMessage()
+                       .contains(String.format(
+                               "E11000 duplicate key error collection: %s.test index: ecm:parentId_1_ecm:name_1 dup key: { : \"%s\", : \"%s\" }",
+                               DEFAULT_MONGODB_DBNAME, folder.getId(), DOCUMENT_NAME)));
+        }
+    }
+
+    @Test
+    public void shouldFailWhenMovingDocumentToExistingChildName() {
+        try (CloseableCoreSession bobSession = CoreInstance.openCoreSession(coreSession.getRepositoryName(), "bob")) {
+            createDocument(bobSession, folder, DOCUMENT_NAME);
+        }
+
+        try (CloseableCoreSession jamesSession = CoreInstance.openCoreSession(coreSession.getRepositoryName(),
+                "james")) {
+            DocumentModel documentModel = createDocument(jamesSession, folder, "jamesFileName");
+
+            // DBSession.move will check the unicity and throw a DocumentExistsException before calling the backend
+            jamesSession.move(documentModel.getRef(), null, DOCUMENT_NAME);
+            fail("should throw a DocumentExistsException");
+        } catch (DocumentExistsException dee) {
+            assertEquals(String.format("Destination name already exists: %s", DOCUMENT_NAME), dee.getMessage());
+        }
+    }
+
+    protected DocumentModel createDocument(CoreSession session, DocumentModel parent, String fileName) {
+        DocumentModel doc = session.createDocumentModel(parent.getPathAsString(), fileName, "File");
+        session.createDocument(doc);
+        session.save();
+        return doc;
+    }
+
+    protected void addPermissions(DocumentModel documentModel, String userName) {
+        ACPImpl acp = new ACPImpl();
+        ACL acl = acp.getOrCreateACL();
+        acl.add(new ACE(userName, SecurityConstants.READ_WRITE, true));
+        coreSession.setACP(documentModel.getRef(), acp, false);
+    }
+}
--- a/nuxeo-core/nuxeo-core-mongodb/src/test/resources/OSGI-INF/mongodb-disable-renaming-listener-test.xml
+++ b/nuxeo-core/nuxeo-core-mongodb/src/test/resources/OSGI-INF/mongodb-disable-renaming-listener-test.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0"?>
+
+<component name="org.nuxeo.mongodb.duplicate.key.test.contrib" version="1.0">
+  <extension target="org.nuxeo.ecm.core.event.EventServiceComponent"
+             point="listener">
+    <listener name="duplicatedNameFixer" enabled="false">
+    </listener>
+  </extension>
+
+</component>
--- a/nuxeo-core/nuxeo-core-mongodb/src/test/resources/OSGI-INF/mongodb-enable-unique-indices-contrib.xml
+++ b/nuxeo-core/nuxeo-core-mongodb/src/test/resources/OSGI-INF/mongodb-enable-unique-indices-contrib.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0"?>
+<component name="org.nuxeo.ecm.core.test.mongodb.repository.indices">
+  <extension target="org.nuxeo.ecm.core.storage.mongodb.MongoDBRepositoryService" point="repository">
+    <repository name="test">
+      <childNameUniqueConstraintEnabled>true</childNameUniqueConstraintEnabled>
+    </repository>
+  </extension>
+
+</component>
--- a/nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSTransactionState.java
+++ b/nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSTransactionState.java
@@ -1186,7 +1186,7 @@ public class DBSTransactionState {
         }
 
         protected void findDirtyPaths(ListDiff value, String path) {
-            String newPath = path + "/*";
+            String newPath = path;
             if (value.diff != null) {
                 findDirtyPaths(value.diff, newPath);
             }
--- a/nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
+++ b/nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
@@ -18,6 +18,8 @@
  */
 package org.nuxeo.ecm.core.storage.mongodb;
 
+import static com.mongodb.ErrorCategory.DUPLICATE_KEY;
+import static com.mongodb.ErrorCategory.fromErrorCode;
 import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.KEY_ACE_STATUS;
 import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.KEY_ACE_USER;
 import static org.nuxeo.ecm.core.storage.dbs.DBSDocument.KEY_ACL;
@@ -85,7 +87,11 @@ import org.nuxeo.runtime.api.Framework;
 import org.nuxeo.runtime.mongodb.MongoDBConnectionService;
 
 import com.mongodb.Block;
+import com.mongodb.DuplicateKeyException;
+import com.mongodb.MongoBulkWriteException;
+import com.mongodb.MongoWriteException;
 import com.mongodb.QueryOperators;
+import com.mongodb.bulk.BulkWriteError;
 import com.mongodb.client.FindIterable;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoCursor;
@@ -191,7 +197,7 @@ public class MongoDBRepository extends DBSRepositoryBase {
         }
         converter = new MongoDBConverter(useCustomId ? null : KEY_ID);
         cursorService = new CursorService<>(ob -> (String) ob.get(converter.keyToBson(KEY_ID)));
-        initRepository();
+        initRepository(descriptor);
     }
 
     @Override
@@ -205,7 +211,21 @@ public class MongoDBRepository extends DBSRepositoryBase {
         cursorService.clear();
     }
 
+    /**
+     * @deprecated since 11.1. use {@link #initRepository(MongoDBRepositoryDescriptor)} instead.
+     */
+    @Deprecated(since = "11.1", forRemoval = true)
     protected void initRepository() {
+        initRepository(null);
+    }
+
+    /**
+     * Initializes the MongoDB repository
+     * 
+     * @param descriptor the MongoDB repository descriptor
+     * @since 11.1
+     */
+    protected void initRepository(MongoDBRepositoryDescriptor descriptor) {
         // check root presence
         if (coll.countDocuments(Filters.eq(idKey, getRootId())) > 0) {
             return;
@@ -221,7 +241,11 @@ public class MongoDBRepository extends DBSRepositoryBase {
         coll.createIndex(Indexes.ascending(KEY_PROXY_TARGET_ID));
         coll.createIndex(Indexes.ascending(KEY_PROXY_VERSION_SERIES_ID));
         coll.createIndex(Indexes.ascending(KEY_READ_ACL));
-        coll.createIndex(Indexes.ascending(KEY_PARENT_ID, KEY_NAME));
+        IndexOptions parentNameIndexOptions = new IndexOptions();
+        if (descriptor != null) {
+            parentNameIndexOptions.unique(Boolean.TRUE.equals(descriptor.getChildNameUniqueConstraintEnabled()));
+        }
+        coll.createIndex(Indexes.ascending(KEY_PARENT_ID, KEY_NAME), parentNameIndexOptions);
         // often used in user-generated queries
         coll.createIndex(Indexes.ascending(KEY_PRIMARY_TYPE));
         coll.createIndex(Indexes.ascending(KEY_LIFECYCLE_STATE));
@@ -294,9 +318,12 @@ public class MongoDBRepository extends DBSRepositoryBase {
     public void createState(State state) {
         Document doc = converter.stateToBson(state);
         log.trace("MongoDB: CREATE {}: {}", doc.get(idKey), doc);
-        coll.insertOne(doc);
-        // TODO dupe exception
-        // throw new DocumentException("Already exists: " + id);
+        try {
+            coll.insertOne(doc);
+        } catch (DuplicateKeyException dke) {
+            log.trace("MongoDB:    -> DUPLICATE KEY: {}", doc.get(idKey));
+            throw new ConcurrentUpdateException(dke);
+        }
     }
 
     @Override
@@ -305,7 +332,24 @@ public class MongoDBRepository extends DBSRepositoryBase {
         log.trace("MongoDB: CREATE [{}]: {}",
                 () -> docs.stream().map(doc -> doc.get(idKey).toString()).collect(Collectors.joining(", ")),
                 () -> docs);
-        coll.insertMany(docs);
+        try {
+            coll.insertMany(docs);
+        } catch (MongoBulkWriteException mbwe) {
+            List<String> duplicates = mbwe.getWriteErrors()
+                                          .stream()
+                                          .filter(wr -> DUPLICATE_KEY.equals(fromErrorCode(wr.getCode())))
+                                          .map(BulkWriteError::getMessage)
+                                          .collect(Collectors.toList());
+            // Avoid hiding any others bulk errors
+            if (duplicates.size() == mbwe.getWriteErrors().size()) {
+                log.trace("MongoDB:    -> DUPLICATE KEY: {}", duplicates);
+                ConcurrentUpdateException concurrentUpdateException = new ConcurrentUpdateException();
+                duplicates.forEach(concurrentUpdateException::addInfo);
+                throw concurrentUpdateException;
+            }
+
+            throw mbwe;
+        }
     }
 
     @Override
@@ -347,13 +391,19 @@ public class MongoDBRepository extends DBSRepositoryBase {
                 log.trace("MongoDB: UPDATE {}: IF {} THEN {}", id, conditions, update);
                 filter.putAll(conditions);
             }
-            UpdateResult w = coll.updateMany(filter, update);
-            if (w.getModifiedCount() != 1) {
-                log.trace("MongoDB:    -> CONCURRENT UPDATE: {}", id);
-                throw new ConcurrentUpdateException(id);
+            try {
+                UpdateResult w = coll.updateMany(filter, update);
+                if (w.getModifiedCount() != 1) {
+                    log.trace("MongoDB:    -> CONCURRENT UPDATE: {}", id);
+                    throw new ConcurrentUpdateException(id);
+                }
+            } catch (MongoWriteException mwe) {
+                if (DUPLICATE_KEY.equals(fromErrorCode(mwe.getCode()))) {
+                    log.trace("MongoDB:    -> DUPLICATE KEY: {}", id);
+                    throw new ConcurrentUpdateException(mwe.getError().getMessage(), mwe);
+                }
+                throw mwe;
             }
-            // TODO dupe exception
-            // throw new DocumentException("Missing: " + id);
         }
     }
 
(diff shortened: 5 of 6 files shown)


nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/configuration/TestMongoDBBackEnd.java
```diff
@@ -0,0 +1,161 @@
...
+import org.nuxeo.ecm.core.api.security.impl.ACPImpl;
+import org.nuxeo.ecm.core.test.CoreFeature;
+import org.nuxeo.runtime.api.Framework;
+import org.nuxeo.runtime.mongodb.MongoDBConnectionService;
+import org.nuxeo.runtime.mongodb.MongoDBFeature;
+import org.nuxeo.runtime.test.runner.Deploy;
+import org.nuxeo.runtime.test.runner.Features;
+import org.nuxeo.runtime.test.runner.FeaturesRunner;
+
+import com.google.inject.Inject;
```
efge: `javax.inject.Inject`

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/configuration/TestMongoDBBackEnd.java
```diff
@@ -0,0 +1,161 @@
...
+
+    protected static final String DOCUMENT_NAME = "anyFile";
+
+    @Inject
+    protected CoreSession coreSession;
+
+    @Test
+    public void shouldCreateUniqueIndex() {
+        MongoDBConnectionService mongoService = Framework.getService(MongoDBConnectionService.class);
+        MongoDatabase database = mongoService.getDatabase(REPOSITORY_CONNECTION_PREFIX + "test");
```
efge: I don't think this kind of white-box test is useful. You're basically testing that you correctly wrote code to create the unique index, but this is not that useful. What's useful is to test the _consequence_ of the index, that we have an exception when we attempt to create two identical parentid+childname.
So I'd remove this test.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/configuration/TestMongoDBBackEnd.java
```diff
@@ -0,0 +1,161 @@
...
+        assertNotNull(document);
+        assertTrue((Boolean) document.get("unique"));
+    }
+
+    @Test
+    public void shouldFailsWhenCreateExistingChildNameDocument() {
+        DocumentModel domain = coreSession.createDocumentModel("/", "Folder", "Folder");
+        domain = coreSession.createDocument(domain);
+        addPermissions(domain, "james");
+        addPermissions(domain, "bob");
```
efge: In what way is the permission stuff relevant to the test, which is about uniqueness of parentid+childname? I'd remove all the ACL stuff and the use of various sessions for different users.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/configuration/TestMongoDBBackEnd.java
```diff
@@ -0,0 +1,161 @@
...
+            assertTrue(
+                    cue.getMessage()
+                       .contains(String.format(
+                               "E11000 duplicate key error collection: %s.test index: ecm:parentId_1_ecm:name_1 dup key: { : \"%s\", : \"%s\" }",
+                               DEFAULT_MONGODB_DBNAME, domain.getId(), DOCUMENT_NAME)));
+        }
+    }
+
+    @Test
+    public void shouldFailsWhenUpdateExistingChildNameDocument() {
```
efge: I took a long time to understand that by "Update" you refer to the update of the internal "ecm:name" of the doc. I'd rename this to use "Move" instead of "Update".

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
```diff
@@ -221,7 +240,11 @@ protected void initRepository() {
         coll.createIndex(Indexes.ascending(KEY_PROXY_TARGET_ID));
         coll.createIndex(Indexes.ascending(KEY_PROXY_VERSION_SERIES_ID));
         coll.createIndex(Indexes.ascending(KEY_READ_ACL));
-        coll.createIndex(Indexes.ascending(KEY_PARENT_ID, KEY_NAME));
+        IndexOptions parentNameIndexOptions = new IndexOptions();
+        if (!isNull(descriptor)) {
```
efge: `!isNull(descriptor)` -> `descriptor != null`

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
```diff
@@ -221,7 +240,11 @@ protected void initRepository() {
...
+            parentNameIndexOptions.unique(isTrue(descriptor.getChildNameUniqueConstraintEnabled()));
```
efge: `Boolean.TRUE.equals(...)` is enough, no need to import a helper for this. Also it makes it explicit that we're interested in `Boolean.TRUE` and not just the basic `true`.

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
```diff
@@ -305,7 +331,27 @@ public void createStates(List<State> states) {
...
                 () -> docs.stream().map(doc -> doc.get(idKey).toString()).collect(Collectors.joining(", ")),
                 () -> docs);
-        coll.insertMany(docs);
+        try {
+            coll.insertMany(docs);
+        } catch (MongoBulkWriteException mbwe) {
+            //FIXME extract constant / find the list of mongodb code error (11000)
+            //https://jira.mongodb.org/browse/DOCS-335 (Won't Fix)
+            //https://docs.mongodb.com/manual/core/index-unique/
+            final List<String> duplicates = mbwe.getWriteErrors()
```
efge: Remove unneeded `final`.

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
```diff
@@ -305,7 +331,27 @@ public void createStates(List<State> states) {
...
+                                                .stream()
+                                                .filter(wr -> wr.getCode() == 11000)
+                                                .map(BulkWriteError::getMessage)
+                                                .collect(Collectors.toList());
+            // Avoid hiding any others bulk errors
+            if (duplicates.size() == mbwe.getWriteErrors().size()) {
+                log.trace("MongoDB:    -> CONCURRENT UPDATE: {}", duplicates);
+                final ConcurrentUpdateException concurrentUpdateException = new ConcurrentUpdateException();
```
efge: Remove unneeded `final`.

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
```diff
@@ -305,7 +331,27 @@ public void createStates(List<State> states) {
...
+                duplicates.stream().forEach(concurrentUpdateException::addInfo);
```
efge: `duplicates.forEach(...)`

efge:
Regarding your initial comment https://github.com/nuxeo/nuxeo/pull/3632#issue-356620746 yes our testing sorry is not ideal, and we don't have for DBS an equivalent of `TestSQLBackend`.

What I'd like to have in the future is actually a test at the `org.nuxeo.ecm.core.model.Session`/`org.nuxeo.ecm.core.model.Document` level because that's what makes most sense.

Running unit tests with variable configuration for the repository-level config options is also something for which we don't have a clean answer.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/configuration/TestMongoDBBackEnd.java
RSalem07: yes i agree with your point view of we `What's useful is to test the consequence of the index,` and i asked myself the same question, but i did it to ensure that the whole process  is correctly handled when i add the contribution.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/configuration/TestMongoDBBackEnd.java
RSalem07: sorry for that i completely forgot that we cannot just update the name, and to do it we should use move and at the end of the test  i forgot to rename the test method.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/configuration/TestMongoDBBackEnd.java
RSalem07: i added because i had an exception `Privilege 'AddChildren' is not granted to 'bob'` when i tried to create the document [here](https://github.com/nuxeo/nuxeo/pull/3632/files/c91239adfefe45b863c6eb6b4f90bebf2b914289#diff-4f5ec1ee49aed6c5699ddd058d91fa6eR107)

RSalem07:
I took into account the the code review i let's a comment about the added permissions. i will launch a T&P and i will change the status to ready for review.


Regarding your comment https://github.com/nuxeo/nuxeo/pull/3632#issuecomment-569131598 i rename the test class to `TestMongoDBIndices` and i changed his package to be more closer with the test context ... 

finally i have an interrogation about `childNameUniqueConstraintEnabled` in the `test-storage-repo-mongodb-contrib.xml`: we can keep the configuration as i did it, i am wondering if with this configuration if we will have a performances issues(more slower...) in the context the tests with Mongodb profile.

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
```diff
@@ -205,7 +211,20 @@ public void shutdown() {
         cursorService.clear();
     }
 
+    /**
+     * @deprecated since 11.1. use {@link #initRepository(MongoDBRepositoryDescriptor)} instead.
+     */
     protected void initRepository() {
```
kevinleturc: Missing `@Deprecated`.

nuxeo-core/nuxeo-core-test/src/main/resources/OSGI-INF/test-storage-repo-mongodb-contrib.xml
```diff
@@ -15,6 +15,7 @@
       <sequenceBlockSize>1</sequenceBlockSize>
       <nativeId>false</nativeId>
       -->
+      <childNameUniqueConstraintEnabled>true</childNameUniqueConstraintEnabled>
```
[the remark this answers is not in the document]
RSalem07: initially i added a new contribution to enable the unique index only on my class test but it didn't work. Its why i put it on the global configuration (https://github.com/nuxeo/nuxeo/pull/3632#issue-356620746). Now i retry it with new xml contribution and its work correctly i don't know what i did wrong the first time.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
```diff
@@ -0,0 +1,128 @@
...
+    @Inject
+    protected CoreSession coreSession;
+
+    @Test
+    public void shouldFailsWhenCreateExistingChildNameDocument() {
+        DocumentModel domain = coreSession.createDocumentModel("/", "Folder", "Folder");
+        domain = coreSession.createDocument(domain);
+        addPermissions(domain, "james");
+        addPermissions(domain, "bob");
+        coreSession.save();
```
kevinleturc: You can move it to a `@Before` method.

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
```diff
@@ -294,9 +318,12 @@ public String generateNewId() {
...
         Document doc = converter.stateToBson(state);
         log.trace("MongoDB: CREATE {}: {}", doc.get(idKey), doc);
-        coll.insertOne(doc);
-        // TODO dupe exception
-        // throw new DocumentException("Already exists: " + id);
+        try {
+            coll.insertOne(doc);
+        } catch (DuplicateKeyException dke) {
+            log.trace("MongoDB:    -> DUPLICATE KEY: {}", doc.get(idKey));
+            throw new ConcurrentUpdateException(dke);
```
kevinleturc: Why not using `DocumentExistsException`? As it, we will be consistent with the `CoreSession#move` API.

RSalem07: The idea behind this test is to ensure that the MongoDB will throw a concurrentUpdateException when a unique index is defined (here on parent and his child name). When we insert the call directly the backend (mongo) there is no check in Nuxeo code before insert. On other hand when we move there is a check on the name it's why on move i just kept the test and i check on `DocumentExistsException`

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
```diff
@@ -0,0 +1,128 @@
...
+        try (CloseableCoreSession bobSession = CoreInstance.openCoreSession(coreSession.getRepositoryName(), "bob")) {
+            createDocument(bobSession, domain, DOCUMENT_NAME);
+        }
+
+        try (CloseableCoreSession jamesSession = CoreInstance.openCoreSession(coreSession.getRepositoryName(),
+                "james")) {
+            DocumentModel documentModel = createDocument(jamesSession, domain, "jamesFileName");
+
+            // org.nuxeo.ecm.core.storage.dbs.DBSSession.move (Objects.equals(oldParentId, parentId)) checks the unique
+            // name before calling the mongodb update backend
```
kevinleturc:
```java
// DBSession.move will check the unicity and throw DocumentExistsException before calling backend
```

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
```diff
@@ -0,0 +1,128 @@
...
+            documentModel = jamesSession.move(documentModel.getRef(), null, DOCUMENT_NAME);
+            jamesSession.saveDocument(documentModel);
+            jamesSession.save();
```
kevinleturc: Isn't needed no?

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
```diff
@@ -0,0 +1,128 @@
...
+            fail("should throw a DocumentExistsException");
+        } catch (DocumentExistsException dee) {
+            assertEquals(String.format("Destination name already exists: %s", DOCUMENT_NAME), dee.getMessage());
+        }
+    }
+
+    protected DocumentModel createDocument(CoreSession session, DocumentModel parent, String fileName) {
+        DocumentModel doc = session.createDocumentModel(parent.getPathAsString(), fileName, "File");
+        session.createDocument(doc);
+        session.save();
```
kevinleturc: Not needed

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
RSalem07: yes i agree, i just kept it :).

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
RSalem07: you mean the whole method, if it's the case i created it just to be more concise about the test and keep the creation process outer and to avoid duplicate of some code.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
kevinleturc: Why?

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
kevinleturc: No, the line. You're doing a save at the end of `@Before` method, this one should be enough, if not due to ACLs, put save before adding ACLs.

Furthermore, AFAIK it's better to get the document from createDocument as it may contain data that the input won't have.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
RSalem07: it's was mainly to have the same scenario then the creation where i create the document (the update is done by move in this case) and then i save it and the session juste after.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
RSalem07: i test with removing the `session.save()`  on `createDocument ` and my test failed because the second creation will be able to put the same name as our session is not saved. for your other suggestion of adding `save` before `addPermission` i did it

nuxeo-core/nuxeo-core-mongodb/src/test/resources/OSGI-INF/mongodb-disable-renaming-listener-test.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+
+<component name="org.nuxeo.mongodb.duplicate.key.test.contrib" version="1.0">
+   <extension target="org.nuxeo.ecm.core.event.EventServiceComponent"
+    point="listener">
+    <listener name="duplicatedNameFixer"
```
troger: Why not just using `enabled=false` to disable the listener, as it's stated in the file name?

Worth changing the component name to reflect you are contributing to listeners?

nuxeo-core/nuxeo-core-mongodb/src/test/resources/OSGI-INF/mongodb-enable-unique-indices-contrib.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+<component name="org.nuxeo.ecm.core.test.mongodb.repository.indices">
+  <extension target="org.nuxeo.ecm.core.storage.mongodb.MongoDBRepositoryService" point="repository">
+    <repository name="test" label="MongoDB Repository">
```
troger: You can remove the `label` here as you are merging with an existing contribution.

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
```diff
@@ -205,7 +211,21 @@ public void shutdown() {
...
+    /**
+     * @deprecated since 11.1. use {@link #initRepository(MongoDBRepositoryDescriptor)} instead.
+     */
+    @Deprecated(since = "11.1", forRemoval = true)
     protected void initRepository() {
+        initRepository(null);
+    }
+
+    /**
+     * Initialises the mongo db repository
```
troger:
```suggestion
     * Initializes the MongoDB repository
```

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
```diff
@@ -205,7 +211,21 @@ public void shutdown() {
...
+     * 
+     * @param descriptor the mongo db repository descriptor
```
troger:
```suggestion
     * @param descriptor the MongoDB repository descriptor
```

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepositoryDescriptor.java
```diff
@@ -54,6 +54,14 @@ public MongoDBRepositoryDescriptor() {
 
     @XNode("sequenceBlockSize")
     public Integer sequenceBlockSize;
+    
+    /** @since 11.1 **/
+    @XNode("childNameUniqueConstraintEnabled")
+    public Boolean childNameUniqueConstraintEnabled;
+
+    public Boolean getChildNameUniqueConstraintEnabled() {
```
troger: Worth adding `@since` here also.

nuxeo-core/nuxeo-core-mongodb/src/test/resources/OSGI-INF/mongodb-disable-renaming-listener-test.xml
RSalem07: `enabled=false` -> 👍 

for the name generally i rename the component to avoid overriding any existing contribution

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
```diff
@@ -0,0 +1,125 @@
...
+import org.nuxeo.runtime.test.runner.FeaturesRunner;
+
+/**
+ * @since 11.1
+ */
+@RunWith(FeaturesRunner.class)
+@Features({ MongoDBFeature.class, CoreFeature.class })
+@Deploy("org.nuxeo.ecm.core.mongodb.test:OSGI-INF/mongodb-disable-renaming-listener-test.xml")
+@Deploy("org.nuxeo.ecm.core.mongodb.test:OSGI-INF/mongodb-enable-unique-indices-contrib.xml")
+public class TestMongoDBIndices {
```
ataillefer: I'd go with a more functional name here, such as `TestMongoDBChildNameUniqueConstraint` since the notion of indice is technical.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
```diff
@@ -0,0 +1,125 @@
...
+    protected static final String DOCUMENT_NAME = "anyFile";
+
+    @Inject
+    protected CoreSession coreSession;
+
+    protected DocumentModel domain;
+
+    @Before
+    public void before() {
+        domain = coreSession.createDocumentModel("/", "Folder", "Folder");
```
ataillefer: `domain` sounds weird as it's just a regular folder

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
```diff
@@ -0,0 +1,125 @@
...
+        domain = coreSession.createDocument(domain);
+        coreSession.save();
+        addPermissions(domain, "james");
+        addPermissions(domain, "bob");
+    }
+
+    @Test
+    public void shouldFailsWhenCreateExistingChildNameDocument() {
```
ataillefer:
```suggestion
    public void shouldFailWhenCreatingExistingChildNameDocument() {
```

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
```diff
@@ -0,0 +1,125 @@
...
+            assertTrue(
+                    cue.getMessage()
+                       .contains(String.format(
+                               "E11000 duplicate key error collection: %s.test index: ecm:parentId_1_ecm:name_1 dup key: { : \"%s\", : \"%s\" }",
+                               DEFAULT_MONGODB_DBNAME, domain.getId(), DOCUMENT_NAME)));
+        }
+    }
+
+    @Test
+    public void shouldFailsWhenMovingDocumentToExistingChildName() {
```
ataillefer:
```suggestion
    public void shouldFailWhenMovingDocumentToExistingChildName() {
```

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
```diff
@@ -0,0 +1,125 @@
...
+            fail("should throw a ConcurrentUpdateException");
+        } catch (ConcurrentUpdateException cue) {
+            assertEquals(SC_CONFLICT, cue.getStatusCode());
+            assertTrue(
+                    cue.getMessage()
+                       .contains(String.format(
+                               "E11000 duplicate key error collection: %s.test index: ecm:parentId_1_ecm:name_1 dup key: { : \"%s\", : \"%s\" }",
+                               DEFAULT_MONGODB_DBNAME, domain.getId(), DOCUMENT_NAME)));
+        }
+    }
```
ataillefer: Just curious: are we sure that creating a document twice with 2 different sessions in a sequential way will lead to the race condition?

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSTransactionState.java
```diff
@@ -1186,7 +1186,7 @@ protected void findDirtyPaths(List<?> value, String path) {
         }
 
         protected void findDirtyPaths(ListDiff value, String path) {
-            String newPath = path + "/*";
+            String newPath = path;
```
ataillefer: Out of curiosity, was there a bug before that wasn't covered by any test? This looks like a more general issue.

nuxeo-core/nuxeo-core-mongodb/src/test/java/org/nuxeo/ecm/core/mongodb/TestMongoDBIndices.java
RSalem07: my idea about this is to group any related stuff to indices. `TestMongoDBChildNameUniqueConstraint` i think is more about a subtask of testing indices

nuxeo-core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSTransactionState.java
RSalem07: yes from point of view it's a general issue. i think before the development of https://jira.nuxeo.com/browse/NXP-26669 we didn't have a test that cover the using of the `relatedText` of document in full text search and when i did NXP-26669 i didn't launch a T&P with a mongo profile. I asked Florent on the platform dev to get lore feedbacks

ataillefer: OK, so it really seems unrelated to NXP-28406.
I think it deserves a separate commit with NXP-26669, or even better a dedicated JIRA issue? (but then it would need a dedicated unit test, so if running a T&P with MongoDB fails once NXP-26669 has been merged, then using this issue in the comment should be OK).

nuxeo-core/nuxeo-core-mongodb/src/test/resources/OSGI-INF/mongodb-disable-renaming-listener-test.xml
troger: AFAIK, you can also remove `class="org.nuxeo.ecm.core.model.DuplicatedNameFixer"`, the listener name is enough to be able to disable it.

nuxeo-core/nuxeo-core-mongodb/src/test/resources/OSGI-INF/mongodb-disable-renaming-listener-test.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+
+<component name="org.nuxeo.mongodb.duplicate.key.test.contrib" version="1.0">
+  <extension target="org.nuxeo.ecm.core.event.EventServiceComponent"
+             point="listener">
+    <listener name="duplicatedNameFixer" enabled="false">
```
ataillefer: you can use an autocloseable tag but that's nitpicking :smile:

PR #3669, opened 2020-01-15
NXP-28521: fix MongoDB upsert retry to avoid duplicate key error

efge:


ataillefer:
Looks good, can this be unit tested for non regression?

efge:
@ataillefer I couldn't find a way to unit test, as the behavior is dependent on MongoDB server version, and isn't systematic (we only had the error in production).

PR #3673, opened 2020-01-15
NXP-28481: Fix for commit failing on timer / record, allow partition …

galvinhsiu:
Fix for NXP-28481. 

Whenever a CommitFailedException happens in the processLoop, we need to catch this exception to prevent the ComputationRunner from exiting, and would call the onPartitionAssigment to realign the consumer context.  This allows retries to happen and prevents the need to rebounce the servers in network fluctuations to the broker.

@mcedica

### DIFF
--- a/nuxeo-runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/computation/log/ComputationRunner.java
+++ b/nuxeo-runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/computation/log/ComputationRunner.java
@@ -31,6 +31,7 @@ import java.util.stream.Collectors;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.kafka.clients.consumer.CommitFailedException;
 import org.nuxeo.lib.stream.computation.Computation;
 import org.nuxeo.lib.stream.computation.ComputationMetadataMapping;
 import org.nuxeo.lib.stream.computation.ComputationPolicy;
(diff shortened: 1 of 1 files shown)


bdelbosc:
Thanks for submitting a pull request.

I see a few problems with this approach:
- Kafka exception is an implementation detail that should not be present at this level.
- CommitFailedException happens because the consumer doesn't respect the poll interval and it is fenced by the controller we should not hide this fact.
- If the error systematic we are going to loop forever, the health check will not report any problem

CommitFailedException should not happen with proper max.poll.interval

In order to improve resiliency when Kafka is temporary not reachable we can simply increase the commit timeout from 1min to 15min.

galvinhsiu:
- Kafka exception is an implementation detail that should not be present at this level.
   Agreed, should use org.nuxeo.lib.stream.log.RebalanceException to be thrown within the commit call of KafkaLogTailer. Will revise the PR later on today.

- CommitFailedException happens because the consumer doesn't respect the poll interval and it is fenced by the controller, we should not hide this fact.
  Will need to try this out later on and reproduce this.  I suspect even if you jack up the poll interval, i believe that the commit failed exception will still happen since the offsets cannot be saved and the thread bails out. Again, this is happening on saveOffset(...) as opposed to the poll interval and the thread needs to retry as opposed to dying.

- If the error is systematic we are going to loop forever, the health check will not report any problem

Depends on how the probe in kafka checks - one option is a configurable fixed # of retries to attempt to recover.  If successful, the counter should be set to 0.  This way the health check can eventually report the error after a reasonable attempt to recover.

I don't see where default.api.timeout.ms is declared at or how it is hooked in - where is this configuration property and I fail to see how this is tied in with the consumer?

bdelbosc:
Hi,

About the CommitFailedException indeed it can be raised for 2 different reasons:
1. When a consumer is fenced by a broker because it doesn't respect the max.poll interval. This is the main reason documented in the Javadoc and traced in the error log.
2. During offset commit when brokers are not resolvable or refuse the connection, in this case, the consumer coordinator prefers to fail using the CommitFailedException (even if the poll interval is respected)
https://github.com/apache/kafka/blob/2.3.0/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L816
At the moment I don't have the solution to handle both problems, rebalance is not good for the main case and I want to avoid to add another retry strategy.

About default.api.timeout.ms, in the case the broker is reachable (name resolvable and accept the connection) but don't answer a TimeoutException after default.api.timeout.ms is raised, this is a consumer property to add in the kafkaConfig contribution.

It is better to discuss this on the NXP tickets:
- NXP-28481 to make sure any failure is reported by the probe (almost done)
- NXP-28524 to improve resiliency when Kafka is not reachable

Regards

bdelbosc:
For now, the strategy is to support network outage using Kafka producer and consumer properties (session, default.api and delivery) and fallback to failure where the thread is stopped and failure is reported by the health probe.
Changes have been done to expose the Kafka properties and improve the health probe.

PR #3677, opened 2020-01-17
NXP-28425: Add MoveToColdStorage operation / REST API

RSalem07:
**IMPORTANT** / **DON'T REVIEW THIS PR**,   but review the main one: https://github.com/nuxeo/nuxeo/pull/3666

### DIFF
--- a/nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
+++ b/nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
@@ -1,0 +1,124 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Salem Aouana
+ */
+
+package org.nuxeo.ecm.core;
+
+import static javax.servlet.http.HttpServletResponse.SC_CONFLICT;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.nuxeo.ecm.core.DummyThumbnailFactory.DUMMY_THUMBNAIL_CONTENT;
+import static org.nuxeo.ecm.core.blob.ColdStorageHelper.COLD_STORAGE_CONTENT_PROPERTY;
+import static org.nuxeo.ecm.core.blob.ColdStorageHelper.FILE_CONTENT_PROPERTY;
+import static org.nuxeo.ecm.core.blob.ColdStorageHelper.moveContentToColdStorage;
+import static org.nuxeo.ecm.core.schema.FacetNames.COLD_STORAGE;
+
+import java.io.IOException;
+import java.io.Serializable;
+
+import javax.inject.Inject;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.nuxeo.ecm.core.api.Blob;
+import org.nuxeo.ecm.core.api.Blobs;
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.NuxeoException;
+import org.nuxeo.ecm.core.test.CoreFeature;
+import org.nuxeo.runtime.test.runner.Deploy;
+import org.nuxeo.runtime.test.runner.Features;
+import org.nuxeo.runtime.test.runner.FeaturesRunner;
+
+/**
+ * @since 11.1
+ */
+@RunWith(FeaturesRunner.class)
+@Features(CoreFeature.class)
+@Deploy("org.nuxeo.ecm.core.test.tests:OSGI-INF/test-coldstorage-contrib.xml")
+public class TestColdStorage {
+
+    protected static final String FILE_CONTENT = "foo";
+
+    @Inject
+    protected CoreSession session;
+
+    @Test
+    public void shouldMoveBlobDocumentToColdStorage() throws IOException {
+        DocumentModel documentModel = createDocument(true);
+
+        // move the blob to cold storage
+        documentModel = moveContentToColdStorage(session, documentModel.getRef());
+        session.save();
+        assertTrue(documentModel.hasFacet(COLD_STORAGE));
+
+        // check if the `file:content` contains the thumbnail blob
+        checkBlobContent(documentModel, FILE_CONTENT_PROPERTY, DUMMY_THUMBNAIL_CONTENT);
+
+        // check if the `coldstorage:coldContent` contains the original file content
+        checkBlobContent(documentModel, COLD_STORAGE_CONTENT_PROPERTY, FILE_CONTENT);
+    }
+
+    @Test
+    public void shouldFailWhenMovingDocumentBlobAlreadyInColdStorage() {
+        DocumentModel documentModel = createDocument(true);
+
+        // move for the first time
+        documentModel = moveContentToColdStorage(session, documentModel.getRef());
+
+        // try to make another move
+        try {
+            moveContentToColdStorage(session, documentModel.getRef());
+            fail("Should fail because the content is already in cold storage");
+        } catch (NuxeoException ne) {
+            assertEquals(SC_CONFLICT, ne.getStatusCode());
+            assertEquals(String.format("The main content for document: %s is already in cold storage.", documentModel),
+                    ne.getMessage());
+        }
+    }
+
+    @Test
+    public void shouldFailWhenMovingToColdStorageDocumentWithoutContent() {
+        DocumentModel documentModel = createDocument(false);
+        try {
+            moveContentToColdStorage(session, documentModel.getRef());
+            fail("Should fail because there is no main content associated with the document");
+        } catch (NuxeoException ne) {
+            assertEquals(SC_NOT_FOUND, ne.getStatusCode());
+            assertEquals(String.format("There is no main content for document: %s.", documentModel), ne.getMessage());
+        }
+    }
+
+    protected DocumentModel createDocument(boolean addBlobContent) {
+        DocumentModel documentModel = session.createDocumentModel("/", "anyFile", "File");
+        if (addBlobContent) {
+            documentModel.setPropertyValue("file:content", (Serializable) Blobs.createBlob(FILE_CONTENT));
+        }
+        return session.createDocument(documentModel);
+    }
+
+    protected void checkBlobContent(DocumentModel documentModel, String xpath, String expectedContent)
+            throws IOException {
+        Blob content = (Blob) documentModel.getPropertyValue(xpath);
+        assertNotNull(content);
+        assertEquals(expectedContent, content.getString());
+    }
+}
--- a/nuxeo-core/nuxeo-core-test/src/test/resources/OSGI-INF/test-coldstorage-contrib.xml
+++ b/nuxeo-core/nuxeo-core-test/src/test/resources/OSGI-INF/test-coldstorage-contrib.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0"?>
+<component name="org.nuxeo.ecm.core.coldstorage.test">
+
+  <extension target="org.nuxeo.ecm.core.api.thumbnail.ThumbnailService"
+             point="thumbnailFactory">
+    <thumbnailFactory name="fakeThumbnailFactory"
+                      factoryClass="org.nuxeo.ecm.core.DummyThumbnailFactory" />
+  </extension>
+
+</component>
--- a/nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
+++ b/nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
@@ -1,0 +1,85 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Salem Aouana
+ */
+
+package org.nuxeo.ecm.core.blob;
+
+import static javax.servlet.http.HttpServletResponse.SC_CONFLICT;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+import static org.nuxeo.ecm.core.schema.FacetNames.COLD_STORAGE;
+
+import java.io.Serializable;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.nuxeo.ecm.core.api.Blob;
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.DocumentRef;
+import org.nuxeo.ecm.core.api.NuxeoException;
+import org.nuxeo.ecm.core.api.thumbnail.ThumbnailService;
+import org.nuxeo.runtime.api.Framework;
+
+/**
+ * Manages the cold storage of the main content of a {@link DocumentModel}.
+ *
+ * @since 11.1
+ */
+public class ColdStorageHelper {
+
+    private static final Logger log = LogManager.getLogger(ColdStorageHelper.class);
+
+    public static final String FILE_CONTENT_PROPERTY = "file:content";
+
+    public static final String COLD_STORAGE_CONTENT_PROPERTY = "coldstorage:coldContent";
+
+    /**
+     * Moves the main content associated with the document of the given {@link DocumentRef} to a cold storage.
+     * 
+     * @return the updated document model if the move succeeds
+     * @throws NuxeoException if there is no main content associated with the given document, or if the main content is
+     *             already in the cold storage
+     */
+    public static DocumentModel moveContentToColdStorage(CoreSession session, DocumentRef documentRef) {
+        DocumentModel documentModel = session.getDocument(documentRef);
+        log.debug("Move to cold storage the main content of document: {}", documentModel);
+
+        Serializable mainContent = documentModel.getPropertyValue(FILE_CONTENT_PROPERTY);
+        if (mainContent == null) {
+            throw new NuxeoException(String.format("There is no main content for document: %s.", documentModel),
+                    SC_NOT_FOUND);
+        }
+
+        if (documentModel.hasFacet(COLD_STORAGE)
+                && documentModel.getPropertyValue(COLD_STORAGE_CONTENT_PROPERTY) != null) {
+            throw new NuxeoException(
+                    String.format("The main content for document: %s is already in cold storage.", documentModel),
+                    SC_CONFLICT);
+        }
+
+        documentModel.addFacet(COLD_STORAGE);
+        documentModel.setPropertyValue(COLD_STORAGE_CONTENT_PROPERTY, mainContent);
+        Blob thumbnail = Framework.getService(ThumbnailService.class).getThumbnail(documentModel, session);
+        documentModel.setPropertyValue(FILE_CONTENT_PROPERTY, (Serializable) thumbnail);
+        return session.saveDocument(documentModel);
+    }
+
+    private ColdStorageHelper() {
+        // no instance allowed
+    }
+}
--- a/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorage.java
+++ b/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorage.java
@@ -1,0 +1,49 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *  
+ * Contributors:
+ *     Salem Aouana
+ */
+
+package org.nuxeo.ecm.automation.core.operations.document;
+
+import static org.nuxeo.ecm.core.blob.ColdStorageHelper.moveContentToColdStorage;
+
+import org.nuxeo.ecm.automation.core.Constants;
+import org.nuxeo.ecm.automation.core.annotations.Context;
+import org.nuxeo.ecm.automation.core.annotations.Operation;
+import org.nuxeo.ecm.automation.core.annotations.OperationMethod;
+import org.nuxeo.ecm.automation.core.collectors.DocumentModelCollector;
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+
+/**
+ * Moves the main content associated with the input {@link DocumentModel} to the cold storage.
+ * 
+ * @since 11.1
+ */
+@Operation(id = MoveToColdStorage.ID, category = Constants.CAT_BLOB, label = "Move to Cold Storage", description = "Move the main document content to the cold storage.")
+public class MoveToColdStorage {
+
+    public static final String ID = "Document.MoveToColdStorage";
+
+    @Context
+    protected CoreSession session;
+
+    @OperationMethod(collector = DocumentModelCollector.class)
+    public DocumentModel run(DocumentModel doc) {
+        return moveContentToColdStorage(session, doc.getRef());
+    }
+}
--- a/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorageTest.java
+++ b/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorageTest.java
@@ -1,0 +1,121 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *  
+ * Contributors:
+ *     Salem Aouana
+ */
+
+package org.nuxeo.ecm.automation.core.operations.document;
+
+import static javax.servlet.http.HttpServletResponse.SC_CONFLICT;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.nuxeo.ecm.core.DummyThumbnailFactory.DUMMY_THUMBNAIL_CONTENT;
+import static org.nuxeo.ecm.core.blob.ColdStorageHelper.COLD_STORAGE_CONTENT_PROPERTY;
+import static org.nuxeo.ecm.core.blob.ColdStorageHelper.FILE_CONTENT_PROPERTY;
+import static org.nuxeo.ecm.core.schema.FacetNames.COLD_STORAGE;
+
+import java.io.IOException;
+import java.io.Serializable;
+
+import javax.inject.Inject;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.nuxeo.ecm.automation.AutomationService;
+import org.nuxeo.ecm.automation.OperationContext;
+import org.nuxeo.ecm.automation.OperationException;
+import org.nuxeo.ecm.core.api.Blob;
+import org.nuxeo.ecm.core.api.Blobs;
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.NuxeoException;
+import org.nuxeo.ecm.core.test.CoreFeature;
+import org.nuxeo.runtime.test.runner.Deploy;
+import org.nuxeo.runtime.test.runner.Features;
+import org.nuxeo.runtime.test.runner.FeaturesRunner;
+
+/**
+ * @since 11.1
+ */
+@RunWith(FeaturesRunner.class)
+@Features(CoreFeature.class)
+@Deploy("org.nuxeo.ecm.automation.core")
+@Deploy("org.nuxeo.ecm.core.test.tests:OSGI-INF/test-coldstorage-contrib.xml")
+public class MoveToColdStorageTest {
+
+    protected static final String FILE_CONTENT = "foo and boo";
+
+    @Inject
+    protected CoreSession session;
+
+    @Inject
+    protected AutomationService automationService;
+
+    @Test
+    public void shouldMoveToColdStorage() throws OperationException, IOException {
+        DocumentModel documentModel = createDocument(true);
+        moveContentToColdStorage(documentModel);
+    }
+
+    @Test
+    public void shouldFailWhenMovingDocumentBlobAlreadyInColdStorage() throws OperationException, IOException {
+        DocumentModel documentModel = createDocument(true);
+        // make a move
+        moveContentToColdStorage(documentModel);
+        try {
+            // try to make a second move
+            moveContentToColdStorage(documentModel);
+            fail("Should fail because the content is already in cold storage");
+        } catch (NuxeoException ne) {
+            assertEquals(SC_CONFLICT, ne.getStatusCode());
+        }
+    }
+
+    @Test
+    public void shouldFailWhenMovingToColdStorageDocumentWithoutContent() throws OperationException, IOException {
+        DocumentModel documentModel = createDocument(false);
+        try {
+            moveContentToColdStorage(documentModel);
+            fail("Should fail because there is no main content associated with the document");
+        } catch (NuxeoException ne) {
+            assertEquals(SC_NOT_FOUND, ne.getStatusCode());
+        }
+    }
+
+    protected void moveContentToColdStorage(DocumentModel documentModel) throws OperationException, IOException {
+        try (OperationContext context = new OperationContext(session)) {
+            context.setInput(documentModel);
+            DocumentModel updatedDocModel = (DocumentModel) automationService.run(context, MoveToColdStorage.ID);
+            Blob fileContent = (Blob) updatedDocModel.getPropertyValue(FILE_CONTENT_PROPERTY);
+            Blob coldStorageContent = (Blob) updatedDocModel.getPropertyValue(COLD_STORAGE_CONTENT_PROPERTY);
+            assertEquals(documentModel.getRef(), updatedDocModel.getRef());
+            assertTrue(updatedDocModel.hasFacet(COLD_STORAGE));
+            assertEquals(DUMMY_THUMBNAIL_CONTENT, fileContent.getString());
+            assertEquals(FILE_CONTENT, coldStorageContent.getString());
+        }
+    }
+
+    protected DocumentModel createDocument(boolean withBlobContent) {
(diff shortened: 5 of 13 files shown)


nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.logging.log4j.Logger;
+import org.nuxeo.ecm.core.api.Blob;
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.DocumentRef;
+import org.nuxeo.ecm.core.api.thumbnail.ThumbnailService;
+import org.nuxeo.runtime.api.Framework;
+
+/**
+ * Manages the cold storage of the {@link Blob} associated to a {@link DocumentModel}.
```
ataillefer:
```suggestion
 * Manages the cold storage of the main content of a {@link DocumentModel}.
```

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @since 11.1
+ */
+public class DocumentBlobColdStorageHelper {
```
ataillefer: Maybe keep it simple with `ColdStorageHelper`?

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private static final Logger log = LogManager.getLogger(DocumentBlobColdStorageHelper.class);
+
+    public static final String FILE_CONTENT_PROPERTY = "file:content";
+
+    public static final String COLD_STORAGE_CONTENT_PROPERTY = "coldstorage:coldContent";
+
+    /**
+     * Moves the blob associated with the document of the given {@link DocumentRef} into a cold storage.
```
ataillefer:
```suggestion
     * Moves the main content associated with the document of the given {@link DocumentRef} to a cold storage.
```

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @param session the core session
+     * @param documentRef the document reference
```
ataillefer: I'd remove these as they don't add any value.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    // FIXME / TODO perhaps return an object that give information, or at least returns the new `file:content` ->
+    // thumbnail
+    public static void moveBlob(CoreSession session, DocumentRef documentRef) {
```
ataillefer: why not `moveToColdStorage`? (Blob is an implementation detail I believe)

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
```diff
@@ -0,0 +1,75 @@
...
+        DocumentModel document = session.getDocument(documentRef);
+        if (document.hasFacet(COLD_STORAGE)) {
```
ataillefer: I think that we should also check if the `coldstorage:coldContent` field is set.
Unless we are totally sure that if a document has this facet it means that its main content is in cold storage, but I wouldn't go for it.
This is linked to https://github.com/nuxeo/nuxeo/pull/3666#discussion_r368465675.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
```diff
@@ -0,0 +1,75 @@
...
+            log.trace("The blob content of document: {} is already on cold storage", document);
```
ataillefer:
```suggestion
            log.trace("The main content of document: {} is already in cold storage", document);
```
And maybe debug? Easier to set without having huge logs, and this case shouldn't happen that often as the UI won't allow to move to cold storage twice.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
```diff
@@ -0,0 +1,75 @@
...
+    public static final String COLD_STORAGE_CONTENT_PROPERTY = "coldstorage:coldContent";
+
+    /**
+     * Moves the blob associated with the document of the given {@link DocumentRef} into a cold storage.
+     *
+     * @param session the core session
+     * @param documentRef the document reference
+     */
+    // FIXME / TODO perhaps return an object that give information, or at least returns the new `file:content` ->
+    // thumbnail
```
ataillefer: Or why not a boolean: true if the main content is not null and was not already in cold storage, false otherwise?

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
```diff
@@ -0,0 +1,75 @@
...
+    public static void moveBlob(CoreSession session, DocumentRef documentRef) {
+        DocumentModel document = session.getDocument(documentRef);
+        if (document.hasFacet(COLD_STORAGE)) {
+            log.trace("The blob content of document: {} is already on cold storage", document);
+            return;
+        }
+        document.addFacet(COLD_STORAGE);
+        Serializable currentContent = document.getPropertyValue(FILE_CONTENT_PROPERTY);
```
ataillefer: I'd go for `mainContent`, that's what we usually use.

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/blob/MoveBlobToColdStorage.java
```diff
@@ -0,0 +1,52 @@
...
+import org.nuxeo.ecm.automation.core.annotations.Context;
+import org.nuxeo.ecm.automation.core.annotations.Operation;
+import org.nuxeo.ecm.automation.core.annotations.OperationMethod;
+import org.nuxeo.ecm.automation.core.collectors.BlobCollector;
+import org.nuxeo.ecm.core.api.Blob;
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+
+/**
+ * Moves the {@link Blob} associated with the {@link DocumentModel} to the cold storage.
```
ataillefer:
```suggestion
 * Moves the main content associated with the input {@link DocumentModel} to cold storage.
```

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/blob/MoveBlobToColdStorage.java
```diff
@@ -0,0 +1,52 @@
...
+@Operation(id = MoveBlobToColdStorage.ID, category = Constants.CAT_BLOB, label = "Move Document Blob To ColdStorage", description = "Moves the blob document content to the cold storage.")
+public class MoveBlobToColdStorage {
+
+    public static final String ID = "Blob.MoveBlobToColdStorage";
+
+    @Context
+    protected CoreSession session;
+
+    @OperationMethod(collector = BlobCollector.class)
+    public Blob run(DocumentModel doc) {
```
ataillefer: I would rather return the updated document, like the `UpdateDocument` operation, rather than the main content which is arbitrary.

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/blob/MoveBlobToColdStorage.java
```diff
@@ -0,0 +1,52 @@
...
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+
+/**
+ * Moves the {@link Blob} associated with the {@link DocumentModel} to the cold storage.
+ * 
+ * @since 11.1
+ */
+@Operation(id = MoveBlobToColdStorage.ID, category = Constants.CAT_BLOB, label = "Move Document Blob To ColdStorage", description = "Moves the blob document content to the cold storage.")
+public class MoveBlobToColdStorage {
```
ataillefer:
```suggestion
public class MoveToColdStorage {
```

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorageTest.java
```diff
@@ -0,0 +1,85 @@
...
+    protected CoreSession session;
+
+    @Inject
+    protected AutomationService automationService;
+
+    @Inject
+    protected TransactionalFeature txFeature;
+
+    @Test
+    public void shouldMoveBlobDocument() throws OperationException, IOException {
```
ataillefer:
```suggestion
    public void shouldMoveToColdStorage() throws OperationException, IOException {
```

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorageTest.java
```diff
@@ -0,0 +1,85 @@
...
+        DocumentModel documentModel = session.createDocumentModel("/", "MyFile", "File");
+        documentModel.setPropertyValue(FILE_CONTENT_PROPERTY, (Serializable) Blobs.createBlob("foo"));
+        documentModel = session.createDocument(documentModel);
+        assertFalse(documentModel.hasFacet(COLD_STORAGE));
+        try (OperationContext context = new OperationContext(session)) {
+            context.setInput(documentModel);
+            Blob fileContent = (Blob) automationService.run(context, MoveBlobToColdStorage.ID, Map.of());
```
ataillefer: If we agree to return the document itself, then here you can add an assertion on the uid of the returned document (must be equal to `documentModel.getId()`).

nuxeo-core/nuxeo-core/src/test/resources/OSGI-INF/test-blob-provider-inmemory.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<component name="org.nuxeo.ecm.core.blob.in-memory" version="1.0.0">
+  <extension target="org.nuxeo.ecm.core.blob.BlobManager" point="configuration">
+    <blobprovider name="test">
+      <class>org.nuxeo.ecm.core.blob.InMemoryBlobProvider</class>
+    </blobprovider>
+  </extension>
+</component>
```
ataillefer: Is that used anywhere? Doesn't seem to be.

nuxeo-core/nuxeo-core-test/src/test/resources/OSGI-INF/test-document-blob-coldstorage-contribs.xml
```diff
@@ -0,0 +1,27 @@
...
+             point="thumbnailFactory">
+    <thumbnailFactory name="fakeThumbnailFactory"
+                      factoryClass="org.nuxeo.ecm.core.FakeThumbnailFactory" />
+  </extension>
+
+  <extension target="org.nuxeo.ecm.core.blob.BlobManager" point="configuration">
+    <blobprovider name="coldstorage">
+      <class>org.nuxeo.ecm.core.blob.InMemoryBlobProvider</class>
+    </blobprovider>
+    <blobprovider name="dummy">
```
ataillefer: Since this contribution already exists in `test-dummy-blob-provider.xml`, maybe we could deploy this contribution in the test to avoid duplication here?

And probably, at some point, we'll have a `ColdStorageFeature` deploying `test-dummy-blob-provider.xml`, `test-document-blob-coldstorage-contribs.xml`, etc.

nuxeo-core/nuxeo-core-test/src/test/resources/OSGI-INF/test-document-blob-coldstorage-contribs.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0"?>
```
ataillefer: Again, to simplify, this file could be named `test-coldstorage-contrib.xml` WDYT?
Also for a quicker retrieval when looking for a resource in the IDE :)

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
RSalem07: i though the same thing, but i told my self perhaps it will not be clear enough as we pass only the document ref as parameter, perhaps people will think that we move the document.

nuxeo-core/nuxeo-core-test/src/test/resources/OSGI-INF/test-document-blob-coldstorage-contribs.xml
```diff
@@ -0,0 +1,27 @@
...
+    <blobprovider name="dummy">
+      <class>org.nuxeo.ecm.core.DummyBlobProvider</class>
+    </blobprovider>
+  </extension>
+  <extension target="org.nuxeo.ecm.core.blob.DocumentBlobManager" point="configuration">
+    <blobdispatcher>
+      <class>org.nuxeo.ecm.core.blob.DefaultBlobDispatcher</class>
+      <property name="blob:xpath=coldContent">coldstorage</property>
+      <property name="blob:xpath=content">dummy</property>
+      <property name="default">test</property>
```
ataillefer: Do we really need to define a default one?

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestDocumentBlobColdStorage.java
```diff
@@ -0,0 +1,124 @@
...
+import org.nuxeo.runtime.test.runner.LoggerLevel;
+import org.nuxeo.runtime.test.runner.TransactionalFeature;
+
+/**
+ * @since 11.1
+ */
+@RunWith(FeaturesRunner.class)
+@Features({ CoreFeature.class, LogFeature.class, LogCaptureFeature.class })
+@Deploy("org.nuxeo.ecm.core.test.tests:OSGI-INF/test-document-blob-coldstorage-contribs.xml")
+public class TestDocumentBlobColdStorage {
```
ataillefer:
```suggestion
public class TestColdStorage {
```

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestDocumentBlobColdStorage.java
```diff
@@ -0,0 +1,124 @@
...
+
+    protected static final String BLOB_FILE_CONTENT = "foo";
```
ataillefer:
```suggestion
    protected static final String FILE_CONTENT = "foo";
```

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestDocumentBlobColdStorage.java
```diff
@@ -0,0 +1,124 @@
...
+    @Test
+    public void shouldMoveBlobDocument() throws IOException {
+        DocumentModel documentModel = session.createDocumentModel("/", "AnyFile", "File");
+        documentModel.setPropertyValue("file:content", (Serializable) Blobs.createBlob(BLOB_FILE_CONTENT));
+        documentModel = session.createDocument(documentModel);
+        assertFalse(documentModel.hasFacet(COLD_STORAGE));
+
+        // move the blob to cold storage
+        moveBlob(session, documentModel.getRef());
+        session.saveDocument(documentModel);
```
ataillefer: not needed I believe

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestDocumentBlobColdStorage.java
```diff
@@ -0,0 +1,124 @@
...
+        transactionalFeature.nextTransaction();
+
+        documentModel = session.getDocument(documentModel.getRef());
```
ataillefer: Florent recently suggested to use `DocumentModel#refresh` instead.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestDocumentBlobColdStorage.java
```diff
@@ -0,0 +1,124 @@
...
+
+        // make another move
+        moveBlob(session, documentModel.getRef());
+        List<String> caughtEvents = logCaptureResult.getCaughtEventMessages();
+        assertEquals(1, caughtEvents.size());
+        assertEquals(String.format("The blob content of document: %s is already on cold storage", documentModel),
+                caughtEvents.get(0));
+    }
+
+    protected void checkBlobContent(DocumentModel documentModel, String xpath, String expectedContent)
```
ataillefer: You can simply use `((Blob) documentModel.getPropertyValue(xpath)).getString()`, directly in the test.

In fact, to enforce the check, maybe:
```
Blob blob content = (Blob) documentModel.getPropertyValue(xpath);
assertNotNull(content);
assertEquals("XYZ", content.getString())
```

But I would still keep these assertions inside the test, for readability.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestDocumentBlobColdStorage.java
```diff
@@ -0,0 +1,124 @@
...
+    @LoggerLevel(klass = DocumentBlobColdStorageHelper.class, level = "TRACE")
+    public void shouldLogMessageWhenMovingDocumentBlobAlreadyOnColdStorage() {
+        DocumentModel documentModel = session.createDocumentModel("/", "MyFile", "File");
+        documentModel.setPropertyValue("file:content", (Serializable) Blobs.createBlob(BLOB_FILE_CONTENT));
+        documentModel = session.createDocument(documentModel);
+        assertFalse(documentModel.hasFacet(COLD_STORAGE));
+
+        // move the first time
+        moveBlob(session, documentModel.getRef());
+        session.saveDocument(documentModel);
```
ataillefer: not needed

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/FakeThumbnailFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import org.nuxeo.ecm.core.api.Blob;
+import org.nuxeo.ecm.core.api.Blobs;
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.thumbnail.ThumbnailFactory;
+
+/**
+ * @since 11.1
+ */
+public class FakeThumbnailFactory implements ThumbnailFactory {
```
ataillefer:
```suggestion
public class DummyThumbnailFactory implements ThumbnailFactory {
```

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobColdStorageHelper.java
ataillefer: OK, so `moveContentToColdStorage`, or `moveMainContentToColdStorage`?
We need to all agree on terminolgy and stick to it.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestDocumentBlobColdStorage.java
RSalem07: for the first test `shouldMoveBlobDocument` i will do it but the second when i replace the get document but the refresh then my test fail on: 
```
  assertEquals(String.format("The main content of document: %s is already in cold storage", documentModel),
                caughtEvents.get(0));
```
because the the expected one had in his `title` the value of `Doc Ref`

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/blob/MoveBlobToColdStorage.java
RSalem07: yep, then i will move the `moveContentToColdStorage ` under the document package instead the `blob` as the input/output a document.  Are you agree ?

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Moves the main content associated with the document of the given {@link DocumentRef} to a cold storage.
+     */
+    public static boolean moveContentToColdStorage(CoreSession session, DocumentRef documentRef) {
+        DocumentModel document = session.getDocument(documentRef);
+        if (document.hasFacet(COLD_STORAGE)) {
+            log.debug("The main content of document: {} is already in cold storage", document);
+            return false;
+        }
```
ataillefer: I think the logic is not correct.
We should have:
```
if (document.hasFacet(COLD_STORAGE) && document.getPropertyValue(COLD_STORAGE_CONTENT_PROPERTY) != null) {
    log.debug("The main content of document: {} is already in cold storage", document);
    return false;

File: nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/blob/MoveBlobToColdStorage.java (hunk 11 above)
Comment (ataillefer) [reply nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/blob/MoveBlobToColdStorage.java]: yes

File: nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorageTest.java [hunk 27]
```diff
@@ -0,0 +1,86 @@
...
+        DocumentModel documentModel = session.createDocumentModel("/", "MyFile", "File");
+        documentModel.setPropertyValue(FILE_CONTENT_PROPERTY, (Serializable) Blobs.createBlob("foo"));
+        documentModel = session.createDocument(documentModel);
+        assertFalse(documentModel.hasFacet(COLD_STORAGE));
+        try (OperationContext context = new OperationContext(session)) {
+            context.setInput(documentModel);
+            DocumentModel updatedDocModel = (DocumentModel) automationService.run(context, MoveToColdStorage.ID,
+                    Map.of());
+            Blob fileContent = (Blob) documentModel.getPropertyValue(FILE_CONTENT_PROPERTY);
+            txFeature.nextTransaction();
```
ataillefer: I think this is not needed since we are in the same transaction (no REST call for instance), in fact, you are retrieving `fileContent` before this call and the assertion that happens after is working...

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorageTest.java
```diff
@@ -0,0 +1,86 @@
...
+            assertEquals(documentModel.getRef(), updatedDocModel.getRef());
+            assertTrue(updatedDocModel.hasFacet(COLD_STORAGE));
+            assertEquals(FAKE_THUMBNAIL_CONTENT, new String(fileContent.getByteArray(), UTF_8));
```
ataillefer: Maybe you could add an assertion on the cold storage content as well?

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
RSalem07: yes,  i was just waiting for https://github.com/nuxeo/nuxeo/pull/3666#discussion_r368565715 before making the changes it's why i didn't include the second part of the check

nuxeo-core/nuxeo-core-test/src/test/resources/OSGI-INF/test-coldstorage-contrib.xml
```diff
@@ -0,0 +1,23 @@
...
+<component name="org.nuxeo.ecm.core.coldstorage.test">
+
+  <extension target="org.nuxeo.ecm.core.api.thumbnail.ThumbnailService"
+             point="thumbnailFactory">
+    <thumbnailFactory name="fakeThumbnailFactory"
+                      factoryClass="org.nuxeo.ecm.core.DummyThumbnailFactory" />
+  </extension>
+
+  <extension target="org.nuxeo.ecm.core.blob.BlobManager" point="configuration">
+    <blobprovider name="dummy">
```
ataillefer: I thought you would keep the `coldstorage` one and remove the `dummy` one, relying `test-dummy-blob-provider.xml` to be deployed in the tests.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
RSalem07: i have my answer i will include it on my next push after your final review ;)

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
```diff
@@ -0,0 +1,124 @@
...
+    @Test
+    public void shouldMoveBlobDocument() throws IOException {
+        DocumentModel documentModel = session.createDocumentModel("/", "AnyFile", "File");
+        documentModel.setPropertyValue("file:content", (Serializable) Blobs.createBlob(FILE_CONTENT));
+        documentModel = session.createDocument(documentModel);
+        assertFalse(documentModel.hasFacet(COLD_STORAGE));
+
+        // move the blob to cold storage
+        moveContentToColdStorage(session, documentModel.getRef());
+        transactionalFeature.nextTransaction();
```
ataillefer: I guess this isn't needed since we're in the same transaction.?

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
```diff
@@ -0,0 +1,124 @@
...
+    @LoggerLevel(klass = ColdStorageHelper.class, level = "DEBUG")
+    public void shouldLogMessageWhenMovingDocumentBlobAlreadyOnColdStorage() {
+        DocumentModel documentModel = session.createDocumentModel("/", "MyFile", "File");
+        documentModel.setPropertyValue("file:content", (Serializable) Blobs.createBlob(FILE_CONTENT));
+        documentModel = session.createDocument(documentModel);
+        assertFalse(documentModel.hasFacet(COLD_STORAGE));
+
+        // move the first time
+        moveContentToColdStorage(session, documentModel.getRef());
+        transactionalFeature.nextTransaction();
```
ataillefer: Not needed I believe.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestDocumentBlobColdStorage.java
ataillefer: OK so if it's not working as expected maybe not use at all here to be consistent between the 2 tests, WDYT?

nuxeo-core/nuxeo-core-test/src/test/resources/OSGI-INF/test-coldstorage-contrib.xml
ataillefer: OK, I see that you're using the "other" blobProvider for coldContent, so you could not define any `blobprovider` at all here I guess, and rely on both contributions:
- `test-blob-provider-inmemory.xml` -> ` <blobprovider name="other">`
- `test-dummy-blob-provider.xml` -> `<blobprovider name="dummy">`

Or I'm missing something.

RSalem07: as we discussed i rollback/revert this part of using a dispatcher i did it to prepare some stuff on the nexts development when we will have the low implementation on S3 Glacier. But  i agree we don't need it right now

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
RSalem07: as above i removed it, mainly our move method will returns the updated document no need for waiting a next transaction :)

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
```diff
@@ -0,0 +1,87 @@
...
+    private static final Logger log = LogManager.getLogger(ColdStorageHelper.class);
+
+    public static final String FILE_CONTENT_PROPERTY = "file:content";
+
+    public static final String COLD_STORAGE_CONTENT_PROPERTY = "coldstorage:coldContent";
+
+    /**
+     * Moves the main content associated with the document of the given {@link DocumentRef} to a cold storage.
+     * 
+     * @return the updated document model if the move has succeed
```
ataillefer: "if the move has succeeded" or (better I think) "if the move succeeds"

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
```diff
@@ -0,0 +1,87 @@
...
+     * @throws NuxeoException if there is no main content associated with the given document, or if the main content is
+     *             already in the cold storage
+     */
+    public static DocumentModel moveContentToColdStorage(CoreSession session, DocumentRef documentRef) {
+        DocumentModel documentModel = session.getDocument(documentRef);
+
+        Serializable mainContent = documentModel.getPropertyValue(FILE_CONTENT_PROPERTY);
+        if (mainContent == null) {
+            throw new NuxeoException(
+                    String.format("There is no main content for the documentModel: %s.", documentModel), SC_NOT_FOUND);
```
ataillefer: maybe just "for document"

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        if (documentModel.hasFacet(COLD_STORAGE)) {
+            Blob coldStorageContent = (Blob) documentModel.getPropertyValue(COLD_STORAGE_CONTENT_PROPERTY);
+            if (coldStorageContent != null) {
+                throw new NuxeoException(
+                        String.format("The main content of documentModel: %s is already in cold storage.",
```
ataillefer: "of document"

RSalem07: yes :) i saw it when i started the new ticket, sorry about it

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
```diff
@@ -0,0 +1,87 @@
...
+
+        Serializable mainContent = documentModel.getPropertyValue(FILE_CONTENT_PROPERTY);
+        if (mainContent == null) {
+            throw new NuxeoException(
+                    String.format("There is no main content for the documentModel: %s.", documentModel), SC_NOT_FOUND);
+        }
+
+        if (documentModel.hasFacet(COLD_STORAGE)) {
+            Blob coldStorageContent = (Blob) documentModel.getPropertyValue(COLD_STORAGE_CONTENT_PROPERTY);
+            if (coldStorageContent != null) {
```
ataillefer: maybe go for
```
if (documentModel.hasFacet(COLD_STORAGE) && documentModel.getPropertyValue(COLD_STORAGE_CONTENT_PROPERTY) != null) {
    ...
}
```
as you're not using `coldStorageContent`, and it will avoid a cast.

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorage.java
```diff
@@ -0,0 +1,49 @@
...
+import org.nuxeo.ecm.automation.core.collectors.DocumentModelCollector;
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+
+/**
+ * Moves the main content associated with the input {@link DocumentModel} to the cold storage.
+ * 
+ * @since 11.1
+ */
+@Operation(id = MoveToColdStorage.ID, category = Constants.CAT_BLOB, label = "Move To ColdStorage", description = "Move the main document content to the cold storage.")
```
ataillefer:
```suggestion
@Operation(id = MoveToColdStorage.ID, category = Constants.CAT_BLOB, label = "Move to Cold Storage", description = "Move the main document content to the cold storage.")
```

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
RSalem07: i agree its what i started to do on the new ticket ;)

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorageTest.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    protected DocumentModel createDocument(boolean withBlobContent) {
+        DocumentModel documentModel = session.createDocumentModel("/", "MyFile", "File");
+        if (withBlobContent) {
+            documentModel.setPropertyValue(FILE_CONTENT_PROPERTY, (Serializable) Blobs.createBlob(FILE_CONTENT));
+        }
+        documentModel = session.createDocument(documentModel);
+
+        return documentModel;
```
ataillefer:
```
return session.createDocument(documentModel);
```

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorageTest.java
```diff
@@ -0,0 +1,127 @@
...
+        } catch (NuxeoException ne) {
+            assertEquals(SC_NOT_FOUND, ne.getStatusCode());
+        }
+    }
+
+    protected void moveContentToColdStorage(DocumentModel documentModel) throws OperationException, IOException {
+        assertFalse(documentModel.hasFacet(COLD_STORAGE));
+        try (OperationContext context = new OperationContext(session)) {
+            context.setInput(documentModel);
+            DocumentModel updatedDocModel = (DocumentModel) automationService.run(context, MoveToColdStorage.ID,
```
ataillefer: I think you can use the 2-parameter method instead of passing an empty Map.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
```diff
@@ -0,0 +1,128 @@
...
+    @Inject
+    protected CoreSession session;
+
+    @Test
+    public void shouldMoveBlobDocumentToColdStorage() throws IOException {
+        DocumentModel documentModel = createDocument(true);
+
+        // move the blob to cold storage
+        documentModel = moveContentToColdStorage(session, documentModel.getRef());
+        documentModel = session.getDocument(documentModel.getRef());
```
ataillefer: useless

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
```diff
@@ -0,0 +1,128 @@
...
+        checkBlobContent(documentModel, COLD_STORAGE_CONTENT_PROPERTY, FILE_CONTENT);
+    }
+
+    @Test
+    public void shouldFailWhenMovingDocumentBlobAlreadyInColdStorage() {
+        DocumentModel documentModel = createDocument(true);
+
+        // move for the first time
+        documentModel = moveContentToColdStorage(session, documentModel.getRef());
+        documentModel = session.getDocument(documentModel.getRef());
```
ataillefer: useless

nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/operations/document/MoveToColdStorageTest.java
```diff
@@ -0,0 +1,127 @@
...
+        try {
+            moveContentToColdStorage(documentModel);
+            fail("Should fail because there is no main content associated with the document");
+        } catch (NuxeoException ne) {
+            assertEquals(SC_NOT_FOUND, ne.getStatusCode());
+        }
+    }
+
+    protected void moveContentToColdStorage(DocumentModel documentModel) throws OperationException, IOException {
+        assertFalse(documentModel.hasFacet(COLD_STORAGE));
```
ataillefer: I think you should remove this, because `shouldFailWhenMovingToColdStorageDocumentWithoutContent` will fail not for the expected reason.

RSalem07:
Closed in favor of #3666

PR #3693, opened 2020-01-23
NXP-28502: Fix permanent deletion of folder containing a large number…

ThierryCasanova:
Need also to make a test with large number of documents to purge

### DIFF
--- a/nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/trash/TrashService.java
+++ b/nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/trash/TrashService.java
@@ -167,7 +167,9 @@ public interface TrashService {
      * @param parent The parent document of trash document.
      * @return All documents in the trash of the current document.
      * @since 7.1
+     * @deprecated since 11.1, not scalable
      */
+    @Deprecated(since = "11.1")
     DocumentModelList getDocuments(DocumentModel parent);
 
     /**
--- a/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/test/CoreOperationsTest.java
+++ b/nuxeo-features/nuxeo-automation/nuxeo-automation-core/src/test/java/org/nuxeo/ecm/automation/core/test/CoreOperationsTest.java
@@ -30,8 +30,6 @@ import static org.junit.Assert.fail;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
 import javax.inject.Inject;
@@ -80,11 +78,11 @@ import org.nuxeo.ecm.core.api.Blob;
 import org.nuxeo.ecm.core.api.Blobs;
 import org.nuxeo.ecm.core.api.CoreSession;
 import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.DocumentModelList;
 import org.nuxeo.ecm.core.api.DocumentRef;
 import org.nuxeo.ecm.core.api.PathRef;
-import org.nuxeo.ecm.core.api.trash.TrashService;
 import org.nuxeo.ecm.core.test.CoreFeature;
-import org.nuxeo.runtime.api.Framework;
+import org.nuxeo.ecm.core.trash.AbstractTrashService;
 import org.nuxeo.runtime.test.runner.Deploy;
 import org.nuxeo.runtime.test.runner.Features;
 import org.nuxeo.runtime.test.runner.FeaturesRunner;
@@ -523,7 +521,7 @@ public class CoreOperationsTest {
     @Test
     public void testTrashUntrash() throws IOException, OperationException {
         DocumentModel parent = session.getDocument(src.getParentRef());
-        assertEquals(0, Framework.getService(TrashService.class).getDocuments(parent).size());
+        assertEquals(0, getTrashedDocuments(parent).size());
 
         try (OperationContext ctx = new OperationContext(session)) {
             ctx.setInput(src);
@@ -532,8 +530,7 @@ public class CoreOperationsTest {
             chain.add(FetchContextDocument.ID);
             chain.add(TrashDocument.ID);
             src = (DocumentModel) service.run(ctx, chain);
-            Framework.getService(TrashService.class).getDocuments(parent);
-            assertEquals(1, Framework.getService(TrashService.class).getDocuments(parent).size());
+            assertEquals(1, getTrashedDocuments(parent).size());
         }
         try (OperationContext ctx = new OperationContext(session)) {
             ctx.setInput(src);
@@ -541,7 +538,7 @@ public class CoreOperationsTest {
             chain.add(UntrashDocument.ID);
             service.run(ctx, chain);
 
-            assertEquals(0, Framework.getService(TrashService.class).getDocuments(parent).size());
+            assertEquals(0, getTrashedDocuments(parent).size());
         }
     }
 
@@ -551,7 +548,7 @@ public class CoreOperationsTest {
     @Test
     public void testEmptyTrash() throws IOException, OperationException {
         DocumentModel parent = session.getDocument(src.getParentRef());
-        assertEquals(0, Framework.getService(TrashService.class).getDocuments(parent).size());
+        assertEquals(0, getTrashedDocuments(parent).size());
 
         try (OperationContext ctx = new OperationContext(session)) {
             ctx.setInput(src);
@@ -560,8 +557,7 @@ public class CoreOperationsTest {
             chain.add(FetchContextDocument.ID);
             chain.add(TrashDocument.ID);
             src = (DocumentModel) service.run(ctx, chain);
-            Framework.getService(TrashService.class).getDocuments(parent);
-            assertEquals(1, Framework.getService(TrashService.class).getDocuments(parent).size());
+            assertEquals(1, getTrashedDocuments(parent).size());
         }
 
         try (OperationContext ctx = new OperationContext(session)) {
@@ -569,7 +565,7 @@ public class CoreOperationsTest {
             chain.add(EmptyTrash.ID).set("parent", parent);
             service.run(ctx, chain);
 
-            assertEquals(0, Framework.getService(TrashService.class).getDocuments(parent).size());
+            assertEquals(0, getTrashedDocuments(parent).size());
             assertEquals(1, session.getChildren(parent.getRef()).size());
         }
     }
@@ -805,4 +801,8 @@ public class CoreOperationsTest {
         }
     }
 
+    protected DocumentModelList getTrashedDocuments(DocumentModel parent) {
+        CoreSession session = parent.getCoreSession();
+        return session.query(String.format(AbstractTrashService.TRASHED_QUERY, parent.getId()));
+    }
 }


nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/trash/TrashService.java
```diff
@@ -84,6 +84,17 @@ default boolean canPurgeOrUntrash(DocumentModel doc, NuxeoPrincipal principal) {
         return canPurgeOrUntrash(Collections.singletonList(doc), principal);
     }
 
+    /**
+     * Check if trashed child documents from a given parent document are "purgeable/untrashable".
+     * Documents need to be in the trash for this to be true, in addition to the standard permission checks.
```
RSalem07: Checks if child document for a given document are  "purgeable/untrashable".

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/trash/TrashService.java
```diff
@@ -84,6 +84,17 @@ default boolean canPurgeOrUntrash(DocumentModel doc, NuxeoPrincipal principal) {
...
+     * @param parent the parent document.
+     * @param principal the user.
```
RSalem07: don't add dot at the end of `@param`
for the return tag you can use:
` @return {@code true} if .......`
`<code>true</code>` -> `{@code}`

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/trash/TrashService.java
```diff
@@ -167,7 +178,9 @@ default void untrashDocument(DocumentModel doc) {
      * @param parent The parent document of trash document.
      * @return All documents in the trash of the current document.
      * @since 7.1
+     * @deprecated since 11.1 not scalable (prefer to use directly {@link CoreSession#query(String)} with a limit
      */
```
RSalem07: here you don't give enough informations, which method will replace the deprecated once, ` CoreSession#query(String)` with a limit its not enough because perhaps the called want the total document without using limit. Perhaps just `@deprecated since 11.1, not scalable or not used anymore` or at least you can propose to replace it by `BAF`

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/trash/TrashService.java
```diff
@@ -167,7 +178,9 @@ default void untrashDocument(DocumentModel doc) {
...
+    @Deprecated(since = "11.1")
```
RSalem07: you can replace it `@Deprecated(since = "11.1", forRemoval = true)`

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/scroll/RepositoryScroll.java
```diff
@@ -41,6 +44,8 @@
  */
 public class RepositoryScroll implements Scroll {
 
+    public static final String SELECT_DOCUMENTS_IN = "SELECT * FROM Document, Relation WHERE ecm:uuid IN ('%s')";
+
```
RSalem07: do we need `Document` and `Relation` ? as there is no joins, i am not sure how this query will be interpreted.
furthermore  can you please give a name field with more context `GET_DOCUMENTS_BY_UUID_QUERY` or something else

RSalem07: In other hand we don't given a specific methods on this Repo scroller, just call it from outside using the `init` method ... create ... see `org.nuxeo.ecm.core.scroll.TestRepositoryScroll` for examples

nuxeo-core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/scroll/RepositoryScroll.java
```diff
@@ -117,6 +122,19 @@ public void close() {
...
+     * @param session the Core session
+     * @param documentIds the document Ids' list.
+     * @return the corresponding DocumentModelList.
+     */
+    public static DocumentModelList loadDocuments(CoreSession session, List<String> documentIds) {
+        if (documentIds == null || documentIds.isEmpty()) {
+            return new DocumentModelListImpl(0);
+        }
+        return session.query(String.format(SELECT_DOCUMENTS_IN, String.join("', '", documentIds)));
+    }
```
RSalem07: see my comment on https://github.com/nuxeo/nuxeo/pull/3693/commits/48c62d0e6e904561781b271ae17ba1c68f67712e#r370538308
furthermore i think you cas use `StaticScroll` as you have the list of ids .... you can ask Benoit for more context

and for more context, you can rewrite this method as below:
````
  String inClause = documentIds.stream().collect(Collectors.joining(", ", "'", "'"));
        String query = String.format(SELECT_DOCUMENTS_IN, NXQL.ECM_UUID, inClause);
        return session.query(query);
```
where ` SELECT_DOCUMENTS_IN = "SELECT * FROM Document, Relation WHERE %s IN (%s)";`

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/trash/test/AbstractTestTrashService.java
```diff
@@ -417,4 +418,18 @@ public void testFollowTransitionBackwardCompatibility() {
...
+        createDocuments();
+        trashService.trashDocument(doc1);
+        transactionalFeature.nextTransaction();
+
+        assertTrue(trashService.getCanEmptyTrash(fold, principal));
+
+        trashService.trashDocument(fold);
+        transactionalFeature.nextTransaction();
+
+        assertTrue(trashService.getCanEmptyTrash(fold, principal));
```
RSalem07: what you want to test here:

- that you can pure the folder children 
- or purge folder and children

i ask this question depending on you answer the java doc should be update / reworked on `getCanEmptyTrash` method ...

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/AbstractTrashService.java
```diff
@@ -112,19 +117,60 @@ public boolean canPurgeOrUntrash(List<DocumentModel> docs, NuxeoPrincipal princi
...
+                                                     .build();
+        ScrollService service = Framework.getService(ScrollService.class);
+        boolean canPurge = true;
+        try (Scroll scroll = service.scroll(request)) {
+            if (!scroll.hasNext()) {
+                return false;
+            }
+            while (scroll.hasNext() && canPurge) {
+                DocumentModelList documents = RepositoryScroll.loadDocuments(parent.getCoreSession(), scroll.next());
+                canPurge = canPurgeOrUntrash(documents, principal);
```
RSalem07: here you don't to call `RespositoryScroll#loadDocuments`  you don't scroll a second time ? is it right 

just make query on the `CoreSession#query` or extract a method in the `TrashService` but we should care about what `scroll.next()` returns i means how the size of the documents list will be, because in some SQL DB there is some limitation when you make `IN` for example Oracle don't accepts more then `1000` with `IN` clause

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/AbstractTrashService.java
```diff
@@ -112,19 +117,60 @@ public boolean canPurgeOrUntrash(List<DocumentModel> docs, NuxeoPrincipal princi
...
+            }
+        }
+        return canPurge;
+    }
+
+    /**
+     * Return the {@link TrashInfo} of a document list with potentially some documents not ok for 'purge' or unTrash
+     * (forbidden, locked ...)
```
RSalem07: Return -> replace it by `Gets`  and the javadoc is not required for `protected` method

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/AbstractTrashService.java
```diff
@@ -112,19 +117,60 @@ public boolean canPurgeOrUntrash(List<DocumentModel> docs, NuxeoPrincipal princi
...
+     * @param docs the documents List to trash
+     * @param principal the user principal
+     * @param checkProxies <code>true</code> to check proxies.
+     * @param checkDeleted <code>true</code> to check if doc is trashed.
+     * @param exitOnFirstDelta <code>true</code> to avoid looping in the whole documents List and exist on first
+     *            document with TrashInfo "forbidden", "proxies" or "locked".
+     * @return the TrashInfo for the documents List.
+     */
+    protected TrashInfo getInfo(List<DocumentModel> docs, NuxeoPrincipal principal, boolean checkProxies,
+            boolean checkDeleted, boolean exitOnFirstDelta) {
```
RSalem07: we dont change the protected/public methods signature for backward compatibility just create a new once and call it from the exiting once

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/AbstractTrashService.java
```diff
@@ -112,19 +117,60 @@ public boolean canPurgeOrUntrash(List<DocumentModel> docs, NuxeoPrincipal princi
...
         TrashInfo info = new TrashInfo();
         info.docs = new ArrayList<>(docs.size());
         if (docs.isEmpty()) {
             return info;
         }
         CoreSession session = docs.get(0).getCoreSession();
         for (DocumentModel doc : docs) {
+            if (exitOnFirstDelta && (info.forbidden > 0 || info.locked >0 || info.proxies >0)){
+                return info;
```
RSalem07: need to be formatted, extra space needed on the

RSalem07: With this premature exit you change/break the exiting behaviour of this method as before you changed it, it main purpose was to give more informations about the trash state, and now you exit the method if your check is true  i think we should rework it depending on the need

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/AbstractTrashService.java
```diff
@@ -299,6 +345,7 @@ protected void notifyEvent(CoreSession session, String eventId, DocumentModel do
         eventService.fireEvent(event);
     }
 
+    @Deprecated(since = "11.1")
```
RSalem07: no need for deprecated on `Override` method as you did it on the interface one

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/AbstractTrashService.java
ThierryCasanova: default scroll size is 50

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/trash/AbstractTrashService.java
NourNuxeo: The behavior is only changed if the exitOnFirstDelta is true, which by default is false in the new internal call so the behavior is still the same:

https://github.com/nuxeo/nuxeo/pull/3693/commits/d61fa8ff5b9e46aa598002900f0e794c635bc76b#diff-7c7663426b7649bd97453eaa63571db0R122

Not advocating for a way or another though

ThierryCasanova: finally roll-backed to reduice commit impact to jsf bean only

RSalem07: @NourNuxeo about https://github.com/nuxeo/nuxeo/pull/3693#discussion_r372988424 here the method is changed on its behaviour, before adding parameters the method will take each parameter as a filter for the process but the new once will exit the loop which is different.

as discussed with thierry this change is not required and it's why he rollback it

NourNuxeo: Of course it does extra stuff, that's why he changes it. But it still behaves the same way in the places it was called before.

Yes I followed up with the discussion about the rollback 👍

RSalem07: > Of course it does extra stuff, that's why he changes it. But it still behaves the same way in the places it was called before.

i know why he changed it,  but i don't agree with this modification as the new method now can do more then just filter and any one can called it without knowing exactly what means the new parameter

NourNuxeo: The method was protected and the param was named. But I like method who do only one thing better than the multi behavioral ones for sure.

NourNuxeo:
Did I miss something or is this commit only just a refactoring ? Then the message would be missleading.
I see:
- deprecation in the main class
- custom method in the test class that does the same query as the trash service in a more compact writing.

PR #3613, opened 2019-12-16
Task nxp 28008 paginate tasks

guirenard:

PR #3711, opened 2020-01-30
Task nxp 28008 paginate tasks 10.10 (Backport)

nmpcunha:
🔵 TestAndPush: https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-ncunha-10.10/17/

master PR: https://github.com/nuxeo/nuxeo/pull/3613

### DIFF
--- a/addons/nuxeo-platform-document-routing/nuxeo-routing-api/src/main/java/org/nuxeo/ecm/platform/routing/api/DocumentRoutingService.java
+++ b/addons/nuxeo-platform-document-routing/nuxeo-routing-api/src/main/java/org/nuxeo/ecm/platform/routing/api/DocumentRoutingService.java
@@ -588,7 +588,11 @@ public interface DocumentRoutingService {
      * @param session
      * @return
      * @since 7.2
+     * @deprecated since 11.1 this method does not scale as it does not paginate results. Use
+     *             {@link org.nuxeo.ecm.platform.routing.core.provider.RoutingTaskPageProvider#getCurrentPage()}
+     *             instead.
      */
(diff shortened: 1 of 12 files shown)


addons/nuxeo-platform-document-routing/nuxeo-routing-api/src/main/java/org/nuxeo/ecm/platform/routing/api/DocumentRoutingService.java
```diff
@@ -588,6 +588,8 @@ void grantPermissionToTaskDelegatedActors(CoreSession session, String permission
      * @param session
      * @return
      * @since 7.2
+     * since 11.1 this method does not scale as it does not paginate results.
```
kevinleturc: There's missing the `@deprecated` in both javadoc and method, if you don't want it in the backport you can remove this kind of line. Note that we're ok to backport deprecation such as them.

nmpcunha: We tried to avoid that in order to have warnings on IDE since it is not deprecated for this version.
Does this sound a good approach?

nmpcunha: So, as per our conversation, we kept the cherry-pick commit.
In order to be clear the difference on 10.10, a separate commit was introduced.

PR #3725, opened 2020-02-05
NXP-28622: fix delete namespace still in terminating state

doulba:
For this PR, we don't to run QA job, it's only for CI purpose.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -389,7 +389,11 @@ pipeline {
           junit testResults: '**/target/surefire-reports/*.xml'
           container('maven') {
             // clean up the redis namespace
-            sh "kubectl delete namespace ${NAMESPACE_REDIS} --ignore-not-found=true"
+            // workaround for Cannot delete namespace still in Terminating #NXP-28622
+            sh """
+              kubectl get namespace ${NAMESPACE_REDIS} -o json | sed -e s/\\"kubernetes\\"//g  | kubectl replace --raw /api/v1/namespaces/${NAMESPACE_REDIS}/finalize -f -
+              kubectl delete namespace ${NAMESPACE_REDIS} --ignore-not-found=true
+            """
           }
         }
         success {


Jenkinsfile
```diff
@@ -388,8 +388,18 @@ pipeline {
...
           container('maven') {
-            // clean up the redis namespace
-            sh "kubectl delete namespace ${NAMESPACE_REDIS} --ignore-not-found=true"
+            script {
+              try {
+                 // clean up the redis namespace
+                // workaround for Cannot delete namespace still in Terminating #NXP-3228
+                sh """
+                  kubectl get namespace ${NAMESPACE_REDIS} -o json | sed -e s/\\"kubernetes\\"//g  | kubectl replace --raw /api/v1/namespaces/${NAMESPACE_REDIS}/finalize -f -
+                  kubectl delete namespace ${NAMESPACE_REDIS} --ignore-not-found=true --cascade=true --force=1 --grace-period=0
```
ataillefer: I think we can remove `--cascade=true --force=1 --grace-period=0` since a plain deletion is expected to work here.

Jenkinsfile
```diff
@@ -388,8 +388,18 @@ pipeline {
         always {
           junit testResults: '**/target/surefire-reports/*.xml'
           container('maven') {
-            // clean up the redis namespace
-            sh "kubectl delete namespace ${NAMESPACE_REDIS} --ignore-not-found=true"
+            script {
+              try {
+                 // clean up the redis namespace
```
ataillefer: Can you please indent with 2 spaces instead of 3

Jenkinsfile
```diff
@@ -388,8 +388,18 @@ pipeline {
...
+                // workaround for Cannot delete namespace still in Terminating #NXP-3228
+                sh """
+                  kubectl get namespace ${NAMESPACE_REDIS} -o json | sed -e s/\\"kubernetes\\"//g  | kubectl replace --raw /api/v1/namespaces/${NAMESPACE_REDIS}/finalize -f -
+                  kubectl delete namespace ${NAMESPACE_REDIS} --ignore-not-found=true --cascade=true --force=1 --grace-period=0
+                """
+              } catch (err) {
```
ataillefer: I'm not in favor of catching any error here: if there is an error in the workaround or the deletion, it is not expected, so we want to be aware of it even if it makes the build fail.

This is how we detected the namespace deletion problem in the first place, because the jobs were red.

Jenkinsfile
doulba: thanks,
 we can delete this options as you said

Jenkinsfile
doulba: done

Jenkinsfile
doulba: we can remove the try catch and monitor the namespace  deletion 
if many build failed due to this error perhaps we should reactivates the try catch.

Jenkinsfile
```diff
@@ -388,8 +388,14 @@ pipeline {
         always {
           junit testResults: '**/target/surefire-reports/*.xml'
           container('maven') {
-            // clean up the redis namespace
-            sh "kubectl delete namespace ${NAMESPACE_REDIS} --ignore-not-found=true"
+            script {
+              // clean up the redis namespace
+              // workaround for Cannot delete namespace still in Terminating #NXP-3228
```
troger: Wrong NXP I think :)

Jenkinsfile
```diff
@@ -388,8 +388,14 @@ pipeline {
         always {
           junit testResults: '**/target/surefire-reports/*.xml'
           container('maven') {
-            // clean up the redis namespace
-            sh "kubectl delete namespace ${NAMESPACE_REDIS} --ignore-not-found=true"
+            script {
```
kevinleturc: Not sure we need `script` here as we don't do script?

doulba: yes, 
i have tested many solutions , that’s way we still have scripts  section.
i will  remove it

Jenkinsfile
ataillefer: yep, I think the workaround will do its job fine

PR #3729, opened 2020-02-07
NXP-28635: Use save param

michaelva:


troger:
Worth adding a simple unit test.

efge:
If this is merged for 10.10 then this should be merged for master too. I agree that we would need a unit test but this can be done in a separate step now. We should have parity in master for all code that is in a maintenance branch.

michaelva:
I would gladly merge but need one more review/approval to do so :)

michaelva:
@efge I don't believe it changes the behavior. The save param default value is true so unless it was explicitly set to false in chains/scripts that use the operation, the behavior remains unchanged. If the param was set to false, the document was saved anyway which wasn't the expected result.

efge:
@michaelva please also resolve the corresponding NXP ticket as well once this is merged

PR #3730, opened 2020-02-07
NXP-28635: Use save param

michaelva:


ataillefer:
@michaelva the PR was approved a bit quickly, sorry about that, though the code looks right, a T&P job should have been launched, especially for a maintenance branch.

PR #3733, opened 2020-02-07
feature-NXP-28579-unit-tests-postgresql-mongodb-parallel

ataillefer:


Jenkinsfile
```diff
@@ -130,7 +130,8 @@ def buildUnitTestStage(env) {
             sh 'envsubst < ci/helm/nuxeo-test-base-values.yaml > nuxeo-test-base-values.yaml'
             def testValues = '--set-file=nuxeo-test-base-values.yaml'
             if (!isDev) {
-              testValues += " --set-file=ci/helm/nuxeo-test-${env}-values.yaml"
+              sh "envsubst < ci/helm/nuxeo-test-${env}-values.yaml > nuxeo-test-${env}-values.yaml"
```
troger: What if the file does not exist? Dunno if we will have the case...

Jenkinsfile
```diff
@@ -100,19 +105,111 @@ void skaffoldBuildAll() {
...
+              sh "cp ci/mvn/nuxeo-test-${env}.properties ${HOME}/nuxeo-test-${env}.properties"
+            }
+            // run unit tests:
+            // - in nuxeo-core and dependent projects only (nuxeo-common and nuxeo-runtime are run in dedicated stages)
+            // - for the given environment (see the customEnvironment profile in pom.xml):
+            //   - in an alternative build directory
+            //   - loading some backend-specific system properties
+            def testCore = env == 'mongodb' ? 'mongodb' : 'vcs'
+            sh """
+              mvn -B -nsu -pl nuxeo-core -amd \
```
troger: Shouldn't we use something like:
```
mvn -B -nsu -rm nuxeo-core \
```
?

Jenkinsfile
ataillefer: Shouldn't have the case, we assume we're maintaining the values files according to the env list/

ataillefer:
T&P OK: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-ataillefer/202/

PR #3741, opened 2020-02-11
NXP-28417: S3 Glacier restore and blob status (10.10)

efge:
No tests due to limited ways doing Glacier tests

### DIFF
--- a/addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
+++ b/addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
@@ -18,10 +18,14 @@
  */
 package org.nuxeo.ecm.blob.s3;
 
+import static java.lang.Boolean.TRUE;
+import static org.nuxeo.ecm.blob.s3.S3BlobStore.VER_SEP;
+
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.time.Instant;
 import java.util.Date;
 import java.util.Map;
 
@@ -34,6 +38,7 @@ import org.apache.logging.log4j.Logger;
 import org.nuxeo.common.utils.RFC2231;
 import org.nuxeo.ecm.core.api.Blob;
 import org.nuxeo.ecm.core.blob.BlobManager;
+import org.nuxeo.ecm.core.blob.BlobStatus;
 import org.nuxeo.ecm.core.blob.BlobStore;
 import org.nuxeo.ecm.core.blob.BlobStoreBlobProvider;
 import org.nuxeo.ecm.core.blob.CachingBlobStore;
@@ -42,10 +47,14 @@ import org.nuxeo.ecm.core.blob.ManagedBlob;
 import org.nuxeo.ecm.core.blob.TransactionalBlobStore;
 import org.nuxeo.ecm.core.io.download.DownloadHelper;
 
+import com.amazonaws.AmazonServiceException;
 import com.amazonaws.HttpMethod;
 import com.amazonaws.services.cloudfront.CloudFrontUrlSigner;
 import com.amazonaws.services.cloudfront.util.SignerUtils.Protocol;
 import com.amazonaws.services.s3.model.GeneratePresignedUrlRequest;
+import com.amazonaws.services.s3.model.GetObjectMetadataRequest;
+import com.amazonaws.services.s3.model.ObjectMetadata;
+import com.amazonaws.services.s3.model.StorageClass;
 
 /**
  * Blob provider that stores files in S3.
@@ -177,6 +186,49 @@ public class S3BlobProvider extends BlobStoreBlobProvider {
 
     protected String getContentDispositionHeader(Blob blob) {
         return RFC2231.encodeContentDisposition(blob.getFilename(), false, null);
+    }
+
+    @Override
+    public BlobStatus getStatus(ManagedBlob blob) throws IOException {
+        String key = stripBlobKeyPrefix(blob.getKey());
+        String objectKey;
+        String versionId;
+        int seppos = key.indexOf(VER_SEP);
+        if (seppos < 0) {
+            objectKey = key;
+            versionId = null;
+        } else {
+            objectKey = key.substring(0, seppos);
+            versionId = key.substring(seppos + 1);
+        }
+        String bucketKey = config.bucketPrefix + objectKey;
+        GetObjectMetadataRequest request = new GetObjectMetadataRequest(config.bucketName, bucketKey, versionId);
+        ObjectMetadata metadata;
+        try {
+            metadata = config.amazonS3.getObjectMetadata(request);
+        } catch (AmazonServiceException e) {
+            if (S3BlobStore.isMissingKey(e)) {
+                // don't crash for a missing blob, even though it means the storage is corrupted
+                log.error("Failed to get information on blob: {}", key, e);
+                return new BlobStatus().withDownloadable(false);
+            }
+            throw new IOException(e);
+        }
+        // storage class is null for STANDARD
+        String storageClass = metadata.getStorageClass();
+        if (StorageClass.Standard.toString().equals(storageClass)) {
+            storageClass = null;
+        }
+        // x-amz-restore absent
+        // x-amz-restore: ongoing-request="true"
+        // x-amz-restore: ongoing-request="false", expiry-date="Fri, 23 Dec 2012 00:00:00 GMT"
+        Boolean ongoingRestore = metadata.getOngoingRestore();
+        boolean downloadable = !TRUE.equals(ongoingRestore);
+        Date date = metadata.getRestoreExpirationTime();
+        Instant downloadableUntil = date == null ? null : date.toInstant();
+        return new BlobStatus().withStorageClass(storageClass)
+                               .withDownloadable(downloadable)
+                               .withDownloadableUntil(downloadableUntil);
     }
 
 }
--- a/addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
+++ b/addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
@@ -71,6 +72,7 @@ import com.amazonaws.services.s3.model.ObjectLockRetention;
 import com.amazonaws.services.s3.model.ObjectLockRetentionMode;
 import com.amazonaws.services.s3.model.ObjectMetadata;
 import com.amazonaws.services.s3.model.PutObjectRequest;
+import com.amazonaws.services.s3.model.RestoreObjectRequest;
 import com.amazonaws.services.s3.model.S3ObjectSummary;
 import com.amazonaws.services.s3.model.S3VersionSummary;
 import com.amazonaws.services.s3.model.SSEAwsKeyManagementParams;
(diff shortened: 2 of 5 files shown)


addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
```diff
@@ -179,4 +189,45 @@ protected String getContentDispositionHeader(Blob blob) {
         return RFC2231.encodeContentDisposition(blob.getFilename(), false, null);
     }
 
+    @Override
+    public BlobStatus getStatus(ManagedBlob blob) throws IOException {
```
RSalem07: throws `IOException` is not needed

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
```diff
@@ -179,4 +189,45 @@ protected String getContentDispositionHeader(Blob blob) {
...
+        if (seppos < 0) {
+            objectKey = key;
+            versionId = null;
+        } else {
+            objectKey = key.substring(0, seppos);
+            versionId = key.substring(seppos + 1);
+        }
+        String bucketKey = config.bucketPrefix + objectKey;
+        GetObjectMetadataRequest request = new GetObjectMetadataRequest(config.bucketName, bucketKey);
+        request.setVersionId(versionId);
```
RSalem07: you can use the constructor with three args (` public GetObjectMetadataRequest(String bucketName, String key, String versionId) `

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
```diff
@@ -179,4 +189,45 @@ protected String getContentDispositionHeader(Blob blob) {
...
+    @Override
+    public BlobStatus getStatus(ManagedBlob blob) throws IOException {
+        String key = stripBlobKeyPrefix(blob.getKey());
+        String objectKey;
+        String versionId;
+        int seppos = key.indexOf(VER_SEP);
+        if (seppos < 0) {
+            objectKey = key;
+            versionId = null;
+        } else {
```
RSalem07: wdyt if we replace it by:
```
        String objectKey = key;
        String versionId = null;
        int seppos = key.indexOf(VER_SEP);
        if(seppos >= 0) {
            objectKey = key.substring(0, seppos);
            versionId = key.substring(seppos + 1);
        }
```

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
```diff
@@ -179,4 +189,45 @@ protected String getContentDispositionHeader(Blob blob) {
...
+            if (S3BlobStore.isMissingKey(e)) {
+                // don't crash for a missing blob, even though it means the storage is corrupted
+                log.error("Failed to get information on blob: " + key, e);
+                return new BlobStatus();
+            }
+            throw e;
+        }
+        // storage class is null for STANDARD
+        String storageClass = defaultIfEmpty(metadata.getStorageClass(), StorageClass.Standard.toString());
+        // x-amz-restore absent
```
RSalem07: i checkout you branch and i get an error compilation here, pehaps i miss something, but your static import 
`import static org.apache.commons.lang.StringUtils.defaultIfEmpty;` doesn't exist i replace it by `import static org.apache.commons.lang3.StringUtils.defaultIfEmpty;` i think we use apache common 3

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
```diff
@@ -179,4 +189,45 @@ protected String getContentDispositionHeader(Blob blob) {
...
+        GetObjectMetadataRequest request = new GetObjectMetadataRequest(config.bucketName, bucketKey);
+        request.setVersionId(versionId);
+        ObjectMetadata metadata;
+        try {
+            metadata = config.amazonS3.getObjectMetadata(request);
+        } catch (AmazonServiceException e) {
+            if (S3BlobStore.isMissingKey(e)) {
+                // don't crash for a missing blob, even though it means the storage is corrupted
+                log.error("Failed to get information on blob: " + key, e);
+                return new BlobStatus();
```
RSalem07: can be replaced by `log.error("Failed to get information on blob: {}" ,  key, e);` but i think you main logging message are doing using the concatenation string message (i mean on others code part like  [here](https://github.com/nuxeo/nuxeo/blob/feature-NXP-28417-glacier-restore-and-status-10.10/addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java#L76)) perhaps we should keep it as it is to more consistency

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
```diff
@@ -179,4 +189,45 @@ protected String getContentDispositionHeader(Blob blob) {
...
+        String bucketKey = config.bucketPrefix + objectKey;
+        GetObjectMetadataRequest request = new GetObjectMetadataRequest(config.bucketName, bucketKey);
+        request.setVersionId(versionId);
+        ObjectMetadata metadata;
+        try {
+            metadata = config.amazonS3.getObjectMetadata(request);
+        } catch (AmazonServiceException e) {
+            if (S3BlobStore.isMissingKey(e)) {
+                // don't crash for a missing blob, even though it means the storage is corrupted
+                log.error("Failed to get information on blob: " + key, e);
```
RSalem07: can be replaced by log.error("Failed to get information on blob: {}" , key, e); but i think you main logging message are doing using the concatenation string message (i mean on others code part like [here](https://github.com/nuxeo/nuxeo/blob/feature-NXP-28417-glacier-restore-and-status-10.10/addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java#L76) perhaps we should keep it as it is to more consistency

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
RSalem07: should we returns directly a new instance of blob status as the default values will be: 
`downloadable=true` and `storageClass` is `null` which means a downloadable standard storage, and in fact we don't have our storage. perhapse should we add a code status or at least returns: 
```
 return new BlobStatus()withDownloadable(false);
```

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
```diff
@@ -676,6 +678,15 @@ public void writeBlobProperties(BlobUpdateContext blobUpdateContext) throws IOEx
                 logTrace("rnote right: " + status.toString());
                 amazonS3.setObjectLegalHold(request);
             }
+            if (blobUpdateContext.restoreForDuration != null) {
+                Duration duration = blobUpdateContext.restoreForDuration.duration;
+                // round up duration to days
+                int days = (int) duration.plusDays(1).minusSeconds(1).toDays();
+                RestoreObjectRequest request = new RestoreObjectRequest(bucketName, bucketKey);
+                request.setVersionId(versionId);
+                request.setExpirationInDays(days);
```
RSalem07: can be replaced by:
```
RestoreObjectRequest request = new RestoreObjectRequest(bucketName, bucketKey, days).withVersionId(versionId);
```

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
```diff
@@ -676,6 +678,15 @@ public void writeBlobProperties(BlobUpdateContext blobUpdateContext) throws IOEx
...
+                amazonS3.restoreObjectV2(request);
+            }
```
RSalem07: i am wondering if should use/explore the returned object:
` RestoreObjectResult restoreObjectResult = amazonS3.restoreObjectV2(request);`
mainly the use of `com.amazonaws.services.s3.model.RestoreObjectResult#restoreOutputPath`
```
    /**
     * The path in the provided S3 output location where Select results will be restored to
     */
    private String restoreOutputPath;
```

Or we shouldn't care about it and when the content is available  call the  `org.nuxeo.ecm.core.blob.BlobProvider#getStatus` then we can make the call to `org.nuxeo.ecm.core.blob.BlobProvider#getStream`

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobStatus.java
```diff
@@ -0,0 +1,74 @@
...
+        this.downloadableUntil = downloadableUntil;
+        return this;
+    }
+
+    /**
+     * The storage class, or {@code null} for the standard storage class.
+     */
+    public String getStorageClass() {
+        return storageClass;
+    }
```
RSalem07: what if we hide the fact that `null` is a the Standard storage as it is low level implementation detail (mainly AWS for instant), i think your idea is to use the blob status like more generic, but `...or {@code null} for the standard storage class.` is mainly an AWS implementation.
or to be more generic perhaps we shouldn't give this information and replace it by more generic term `storageDetails`

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
efge: Actually it should have been, I intended to wrap any `AmazonServiceException` into an `IOException`. I'll change that.

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
efge: I prefer the "static single assignment" form, where variables are assigned only once if possible and if it doesn't make the code unreadable.
Note that this pattern helps lambdas as well (not in this case, there's no lambda) as it makes the variables effectively-final.

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
efge: Thanks. I intended to use commons-lang3 but I still have indirect dependencies to commons-lang somewhere.

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
efge: `restoreOutputPath` is about "select" restores, which is something different.
In our case there's really nothing interesting in the result... It's just launching an async restore.

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobStatus.java
efge: I'll change so that the default storage class is always represented as `null`, even when calling `getStatus` (so if the storage returns a value that corresponds to its standard storage class I'll turn it into null). That helps to build a "standard" availability with just `new BlobStatus()`

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobProvider.java
efge: Good point.

efge:
Updated after suggestions.

PR #3751, opened 2020-02-17
NXP-28571: make abort old S3 multipart uploads async and optional

efge:


addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3BinaryManager.java
```diff
@@ -193,16 +195,32 @@ public void close() {
      *
      * @since 7.2
      */
-    protected void abortOldUploads() throws IOException {
+    protected void abortOldUploads() {
+        ConfigurationService configurationService = Framework.getService(ConfigurationService.class);
+        if (configurationService == null) {
```
ataillefer: in which case does it happen?

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStoreConfiguration.java
```diff
@@ -385,17 +392,33 @@ protected TransferManager createTransferManager() {
     /**
      * Aborts uploads that crashed and are older than 1 day.
      */
-    protected void abortOldUploads() throws IOException {
+    protected void abortOldUploads() {
+        ConfigurationService configurationService = Framework.getService(ConfigurationService.class);
+        if (configurationService == null) {
```
ataillefer: in which case does it happen?

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3BinaryManager.java
efge: It's not something I expect to happen, but I want to be robust in backports (and maybe some pure unit tests).

ataillefer:
As the code is duplicated, can you enlighten us about the purpose of S3BlobStoreConfiguration vs S3BinaryManager. Can the abort method be called twice?

efge:
`S3BinaryManager` is the old implem. `S3BlobStoreConfiguration` is the new implem.

Anyway please hold reviews, I'll change the property check to use the extension point instead of the configuration service.

efge:
@ataillefer udpated

PR #3752, opened 2020-02-17
NXP-28185: allow disabling usage of the proxy environment variables for the connection to the S3 endpoint

efge:


addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStoreConfiguration.java
```diff
@@ -122,6 +123,14 @@
      */
     public static final String MULTIPART_COPY_PART_SIZE_PROPERTY = "nuxeo.s3.multipart.copy.part.size";
 
+    /**
+     * Framework property to disable usage of the proxy environment variables ({@code nuxeo.http.proxy.*}) for the
+     * connection to the S3 endpoint.
+     *
+     * @since 11.1
+     */
+    public static final String DISABLE_PROXY_PROPERTY = "nuxeo.s3.proxy.disabled";
```
kevinleturc: Why not having a blob provider property instead of Framework one?

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStoreConfiguration.java
```diff
@@ -256,6 +265,7 @@ protected AWSCredentialsProvider getAWSCredentialsProvider() {
     }
 
     protected ClientConfiguration getClientConfiguration() {
+        boolean proxyDisabled = Boolean.parseBoolean(Framework.getProperty(DISABLE_PROXY_PROPERTY));
```
kevinleturc: In case it remains a Framework property, _Framework#isBooleanPropertyTrue_ should do the work.

addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStoreConfiguration.java
efge: This controls usage of the proxy configuration properties which are themselves framework properties, so it's logical to have the enabled/disabled flag be the same kind of property.

PR #3764, opened 2020-02-19
NXP-27654: add a unique index on ecm:id in MongoDB

efge:


nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
```diff
@@ -345,7 +345,7 @@ public void createStates(List<State> states) {
             // Avoid hiding any others bulk errors
             if (duplicates.size() == mbwe.getWriteErrors().size()) {
                 log.trace("MongoDB:    -> DUPLICATE KEY: {}", duplicates);
-                ConcurrentUpdateException concurrentUpdateException = new ConcurrentUpdateException();
+                ConcurrentUpdateException concurrentUpdateException = new ConcurrentUpdateException("Concurrent update");
```
ataillefer: is the message useful, or could it be richer?

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestSQLRepositoryAPI.java
```diff
@@ -3743,6 +3743,61 @@ public void testImport() {
         assertFalse(doc.isProxy());
     }
 
+    @Test
+    public void testImportDuplicateIdSameSession() {
+        DocumentModel doc = session.createDocumentModel("/", "doc", "File");
+        doc = session.createDocument(doc);
+        session.save();
+        nextTransaction();
```
ataillefer: could/should these 2 lines be replaced by
```
transactionalFeature.nextTransaction()
```
Though it's probably for backport purpose, don't we want to benefit from it on master?

efge:
https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/772/

nuxeo-core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBRepository.java
efge: The lines below this one call `addInfo` with the details of the duplicates. I'm adding `"Concurrent update"` here to be in sync in with what's done in VCS and have unit tests that work for both DBS and VCS.

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestSQLRepositoryAPI.java
efge: For tests I don't always bother to try to find the optimal sequence if it's not important to the test. Also I prefer to do the `save` explicitly because in some of the tests it's at save time that a failure occurs, and hiding it inside the commit's flush wouldn't help make things obvious. Finally `save` + `nextTransaction` is already used in a number of other places in this same file.

ataillefer: OK, we really need to write these unit testing guidelines as such statements aren't always obvious.

PR #3785, opened 2020-02-25
NXP-28558: disallow concurrent startup for scheduler service

efge:
https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/775/

### DIFF
--- a/nuxeo-common/src/test/java/org/nuxeo/common/utils/TestDurationUtils.java
+++ b/nuxeo-common/src/test/java/org/nuxeo/common/utils/TestDurationUtils.java
@@ -1,0 +1,78 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Florent Guillaume
+ */
+package org.nuxeo.common.utils;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import java.time.Duration;
+import java.time.format.DateTimeParseException;
+
+import org.junit.Test;
+
+public class TestDurationUtils {
+
+    protected static final Duration parse(String s) {
+        return DurationUtils.parse(s);
+    }
+
+    protected static final Duration parsePositive(String s) {
+        return DurationUtils.parsePositive(s, Duration.ofDays(-123));
+    }
+
+    @Test
+    public void testParse() {
+        // ISO 8601 format
+        assertEquals(Duration.ofSeconds(20).plusMillis(345), parse("PT20.345S"));
+        assertEquals(Duration.ofMinutes(15), parse("PT15M"));
+        assertEquals(Duration.ofHours(10), parse("PT10H"));
+        assertEquals(Duration.ofDays(2), parse("P2D"));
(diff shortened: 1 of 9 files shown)


nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/scheduler/SchedulerServiceImpl.java
```diff
@@ -139,11 +149,33 @@ public void deactivate(ComponentContext context) {
...
+                duration = Duration.ZERO;
+            } else {
+                duration = DurationUtils.parse(prop);
+            }
+        } catch (DateTimeParseException e) {
+            duration = Duration.ZERO;
+        }
+        if (duration.isZero() || duration.isNegative()) {
+            duration = Duration.ofMinutes(1);
         }
```
troger: Third time we do the same code, WDYT about adding a method to `ClusterService` or a helper to retrieve the duration from a given Framework property name?

On 10.10 and 9.10, it could be on the `ClusterLockHelper`.

efge: How about I put it in `DurationUtils` instead?

efge: I pushed 2 more commits for this

nuxeo-common/src/test/java/org/nuxeo/common/utils/TestDurationUtils.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.junit.Assert.fail;
+
+import java.time.Duration;
+import java.time.format.DateTimeParseException;
+
+import org.junit.Test;
+
+public class TestDurationUtils {
+
+    protected static final Duration parse(String s) {
```
ataillefer: Is there an interest for static?

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/scheduler/SchedulerServiceImpl.java
```diff
@@ -154,20 +151,8 @@ public void start(ComponentContext context) {
...
-                duration = DurationUtils.parse(prop);
-            }
-        } catch (DateTimeParseException e) {
-            duration = Duration.ZERO;
-        }
-        if (duration.isZero() || duration.isNegative()) {
-            duration = Duration.ofMinutes(1);
-        }
+        String prop = Framework.getProperty(CLUSTER_START_DURATION_PROP);
+        Duration duration = DurationUtils.parsePositive(prop, Duration.ofMinutes(1));
```
ataillefer: It might be more clear to have 1 as a constant: since there is no default for `org.nuxeo.scheduler.cluster.start.duration`, most of the time it is this default parameter of `DurationUtils#parsePositive` that will be used.

Though it's the caller's responsibility to set the timeout, as we are always using 1 minute, it could be in DurationUtils:
```
public static final long DURATION_DEFAULT_1_M = 1;
```

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/scheduler/SchedulerServiceImpl.java
troger: Even better 👍

nuxeo-common/src/test/java/org/nuxeo/common/utils/TestDurationUtils.java
efge: It's an indirection for a static method... But I could go either way. I'll leave it as is though as it doesn't change the test.

nuxeo-core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/scheduler/SchedulerServiceImpl.java
efge: Ok for a constant but I don't want to push it to `DurationUtils` itself, I think it's too low level to have any responsibility in the default value.

PR #3800, opened 2020-02-27
feature-NXP-28618-unit-tests-elasticsearch-cluster

ataillefer:


kevinleturc:
We're going to set `@Repeat` annotation on random tests + create a Jira ticket for each of them.

PR #3802, opened 2020-02-28
NXP-28599: add option to generate download URLs with redirect already taken into account

efge:


efge:
I will force push a better version using `ConfigurationService` and with tests.

nuxeo-core/nuxeo-core-io/src/main/java/org/nuxeo/ecm/core/io/download/DownloadServiceImpl.java
```diff
@@ -167,6 +171,22 @@ public String storeBlobs(List<Blob> blobs) {
...
+    public String getFullDownloadUrl(DocumentModel doc, String xpath, Blob blob, String baseUrl) {
+        ConfigurationService configurationService = Framework.getService(ConfigurationService.class);
+        if (configurationService.isBooleanTrue(DOWNLOAD_URL_FOLLOW_REDIRECT)) {
+            try {
+                URI uri = redirectResolver.getURI(blob, UsageHint.DOWNLOAD, null);
+                if (uri != null) {
+                    return uri.toString();
+                }
+            } catch (IOException e) {
+                log.error(e, e);
```
NourNuxeo: maybe logging the baseUrl and Doc would be helpfull ?

efge: We're not expecting them to fail in the current implementations, `getURI` is just generating a URL. But `IOException` is still in the signature. Not worth having anything more complex for now.

PR #3811, opened 2020-03-04
NXP-28732: Allow to customize the S3 token TTL

BoboTiG:


packages/nuxeo-amazon-s3-package/src/main/resources/install/templates/s3binaries/nxserver/config/s3directupload-config.xml.nxftl
```diff
@@ -76,6 +76,7 @@
       <property name="awstoken">${nuxeo.s3storage.transient.awstoken}</property>
       <property name="bucket">${nuxeo.s3storage.transient.bucket}</property>
       <property name="endpoint">${nuxeo.s3storage.transient.endpoint}</property>
+      <property name="expiration">${nuxeo.s3storage.transient.expiration}</property>
```
efge: You must also add a default value in `s3binaries/nuxeo.defaults`

efge:
However the commit message and the title are not very clear about what this is about. It's specific to direct upload.

BoboTiG:
Yes, you are right. Something like that may be better?
> Allow to customize the S3 direct upload token expiration

PR #3823, opened 2020-03-10
Task NXP-27443 Reorganize Repository

kevinleturc:


Jenkinsfile
```diff
@@ -214,13 +214,13 @@ def buildUnitTestStage(env) {
                 envsubst < ci/mvn/nuxeo-test-${env}.properties > ${HOME}/nuxeo-test-${env}.properties
             """
             // run unit tests:
-            // - in nuxeo-core and dependent projects only (nuxeo-common and nuxeo-runtime are run in dedicated stages)
+            // - in modules/core and dependent projects only (modules/runtime are run in dedicated stages)
```
ataillefer:
```suggestion
            // - in modules/core and dependent projects only (modules/runtime is run in a dedicated stage)
```

ftests/nuxeo-server-gatling-tests/pom.xml
```diff
@@ -0,0 +1,268 @@
...
+  <parent>
+    <groupId>org.nuxeo.ftests</groupId>
+    <artifactId>nuxeo-ftests</artifactId>
+    <version>11.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>nuxeo-server-gatling-tests</artifactId>
+  <name>Nuxeo Server Gatling Bench</name>
+  <description>Gatling bench on Nuxeo Server</description>
+  <packaging>jar</packaging>
```
ataillefer: we could remove it since it's the default value

kevinleturc: I think @troger tried to switch to pom during the `ftests` creation without success.

I'm removing the gatling packaging for now and let this discussion open.

PR #3825, opened 2020-03-11
NXP-28621: add back and deprecate 'mp-init' nuxeoctl command

troger:


nuxeo-distribution/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -2708,6 +2711,16 @@ protected boolean pkgRefreshCache() {
         return true;
     }
 
+    /**
+     * Add packages from the distribution to the local cache
+     *
+     * @since 5.6
+     */
+    protected boolean pkgInit() {
```
ataillefer: Do we want to mark it as `@Deprecated`?

troger: I really don't think it's useful on the launcher...

PR #3829, opened 2020-03-13
NXP-28688: Retrieve the blob size for items returned by NuxeoDrive.GetChildren

BoboTiG:
The test and push worked the 1st time but not the 2nd one because of a failure on "Nuxeo 3D Package JSF UI Functional Tests". So I think this is still safe to open the PR.

cf https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-mschoentgen/2.


addons/nuxeo-drive-server/nuxeo-drive-core/src/main/java/org/nuxeo/drive/adapter/impl/SimpleFileSystemItem.java
```diff
@@ -52,6 +56,10 @@ public String getDigest() {
         return digest;
     }
 
+    public long getSize() {
```
ataillefer: please add `@since` as it's added for the setter

PR #3843, opened 2020-03-18
NXP-28600: add WriteColdStorage permission

nmpcunha:
🔵 T&P: https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-ncunha/39/

### DIFF
--- a/nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/security/SecurityConstants.java
+++ b/nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/security/SecurityConstants.java
@@ -132,4 +132,10 @@ public interface SecurityConstants {
      */
     String MANAGE_LEGAL_HOLD = "ManageLegalHold";
 
+    /**
+     * Permission needed to move document's main content to cold storage.
+     *
+     * @since 11.1
+     */
+    String WRITE_COLD_STORAGE = "WriteColdStorage";
 }
--- a/nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
+++ b/nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
@@ -15,6 +15,7 @@
  *
  * Contributors:
  *     Salem Aouana
+ *     Nuno Cunha <ncunha@nuxeo.com>
  */
 
 package org.nuxeo.ecm.core;
@@ -46,10 +47,16 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.nuxeo.ecm.core.api.Blob;
 import org.nuxeo.ecm.core.api.Blobs;
+import org.nuxeo.ecm.core.api.CloseableCoreSession;
+import org.nuxeo.ecm.core.api.CoreInstance;
 import org.nuxeo.ecm.core.api.CoreSession;
 import org.nuxeo.ecm.core.api.DocumentModel;
 import org.nuxeo.ecm.core.api.IdRef;
 import org.nuxeo.ecm.core.api.NuxeoException;
+import org.nuxeo.ecm.core.api.security.ACE;
+import org.nuxeo.ecm.core.api.security.ACL;
+import org.nuxeo.ecm.core.api.security.ACP;
+import org.nuxeo.ecm.core.api.security.SecurityConstants;
 import org.nuxeo.ecm.core.blob.BlobManager;
 import org.nuxeo.ecm.core.blob.BlobStatus;
 import org.nuxeo.ecm.core.blob.ColdStorageHelper;
@@ -90,23 +97,34 @@ public class TestColdStorage {
 
     @Test
     public void shouldMoveToColdStorage() throws IOException {
-        DocumentModel documentModel = createFileDocument(DEFAULT_DOC_NAME, true);
+        // with regular user with "WriteColdStorage" permission
+        ACE[] aces = { new ACE("john", SecurityConstants.READ, true), //
+                new ACE("john", SecurityConstants.WRITE, true), //
+                new ACE("john", SecurityConstants.WRITE_COLD_STORAGE, true) };
+        DocumentModel documentModel = createFileDocument(DEFAULT_DOC_NAME, true, aces);
 
-        // move the blob to cold storage
-        documentModel = ColdStorageHelper.moveContentToColdStorage(session, documentModel.getRef());
-        session.saveDocument(documentModel);
-        transactionalFeature.nextTransaction();
-        documentModel.refresh();
+        try (CloseableCoreSession userSession = CoreInstance.openCoreSession(documentModel.getRepositoryName(),
+                "john")) {
+            moveAndVerifyContent(userSession, documentModel);
+        }
 
-        assertTrue(documentModel.hasFacet(FacetNames.COLD_STORAGE));
+        // with Administrator
+        documentModel = createFileDocument(DEFAULT_DOC_NAME, true);
+        moveAndVerifyContent(session, documentModel);
+    }
 
-        assertNull(documentModel.getPropertyValue(ColdStorageHelper.FILE_CONTENT_PROPERTY));
+    @Test
+    public void shouldFailWithoutRightPermissions() {
+        ACE[] aces = { new ACE("john", SecurityConstants.READ, true) };
+        DocumentModel documentModel = createFileDocument(DEFAULT_DOC_NAME, true, aces);
 
-        // check if the `coldstorage:coldContent` property contains the original file content
-        Blob content = (Blob) documentModel.getPropertyValue(ColdStorageHelper.COLD_STORAGE_CONTENT_PROPERTY);
-        assertNotNull(content);
-        assertEquals(FILE_CONTENT, content.getString());
-        assertEquals("dummy", ((ManagedBlob) content).getProviderId());
+        try (CloseableCoreSession userSession = CoreInstance.openCoreSession(documentModel.getRepositoryName(),
+                "john")) {
+            ColdStorageHelper.moveContentToColdStorage(userSession, documentModel.getRef());
+            fail("Should fail because the user does not have permissions to move document to cold storage");
+        } catch (NuxeoException e) {
+            assertEquals(SC_FORBIDDEN, e.getStatusCode());
+        }
     }
 
     @Test
@@ -229,6 +247,23 @@ public class TestColdStorage {
         checkAvailabilityOfDocuments(Arrays.asList(documents.get(1), documents.get(2)), downloadableUntil, 0);
     }
 
+    protected void moveAndVerifyContent(CoreSession session, DocumentModel documentModel) throws IOException {
+        documentModel = ColdStorageHelper.moveContentToColdStorage(session, documentModel.getRef());
+        session.saveDocument(documentModel);
+        transactionalFeature.nextTransaction();
+        documentModel.refresh();
+
+        assertTrue(documentModel.hasFacet(FacetNames.COLD_STORAGE));
+
+        assertNull(documentModel.getPropertyValue(ColdStorageHelper.FILE_CONTENT_PROPERTY));
+
+        // check if the `coldstorage:coldContent` property contains the original file content
+        Blob content = (Blob) documentModel.getPropertyValue(ColdStorageHelper.COLD_STORAGE_CONTENT_PROPERTY);
+        assertNotNull(content);
+        assertEquals(FILE_CONTENT, content.getString());
+        assertEquals("dummy", ((ManagedBlob) content).getProviderId());
+    }
+
     protected DocumentModel moveAndRequestRetrievalFromColdStorage(String documentName) {
         DocumentModel documentModel = createFileDocument(documentName, true);
         documentModel = ColdStorageHelper.moveContentToColdStorage(session, documentModel.getRef());
@@ -238,12 +273,19 @@ public class TestColdStorage {
         return session.saveDocument(documentModel);
     }
 
-    protected DocumentModel createFileDocument(String name, boolean addBlobContent) {
+    protected DocumentModel createFileDocument(String name, boolean addBlobContent, ACE... aces) {
         DocumentModel documentModel = session.createDocumentModel("/", name, "File");
         if (addBlobContent) {
             documentModel.setPropertyValue("file:content", (Serializable) Blobs.createBlob(FILE_CONTENT));
         }
-        return session.createDocument(documentModel);
+        DocumentModel document = session.createDocument(documentModel);
+        if (aces.length > 0) {
+            ACP acp = documentModel.getACP();
+            ACL acl = acp.getOrCreateACL();
+            acl.addAll(List.of(aces));
+            document.setACP(acp, true);
+        }
+        return document;
     }
 
     protected void checkAvailabilityOfDocuments(List<String> expectedAvailableDocIds, Instant downloadableUntil,
--- a/nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
+++ b/nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
@@ -15,6 +15,7 @@
  *
  * Contributors:
  *     Salem Aouana
+ *     Nuno Cunha <ncunha@nuxeo.com>
  */
 
 package org.nuxeo.ecm.core.blob;
@@ -37,6 +38,7 @@ import org.nuxeo.ecm.core.api.DocumentModel;
 import org.nuxeo.ecm.core.api.DocumentModelList;
 import org.nuxeo.ecm.core.api.DocumentRef;
 import org.nuxeo.ecm.core.api.NuxeoException;
+import org.nuxeo.ecm.core.api.security.SecurityConstants;
 import org.nuxeo.ecm.core.event.EventService;
 import org.nuxeo.ecm.core.event.impl.DocumentEventContext;
 import org.nuxeo.ecm.core.io.download.DownloadService;
@@ -71,14 +73,24 @@ public class ColdStorageHelper {
 
     /**
      * Moves the main content associated with the document of the given {@link DocumentRef} to a cold storage.
+     * <p/>
+     * The permission {@value org.nuxeo.ecm.core.api.security.SecurityConstants#WRITE_COLD_STORAGE} is required.
      *
      * @return the updated document model if the move succeeds
-     * @throws NuxeoException if the main content is already in the cold storage, or if there is no main content
-     *             associated with the given document
+     * @throws NuxeoException if the main content is already in the cold storage, if there is no main content
+     *             associated with the given document, or if the user does not have the permissions needed to
+     *             perform the action.
      */
     public static DocumentModel moveContentToColdStorage(CoreSession session, DocumentRef documentRef) {
         DocumentModel documentModel = session.getDocument(documentRef);
         log.debug("Move to cold storage the main content of document: {}", documentModel);
+
+        if (!session.hasPermission(documentRef, SecurityConstants.WRITE_COLD_STORAGE)) {
+            log.debug("The user {} does not have the right permissions to move the content of document {}",
+                    session::getPrincipal, () -> documentModel);
+            throw new NuxeoException(String.format("The document: %s cannot be moved to cold storage", documentRef),
+                    SC_FORBIDDEN);
+        }
 
         if (documentModel.hasFacet(FacetNames.COLD_STORAGE)
                 && documentModel.getPropertyValue(COLD_STORAGE_CONTENT_PROPERTY) != null) {
--- a/nuxeo-features/nuxeo-automation/nuxeo-automation-features/src/test/java/org/nuxeo/ecm/automation/core/operations/coldstorage/MoveToColdStorageTest.java
+++ b/nuxeo-features/nuxeo-automation/nuxeo-automation-features/src/test/java/org/nuxeo/ecm/automation/core/operations/coldstorage/MoveToColdStorageTest.java
@@ -15,11 +15,13 @@
  *  
  * Contributors:
  *     Salem Aouana
+ *     Nuno Cunha <ncunha@nuxeo.com>
  */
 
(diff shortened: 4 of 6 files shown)


nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
```diff
@@ -71,24 +73,27 @@
...
      *             associated with the given document
      */
     public static DocumentModel moveContentToColdStorage(CoreSession session, DocumentRef documentRef) {
         DocumentModel documentModel = session.getDocument(documentRef);
+        Serializable mainContent = documentModel.getPropertyValue(FILE_CONTENT_PROPERTY);
         log.debug("Move to cold storage the main content of document: {}", documentModel);
 
-        if (documentModel.hasFacet(FacetNames.COLD_STORAGE)
+        if (!session.getPrincipal().isAdministrator()
+                && !session.hasPermission(documentRef, SecurityConstants.WRITE_COLD_STORAGE)) {
```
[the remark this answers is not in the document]
nmpcunha: Thanks! 👍

nuxeo-features/nuxeo-automation/nuxeo-automation-features/src/test/java/org/nuxeo/ecm/automation/core/operations/coldstorage/MoveToColdStorageTest.java
```diff
@@ -42,8 +50,44 @@
...
+        ACL acl = acp.getOrCreateACL();
+        acl.add(new ACE("john", SecurityConstants.READ, true));
+        acl.add(new ACE("john", SecurityConstants.WRITE, true));
+        acl.add(new ACE("john", SecurityConstants.WRITE_COLD_STORAGE, true));
+        documentModel.setACP(acp, true);
+
+        try (CloseableCoreSession userSession = coreFeature.openCoreSession("john")) {
+            moveContentToColdStorage(userSession, documentModel);
+        } catch (NuxeoException e) {
+            fail("It was not expected an exception while moving document to cold storage since user has the required permissions");
```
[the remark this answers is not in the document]
nmpcunha: Thanks for your input!
Do you have any opinion about the "duplication" of tests here? Since we have generically the same tests on `MoveToColdStorageTest` and `TestColdStorage` classes.

The first one could, eventually, just ensure ColdStorageHelper is called and that a thumbnail is inserted as main content. But, in fact, this helper has static methods and we won't easily mock them unless we insert new dependencies and so on.
Should this helper become a service? Any clear trade off with that approach?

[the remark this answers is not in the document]
efge: I don't know the code in detail but yes test duplication is not ideal. I don't know the tests enough to give an opinion on how to best clean this up. If you decide to do a cleanup then please do it in a separate commit btw.

The helper could become a service yes, but that's more code to write. If there's a benefit then yes it could be done.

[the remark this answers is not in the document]
RSalem07: As discussed, the main purpose of adding tests on the operation code is to ensure the correct integration between low level code (helper class) and the operation code (because the whole layer are part of cold storage feature) => mainly **Integration Test**:
- I ensure that the returned document in operation is correctly updated: the reason why I added some checks that exist on low level, plus others related to the `thumbnail`, notifications...
- on others tests on operations I followed the same logic about integration tests

I asked myself question, should I or not add the seconds check on operations, initially as i explained I tried to add a Mock like you tried, but Mockito don't allow the mock of static class, and even i tried to add the PowerMock library / dependency which allows the mock of statics class, but I abandoned it because it's more things and when we started the developments the idea to fo further in the dev, but we discussed with Florent/Antoine that perhaps in a future we will change the low level into service and i ended by adding the tests as explained above which allow more flexibility on mocking things ....

If we use this feature (cold storing) in others feature, and we test these new features than we must not do any check on the cold storage, if we do then it's a duplicate tests...


that's my point of view :)

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/security/SecurityConstants.java
```diff
@@ -132,4 +132,10 @@
      */
     String MANAGE_LEGAL_HOLD = "ManageLegalHold";
 
+    /**
+     * Permission needed to move/transfer documents to cold storage.
+     *
```
RSalem07: `Permission needed to move/transfer documents to cold storage.` -> can you please rename to include the fact that we transfer the content of the document instead of `move/transfer documents to cold storage.` -> the idea is to be consistent with what we did:
for example javadoc of the move method on the helper:
`Moves the main content associated with the document of the given {@link DocumentRef} to a cold storage...`

RSalem07: `ManageLegalHold ` should be added in a xml contribution to be visible outer internal Nuxeo code,  can you have a look at [here](https://github.com/nuxeo/nuxeo/blob/master/nuxeo-core/nuxeo-core/src/main/resources/OSGI-INF/permissions-contrib.xml) I think without this contribution, the permission will not be visible on the web-ui ...to check

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
```diff
@@ -88,8 +95,11 @@
     @Inject
     protected DownloadService downloadService;
 
+    @Inject
+    protected CoreFeature coreFeature;
+
```
RSalem07: can you remove this injection and replace the open of the core session with: 
` 
try (CloseableCoreSession jamesSession = CoreInstance.openCoreSession(doc.getRepositoryName(), "john")) {
....
}
`
complete example [here](https://github.com/nuxeo/nuxeo/blob/master/nuxeo-features/nuxeo-platform-comment/nuxeo-platform-comment/src/test/java/org/nuxeo/ecm/platform/comment/impl/TestTreeCommentManager.java#L240)

nuxeo-features/nuxeo-automation/nuxeo-automation-features/src/test/java/org/nuxeo/ecm/automation/core/operations/coldstorage/MoveToColdStorageTest.java
```diff
@@ -42,8 +50,42 @@
     @Inject
     protected CoreSession session;
 
+    @Inject
+    protected CoreFeature coreFeature;
+
+    @Test
+    public void shouldFailWhenUserDoNotHaveWriteColdStoragePermission() throws OperationException, IOException {
+        DocumentModel documentModel = createFileDocument(session, true);
```
RSalem07: I am wondering if we can keep the methods names shorter as possible -> 
`shouldFailWithoutRightPermissions` we don't need to explicitly put the permission in the name of the method, which will avoid renaming the method if the permission changes its names in the future

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
```diff
@@ -109,6 +119,48 @@ public void shouldMoveToColdStorage() throws IOException {
         assertEquals("dummy", ((ManagedBlob) content).getProviderId());
     }
```
RSalem07: I think we can group the tests within the permissions to:
- keep the name shorted
- group the tests related to same things on the same tests

as we have a helper method, i think we can use a name:

`shouldMoveToColdStorage` which can test with an Admin and after that with a regular user that has the right permission

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
```diff
@@ -109,6 +119,48 @@ public void shouldMoveToColdStorage() throws IOException {
...
+    @Test
+    public void shouldMoveToColdStorageWhenUserHaveWriteColdStoragePermission() throws IOException {
+        DocumentModel documentModel = createFileDocument(DEFAULT_DOC_NAME, true);
+        ACP acp = documentModel.getACP();
+        ACL acl = acp.getOrCreateACL();
+        acl.add(new ACE("john", SecurityConstants.READ, true));
+        acl.add(new ACE("john", SecurityConstants.WRITE, true));
+        acl.add(new ACE("john", SecurityConstants.WRITE_COLD_STORAGE, true));
+        documentModel.setACP(acp, true);
+
```
RSalem07: I am wondering if we can reduce this part of code and if we do wan reduce code on others parts. For example i was thinking if we can transform the `createFileDocument` to:
```
    protected DocumentModel createFileDocument(CoreSession session, boolean withBlobContent, ACE... aces) {
        DocumentModel documentModel = session.createDocumentModel("/", "MyFile", "File");
        if (withBlobContent) {
            documentModel.setPropertyValue(ColdStorageHelper.FILE_CONTENT_PROPERTY,
                    (Serializable) Blobs.createBlob(FILE_CONTENT));
        }
        DocumentModel document = session.createDocument(documentModel);
        if(aces.length > 0) {
            ACP acp = documentModel.getACP();
            ACL acl = acp.getOrCreateACL();
            acl.addAll(List.of(aces));
            document.setACP(acp, true);
        }
        return document;
    }
``` 

and the callers can be: 
`DocumentModel documentModel = createFileDocument(session, true, new ACE("john", SecurityConstants.READ, true));`
or
```
 ACE[] aces = { new ACE("john", SecurityConstants.READ, true), //
                new ACE("john", SecurityConstants.WRITE, true), //
                new ACE("john", SecurityConstants.WRITE_COLD_STORAGE, true) };

        DocumentModel documentModel = createFileDocument(session, true, aces);
```

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
```diff
@@ -71,6 +73,7 @@
 
     /**
      * Moves the main content associated with the document of the given {@link DocumentRef} to a cold storage.
+     * The permission {@value org.nuxeo.ecm.core.api.security.SecurityConstants#WRITE_COLD_STORAGE} is required.
```
RSalem07: can you add `<p/>` before the new line `The permission {@value org.nuxeo.ecm.core.api.security.SecurityConstants#WRITE_COLD_STORAGE} is required.` and can you complete the `@throws NuxeoException`

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/security/SecurityConstants.java
nmpcunha: I'll do it, thanks! 👍

nuxeo-features/nuxeo-automation/nuxeo-automation-features/src/test/java/org/nuxeo/ecm/automation/core/operations/coldstorage/MoveToColdStorageTest.java
nmpcunha: We can, but I would say it may not be a consensual approach.
Some schools of thought advocate that unit test names should be clear and map specific requirements/acceptance criteria and, in this case, we can end with long names. Because, at the end of the day, who reads the test name should understand what is being tested.

If the permissions change in future, the requirements changed and we probably want to change the tests too 🙂 

Anyway, I can try to follow other approach for consistency sake. What is our convention?

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
nmpcunha: Yes, we can do it. 
Again, we'll duplicate similar logic on the following files:

- AbstractTestColdStorageOperation (https://github.com/nuxeo/nuxeo/blob/master/nuxeo-features/nuxeo-automation/nuxeo-automation-features/src/test/java/org/nuxeo/ecm/automation/core/operations/coldstorage/AbstractTestColdStorageOperation.java#L75-L82)
- TestColdStorage (https://github.com/nuxeo/nuxeo/blob/master/nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java#L241-L247)

nmpcunha: Ok! Let's factor the ACEs 👍

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
nmpcunha: As we discussed, I'll wait for more input on this. I confess I'm not completely comfortable with this approach since we are violating some unit tests principles.
Obviously we can argue if we are doing unit or integration tests and it may influence the decision.

At the end I'll go with the more consensual approach but hoping that this discussion could help creating some awareness for this subjects 🙂

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/security/SecurityConstants.java
RSalem07: thanks, sorry i meant `WriteColdStorage` but i think you got it

nuxeo-features/nuxeo-automation/nuxeo-automation-features/src/test/java/org/nuxeo/ecm/automation/core/operations/coldstorage/MoveToColdStorageTest.java
ataillefer: @nmpcunha, unfortunately, for now, we don't have such conventions.
We plan to provide them in a near future.

So, I guess it's the responsibility of the writer to choose something that is at least consistent with the rest of the test/package/Maven module.
In this case, I would at least replace `WhenUserDoesNotHave` by the suggested `Without`.
We really want to keep the test names as short as possible for readability in the test logs, the CI unit test reports, etc.
As for the permission name, that's your choice, though personally I like @RSalem07 's approach, simply because it's shorter.

nmpcunha: Thanks for your input! 👍 
I've reworked the tests according to your suggestions. Hopefully we can work together on those conventions 🙂

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
```diff
@@ -71,6 +73,8 @@
 
     /**
      * Moves the main content associated with the document of the given {@link DocumentRef} to a cold storage.
+     * <p/>
+     * The permission {@value org.nuxeo.ecm.core.api.security.SecurityConstants#WRITE_COLD_STORAGE} is required.
      *
      * @return the updated document model if the move succeeds
      * @throws NuxeoException if the main content is already in the cold storage, or if there is no main content
```
ataillefer: Can you please add the case where the current user doesn't have the required permission?

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
```diff
@@ -80,6 +84,11 @@ public static DocumentModel moveContentToColdStorage(CoreSession session, Docume
         DocumentModel documentModel = session.getDocument(documentRef);
         log.debug("Move to cold storage the main content of document: {}", documentModel);
 
+        if (!session.hasPermission(documentRef, SecurityConstants.WRITE_COLD_STORAGE)) {
+            throw new NuxeoException(String.format("The document: %s cannot be moved to cold storage", documentRef),
```
ataillefer: You could be more precise on the message about the missing required permission.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
nmpcunha: Thanks! Nice catch! 🙂

nmpcunha: For debugging purposes I understand that it would be interesting to have more details.
But, if I am not mistaken, this exception will be sent to client side and, for security purposes, we should not provide too much details (IMHO).

ataillefer: You are referring to https://github.com/nuxeo/nuxeo/pull/3843#discussion_r395772285.
Don't know if it's considered 
as a security issue or if we want to provide this information to the client, @efge WDYT?

nmpcunha: @ataillefer, not sure how the comment ended here 😅
Moved to the correct thread!

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
nmpcunha: For debugging purposes I understand that it would be interesting to have more details.
But, if I am not mistaken, this exception will be sent to client side and, for security purposes, we should not provide too much details (IMHO).

efge: Yes we try to avoid too many details in the exceptions especially when related to security. It can be worth adding `log.debug` statements with the details if you feel then can be useful.

However in this particular case "cannot be moved to cold storage" seems sufficient to understand the issue.

nmpcunha: Added!
Thanks for your input @efge and @ataillefer!

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
```diff
@@ -90,23 +97,34 @@
...
         // move the blob to cold storage
-        documentModel = ColdStorageHelper.moveContentToColdStorage(session, documentModel.getRef());
-        session.saveDocument(documentModel);
-        transactionalFeature.nextTransaction();
-        documentModel.refresh();
-
-        assertTrue(documentModel.hasFacet(FacetNames.COLD_STORAGE));
-
-        assertNull(documentModel.getPropertyValue(ColdStorageHelper.FILE_CONTENT_PROPERTY));
+        moveAndVerifyContent(session, documentModel);
```
ataillefer: If I'm not mistaken, this second case is for the Administrator user, so:
- it doesn't need aces when creating the document
- it deserves a comment above like "// with administrator user"

nuxeo-core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestColdStorage.java
```diff
@@ -90,23 +97,34 @@
...
+    }
 
-        // check if the `coldstorage:coldContent` property contains the original file content
-        Blob content = (Blob) documentModel.getPropertyValue(ColdStorageHelper.COLD_STORAGE_CONTENT_PROPERTY);
-        assertNotNull(content);
-        assertEquals(FILE_CONTENT, content.getString());
-        assertEquals("dummy", ((ManagedBlob) content).getProviderId());
+    @Test
+    public void shouldFailWhenUserDoNotHaveWriteColdStoragePermission() {
```
ataillefer: For consistency, since you've renamed the test in [MoveToColdStorageTest#shouldFailWithoutRightPermissions](https://github.com/nuxeo/nuxeo/pull/3843/files#diff-cbed78a98d097ba042904ae895850590R52), don't you want to have the same name here?

nmpcunha: Nice catch, thanks!
Yes, clearly a side of effect of maintaining duplicated tests... 😞

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
ataillefer: It seems you haven't updated the javadoc for `@throws NuxeoException` to add the permission case.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/ColdStorageHelper.java
```diff
@@ -80,6 +84,13 @@ public static DocumentModel moveContentToColdStorage(CoreSession session, Docume
         DocumentModel documentModel = session.getDocument(documentRef);
         log.debug("Move to cold storage the main content of document: {}", documentModel);
 
+        if (!session.hasPermission(documentRef, SecurityConstants.WRITE_COLD_STORAGE)) {
+            log.debug("The user {} does not have the right permissions to move the content of document {}",
+                    session.getPrincipal(), documentModel);
```
ataillefer: Can you please use a lambda here to avoid systematically fetching the principal?
```
session::getPrincipal, () -> documentModel
```

PR #3856, opened 2020-03-24
Feature nxp 27713 nuxeo staging preview

NourNuxeo:


Jenkinsfile
```diff
@@ -631,6 +636,77 @@ pipeline {
       }
     }
 
+    stage('Deploy Preview') {
+      when {
+        branch "${REFERENCE_BRANCH}"
```
troger: Use the same `when` as other stages:
```
when {
        not {
          branch 'PR-*'
        }
      }
```

Jenkinsfile
```diff
@@ -282,6 +282,11 @@ pipeline {
     DOCKER_TAG = getDockerTagFrom("${VERSION}")
     CHANGE_BRANCH = "${env.CHANGE_BRANCH != null ? env.CHANGE_BRANCH : BRANCH_NAME}"
     CHANGE_TARGET = "${env.CHANGE_TARGET != null ? env.CHANGE_TARGET : BRANCH_NAME}"
+    ORG = "nuxeo"
+    APP_NAME = "nuxeo"
+    NAMESPACE = "$APP_NAME-${BRANCH_NAME.toLowerCase()}"
+    REFERENCE_BRANCH = "master"
```
troger: No need for this var, either it's a PR, either it's a reference branch.

Jenkinsfile
```diff
@@ -631,6 +636,77 @@ pipeline {
...
+      when {
+        branch "${REFERENCE_BRANCH}"
+      }
+      steps {
+        setGitHubBuildStatus('nuxeo/preview/deploy', 'Deploy nuxeo Preview', 'PENDING')
+        container('maven') {
+          dir('docker/nuxeo') {
+            echo """
+            ----------------------------------------
+            Build Nuxeo preview image
```
troger: Images are already built by another stage "Build Docker images", no need to build them again...

helm/preview/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
```
troger: All the `preview` folder should go in the already existing `ci/helm/` folder, no need to create a new `helm` folder at the root.

Jenkinsfile
```diff
@@ -284,6 +284,11 @@ pipeline {
     CHANGE_BRANCH = "${env.CHANGE_BRANCH != null ? env.CHANGE_BRANCH : BRANCH_NAME}"
     CHANGE_TARGET = "${env.CHANGE_TARGET != null ? env.CHANGE_TARGET : BRANCH_NAME}"
     CONNECT_PREPROD_URL = 'https://nos-preprod-connect.nuxeocloud.com/nuxeo'
+    ORG = "nuxeo"
```
ataillefer: unused

helm/preview/values.yaml
```diff
@@ -0,0 +1,86 @@
...
+          nginx.ingress.kubernetes.io/server-snippet: ignore_invalid_headers off;
+      labels:
+        team: platform
+        resource: service
+        usage: preview
+    image:
+      repository: ${DOCKER_REGISTRY}/nuxeo/nuxeo
+      tag: ${VERSION}
+      pullPolicy: Always
+    packages: nuxeo-web-ui nuxeo-dam
```
ataillefer: we don't need any package

Jenkinsfile
NourNuxeo: This is mandatory

ci/helm/preview/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+    master:
+      persistence:
+        enabled: ${PERSISTENCE}
+    data:
+      persistence:
+        enabled: ${PERSISTENCE}
+    cluster:
+      env:
+        EXPECTED_MASTER_NODES: 1
+        RECOVER_AFTER_MASTER_NODES: 1
```
kevinleturc: This should be in [nuxeo-helm-chart](https://github.com/nuxeo/nuxeo-helm-chart) repository.

Jenkinsfile
```diff
@@ -303,393 +307,61 @@ pipeline {
...
-              string(name: 'NUXEO_COMMIT_SHA', value: "${GIT_COMMIT}"),
-              string(name: 'NUXEO_VERSION', value: "${VERSION}"),
-            ], propagate: false, wait: false
+          setGitHubBuildStatus('nuxeo/preview/deploy', 'Deploy nuxeo Preview', 'FAILURE')
         }
       }
     }
   }
-
+
```
kevinleturc: Whitespace.

Jenkinsfile
```diff
@@ -303,393 +307,61 @@ pipeline {
...
+              // third build and deploy the chart
+              // waiting for https://github.com/jenkins-x/jx/issues/5797 to be fixed in order to remove --source-url
+              sh """
+                jx step helm build --verbose
+                mkdir target && helm template . --output-dir target
+                jx preview \
+                  --namespace ${NAMESPACE} \
+                  --verbose \
+                  --source-url=https://github.com/nuxeo/nuxeo \
+                  --preview-health-timeout 15m \
```
kevinleturc: This was added in arender because nuxeo 10.10 hotfixes itself during boot. This is not the case with master, so:
- probably not needed
- can be lowered

ci/helm/preview/values.yaml
ataillefer: Yes probably, it depends on if we want them all the time or just for the preview.
But in fact:
- What does these values correspond to? I cannot find the in the elasticsearch Helm chart: https://github.com/helm/charts/blob/master/stable/elasticsearch/values.yaml.
- Why do we need them set this way?

kevinleturc: In the context of this PR preview environment, we need them for the cluster to recover because it didn't find preview-elasticsearch-discovery service. The non-existence of the service is strange, it should exist at the same time than pod is created.

These environment values are used in the config map of elasticsearch chart: https://github.com/helm/charts/blob/master/stable/elasticsearch/templates/configmap.yaml
We can see that `MINIMUM_MASTER_NODES`, `EXPECTED_MASTER_NODES ` and `RECOVER_AFTER_MASTER_NODES ` all default to `2` whereas we set it to `1` in nuxeo helm chart: https://github.com/nuxeo/nuxeo-helm-chart/blob/master/nuxeo/values.yaml#L110
I think it makes sense to also declare these two variables as we're lowering the number of required master nodes, what do you think?

Jenkinsfile
ataillefer: why? where?

ci/helm/preview/values.yaml
ataillefer: Great! Yes, let's add them tot he nuxeo Helm chart.
NXP-28618 can be used for the commit message.

Jenkinsfile
NourNuxeo: https://jenkins.platform.dev.nuxeo.com/blue/organizations/jenkins/nuxeo%2Fnuxeo/detail/PR-3856/84/pipeline/

in the last shell script before last line:
```
[2020-03-27T20:56:08.654Z] error: no ORG environment variable found

script returned exit code 1
```
I didn't search extensively about why exactly but it is also present in the arender Jenkinsfile

Jenkinsfile
NourNuxeo: Yes, since it was in the TEMP commit I didn't mind. But it will be removed on the next push 👍

Jenkinsfile
NourNuxeo: Actually the preview often takes 10-11 minutes before being available:
https://jenkins.platform.dev.nuxeo.com/blue/organizations/jenkins/nuxeo%2Fnuxeo/detail/PR-3856/85/pipeline/

NourNuxeo:
Closing to reopen as draft

ci/helm/preview/values.yaml
NourNuxeo: Do I keep this patch in the preview chart then ? 
Wouldn't hurt to have it while waiting for the newer version of nuxeo helm chart.

ci/helm/preview/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+      periodSeconds: 20
+      successThreshold: 1
+      failureThreshold: 5
+      timeoutSeconds: 10
+    readinessProbe:
+      periodSeconds: 20
+      initialDelaySeconds: 600
+      successThreshold: 1
+      timeoutSeconds: 10
+      failureThreshold: 5
```
kevinleturc: This may be linked to my other comment about `preview-health-timeout`, these values are for an environment deploying nuxeo 10.10. On master we probably don't need to wait 10min before checking readiness.

I think we could try the default values coming from nuxeo helm chart.

ci/helm/preview/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+          nginx.ingress.kubernetes.io/server-snippet: ignore_invalid_headers off;
+      labels:
+        team: platform
+        resource: service
+        usage: preview
+    image:
+      repository: ${DOCKER_REGISTRY}/nuxeo/nuxeo
+      tag: ${VERSION}
+      pullPolicy: Always
+    packages: nuxeo-web-ui nuxeo-dam
```
kevinleturc: To remove.

ci/helm/preview/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+      team: platform
+      resource: pod
+      usage: preview
+    service:
+      annotations:
+        fabric8.io/ingress.annotations: |-
+          nginx.ingress.kubernetes.io/proxy-body-size: 0
+          nginx.ingress.kubernetes.io/server-snippet: ignore_invalid_headers off;
+      labels:
+        team: platform
```
kevinleturc: To add maybe?
```
branch: "$BRANCH_NAME"
```

ci/helm/preview/values.yaml
ataillefer: You can do a PR on the nuxeo chart and reference this PR version in the requirements.
Then when the chart is merged `version: ~1.0.0` should fetch the latest release.

Jenkinsfile
ataillefer: I think it's important to understand why it is needed, because reading the Jenkinsfile, defining an environment variable without any obvious reference will lose the reader.
Then, add a comment to explain why/where it is used.

ataillefer: It's indeed present in the ARender Jenkinsfile because it is actually used: https://github.com/nuxeo/nuxeo-arender-connector/blob/10/Jenkinsfile#L48.

Jenkinsfile
```diff
@@ -303,393 +307,61 @@ pipeline {
       }
     }
 
-    stage('Update version') {
+    stage('Deploy Preview') {
       steps {
+        setGitHubBuildStatus('nuxeo/preview/deploy', 'Deploy nuxeo Preview', 'PENDING')
```
ataillefer: Maybe just `nuxeo/preview`? Shorter in the status list.
I know it's like this in ARender, but WDYT @troger @kevinleturc ?

Jenkinsfile
```diff
@@ -279,11 +279,15 @@ pipeline {
...
     MAVEN_ARGS = getMavenArgs()
-    VERSION = getVersion()
+    VERSION = "11.1-SNAPSHOT"
     DOCKER_TAG = getDockerTagFrom("${VERSION}")
     CHANGE_BRANCH = "${env.CHANGE_BRANCH != null ? env.CHANGE_BRANCH : BRANCH_NAME}"
     CHANGE_TARGET = "${env.CHANGE_TARGET != null ? env.CHANGE_TARGET : BRANCH_NAME}"
     CONNECT_PREPROD_URL = 'https://nos-preprod-connect.nuxeocloud.com/nuxeo'
+    ORG = "nuxeo"
+    APP_NAME = "nuxeo"
+    NAMESPACE = "$APP_NAME-${BRANCH_NAME.toLowerCase()}"
```
ataillefer: I'd go with `PREVIEW_NAMESPACE` so it is explicit.
We're also using some namespaces for unit tests, for which the related env vars are prefixed by `TEST_`, see for instance https://github.com/nuxeo/nuxeo/blob/master/Jenkinsfile#L267.

Jenkinsfile
```diff
@@ -307,370 +307,7 @@ pipeline {
       }
     }
 
-    stage('Update version') {
```
ataillefer: For a smaller diff, you wan simply use
```
      when {
        expression {
          return false
        }
      }
```
See https://github.com/nuxeo/nuxeo/pull/3870/commits/2bd2ee91d9c2e922996797b683ea15c809b6f66f.

ci/helm/preview/values.yaml
ataillefer: yes, for consistency

Jenkinsfile
```diff
@@ -661,6 +665,65 @@ pipeline {
...
+                  --preview-health-timeout 15m \
+                  ${noCommentOpt}
+              """
+            }
+          }
+        }
+      }
+      post {
+        always {
+          archiveArtifacts allowEmptyArchive: true, artifacts: '**/requirements.lock, **/charts/*.tgz, **/target/**/*.yaml'
```
ataillefer: I'm wondering if we need/want the chart tgz archives, since we have all the YAML files?
@kevinleturc

ci/helm/preview/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+    helm.sh/hook-delete-policy: hook-succeeded
+
+nuxeo:
+  fullnameOverride: preview
+  nuxeo:
+    customParams: |-
+      elasticsearch.restClient.socketTimeoutMs=300000
+      elasticsearch.restClient.connectionTimeoutMs=300000
+    podAnnotations:
+      preview/scm.ref: ${SCM_REF}
```
ataillefer: Can't we always use `$VAR_NAME` (without `{...}}`) for environment variable substitution?

Jenkinsfile
```diff
@@ -661,6 +665,65 @@ pipeline {
...
+      }
+      steps {
+        setGitHubBuildStatus('nuxeo/preview/deploy', 'Deploy nuxeo Preview', 'PENDING')
+        container('maven') {
+          dir('ci/helm/preview') {
+            echo """
+            ----------------------------------------
+            Deploy Preview environment
+            ----------------------------------------"""
+            // first substitute docker image names and versions
```
ataillefer: The comment about Docker images is not relevant.
It should be limited to something generic like "first substitute environment variables in chart values".

Jenkinsfile
```diff
@@ -661,6 +665,65 @@ pipeline {
...
+            sh """
+              mv values.yaml values.yaml.tosubst
+              envsubst < values.yaml.tosubst > values.yaml
+            """
+            // second create target namespace (if doesn't exist) and copy secrets to target namespace
+            script {
+              sh(returnStdout: true, script: 'jx -b ns | sed -r "s/^Using namespace \'([^\']+)\'.+\\$/\\1/"')
```
ataillefer: Is this line really useful?

Jenkinsfile
```diff
@@ -661,6 +665,65 @@ pipeline {
...
+              boolean nsExist = sh(returnStatus: true, script: "kubectl get namespace ${NAMESPACE}") == 0
+              String noCommentOpt = '';
+              if (nsExist) {
+                noCommentOpt = '--no-comment'
+              } else {
+                sh "kubectl create namespace ${NAMESPACE}"
+              }
+              sh "kubectl create secret generic preview-docker-cfg --namespace=${NAMESPACE} --from-file=.dockerconfigjson=/home/jenkins/.docker/config.json --type=kubernetes.io/dockerconfigjson --dry-run -o yaml | kubectl apply -f -"
+              // third build and deploy the chart
+              // waiting for https://github.com/jenkins-x/jx/issues/5797 to be fixed in order to remove --source-url
```
ataillefer: Do we want to refresh the [issue](https://github.com/jenkins-x/jx/issues/5797#issuecomment-598564797) or add a comment?

Jenkinsfile
```diff
@@ -661,6 +665,65 @@ pipeline {
...
+              sh """
+                jx step helm build --verbose
```
ataillefer: Do we want to keep the verbose output?

Jenkinsfile
```diff
@@ -661,6 +665,65 @@ pipeline {
...
+                mkdir target && helm template . --output-dir target
+                jx preview \
+                  --namespace ${NAMESPACE} \
+                  --verbose \
```
ataillefer: Do we want to keep the verbose output?

Jenkinsfile
troger: Agree 👍

ci/helm/preview/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+
+cleanup:
+  Args:
+    - --cleanup
+  Annotations:
+    helm.sh/hook: pre-delete
+    helm.sh/hook-delete-policy: hook-succeeded
+
+nuxeo:
+  fullnameOverride: preview
```
ataillefer: What about `nuxeo-preview` to have it very clear in the pod list that it's a nuxeo preview (even if the pod are in a dedicated namespace)?

Jenkinsfile
troger: IMO, no we don't.

Jenkinsfile
kevinleturc: Yes sure 👍

ci/helm/preview/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+          nginx.ingress.kubernetes.io/proxy-body-size: 0
+          nginx.ingress.kubernetes.io/server-snippet: ignore_invalid_headers off;
+      labels:
+        team: platform
+        resource: service
+        usage: preview
+    image:
+      repository: ${DOCKER_REGISTRY}/nuxeo/nuxeo
+      tag: ${VERSION}
+      pullPolicy: Always
```
ataillefer: `VERSION` should refer to a release version, so always unique, thus `pullPolicy: Always` should be useles.

ci/helm/preview/values.yaml
NourNuxeo: I think we can because envsubst declares it takes both formats.

ci/helm/preview/values.yaml
NourNuxeo: I lowered --preview-health-timeout to 5m
Seems to be working fine.

Then I realized this is the default value so I'll remove the whole jx preview option on next push

Jenkinsfile
NourNuxeo: This is a jx preview requirement as well as APP_NAME:
https://github.com/jenkins-x/jx/blob/master/pkg/cmd/preview/preview.go#L1008-L1022

Jenkinsfile
NourNuxeo: About ORG and APP_NAME cf ttps://github.com/nuxeo/nuxeo/pull/3856#discussion_r400105125
About renaming NAMESPACE to TEST_NAMESPACE
There is a testNameSpace local variable in buildUnitTestStage.
it's distinguishable but yet buggling.

Jenkinsfile
NourNuxeo: Once things are stable we don't but I would leave it commented so we can uncomment when wanting to debug

Jenkinsfile
NourNuxeo: I did it inline since it's a temp modification that can't go unnoticed
```
     when {expression {return false}}
```

Jenkinsfile
ataillefer: https://github.com//pull/3856#discussion_r400105125 leads to a 404.
APP_NAME is used yes, just don't see ORG.

Yes, please rename NAMESPACE to PREVIEW_NAMESPACE so we know it' the one used for preview.
There are plenty of namespaces involved: the one the pod is running in, the default one, the test ones, ...

Jenkinsfile
ataillefer: OK if you think it can be useful, then clearly state in a comment that it can be uncommented for debugging purpose.
By the way, what  is in the archives that is not in the YAML archived files?

Jenkinsfile
ataillefer: OK, about ORG, the explanation is that `jx preview` is requiring it [here](https://github.com/jenkins-x/jx/blob/8fdc3a1182bc3ed6ae09728721b53ea4fa1d6234/pkg/cmd/preview/preview.go#L1007), though it isn't documented...
And there's a closed issue though it wasn't really resolved: https://github.com/jenkins-x/jx/issues/1992.

Jenkinsfile
NourNuxeo: Thanks!

Jenkinsfile
NourNuxeo: I woud suggest making a DEBUG env so we can control all the verbose and the making of the yaml archive in one go

Jenkinsfile
NourNuxeo: https://github.com/nuxeo/nuxeo/pull/3856#discussion_r400241148

ci/helm/preview/values.yaml
NourNuxeo: I propose preview-nuxeo to stay consistent with the names of all the other pods. 
And It would ease autocomplete if you switch from a pod to another (no need to erase "nuxeo-preview" but just "nuxeo" then type "elas..." for example

ci/helm/preview/values.yaml
NourNuxeo: since this is the default value of the nuxeo helm chart we need to override it with `IfNotPresent` or `Never`
I think `IfNotPresent` would be best

Jenkinsfile
NourNuxeo: doesn't work anyway, this is what is litterally printed actually by jx:
```
sh(returnStdout: true, script: 'jx -b ns | sed -r "s/^Using namespace \'([^\']+)\'.+\\$/\\1/"')
```

Jenkinsfile
```diff
@@ -284,8 +288,15 @@ pipeline {
...
+    // Required by the jx preview command: https://github.com/jenkins-x/jx/blob/8fdc3a1182bc3ed6ae09728721b53ea4fa1d6234/pkg/cmd/preview/preview.go#L1007
+    ORG = "nuxeo"
+    // Required by the jx preview command: https://github.com/jenkins-x/jx/blob/8fdc3a1182bc3ed6ae09728721b53ea4fa1d6234/pkg/cmd/preview/preview.go#L1015
+    APP_NAME = "nuxeo"
+    TEST_NAMESPACE = "$APP_NAME-${BRANCH_NAME.toLowerCase()}"
+    PERSISTENCE = !isPullRequest()
+    DEBUG_OPTION = getDebugOption()
   }
-
+
```
ataillefer: extra spaces

Jenkinsfile
```diff
@@ -632,6 +643,63 @@ pipeline {
...
     }
 
+    stage('Deploy Preview') {
+      when {
+        not {
+          branch 'PR-*'
+        }
+      }
+      steps {
+        setGitHubBuildStatus('nuxeo/preview/deploy', 'Deploy nuxeo Preview', 'PENDING')
```
ataillefer: You marked as resolved but kept `nuxeo/preview/deploy`, we're suggesting simply `nuxeo/preview`.

Jenkinsfile
ataillefer: Maybe because the result of the shell script isn't assigned to any variable.

Just for information, if it were assigned to a variable, it would return the name of the current namespace, you can test it locally without the escaping characters:
```
$ jx -b ns | sed -r "s/^Using namespace '([^\']+)'.+\$/\\1/"
platform
```

Jenkinsfile
ataillefer: I've applied the `/remove-lifecycle stale` to https://github.com/jenkins-x/jx/issues/5797#issuecomment-606466615.

ci/helm/preview/values.yaml
ataillefer: Yes indeed.

In, fact, while we're relying on the `latest` tag in the nuxeo Helm chart, we could remove `pullPolicy: Always` from it, see https://kubernetes.io/fr/docs/concepts/containers/images/#mettre-%C3%A0-jour-des-images.

But we might want to use something else than `latest` in the future, since it is not recommended.

NourNuxeo: Thanks for the extra infos!

Jenkinsfile
NourNuxeo: found some hand vs code settings to trim extra new lines and trailing spaces.

Jenkinsfile
NourNuxeo: Sorry about that. There must have been a manipulation error as I remember adjusting it.

Jenkinsfile
NourNuxeo: I removed the tgz. After all if someone wants to play with the yaml he will just have to helm dependency update on the downloaded yaml

Jenkinsfile
NourNuxeo: indeed `platform` in my case
I should have `echo`ed it

ci/helm/preview/values.yaml
ataillefer: For you information, as I said, instead of keeping the Helm chart changes here, you could have referenced the PR version of the nuxeo Helm chart built here: https://jenkins.platform.dev.nuxeo.com/job/nuxeo/job/nuxeo-helm-chart/view/change-requests/job/PR-8/3/console (look for `nuxeo-1.0.5-PR-8-3.tgz`).

In [requirements.yaml](https://github.com/nuxeo/nuxeo/blob/feature-NXP-27713-nuxeo-staging-preview/ci/helm/preview/requirements.yaml#L12), just use:
```
dependencies:
  - alias: nuxeo
    name: nuxeo
    version: 1.0.5-PR-8-3
```

```
```

ci/helm/preview/values.yaml
```diff
@@ -0,0 +1,77 @@
...
+          nginx.ingress.kubernetes.io/server-snippet: ignore_invalid_headers off;
+      labels:
+        branch: "$BRANCH_NAME"
+        team: platform
+        resource: service
+        usage: preview
+    image:
+      repository: $DOCKER_REGISTRY/nuxeo/nuxeo
+      tag: $VERSION
+      pullPolicy: IfNotPresent
```
ataillefer, line 47: This should go in the TEMP commit as it won't be needed once https://github.com/nuxeo/nuxeo-helm-chart/pull/8/files is merged and the nuxeo Helm chart is updated.

In fact, it would even not be needed in the TEMP commit by following https://github.com/nuxeo/nuxeo/pull/3856/commits/20d169b7cc06156e5161530bb6fa6609eed6860d#r401406736.

Jenkinsfile
```diff
@@ -632,6 +638,72 @@ pipeline {
...
     }
 
+    stage('Deploy Preview') {
+      when {
+        not {
+          branch 'PR-*'
+        }
+      }
+      steps {
+        setGitHubBuildStatus('nuxeo/preview', 'Deploy nuxeo Preview', 'PENDING')
```
ataillefer, line 663: could be "preview" (lowercase, not a title) :grinning:

Jenkinsfile
```diff
@@ -632,6 +638,72 @@ pipeline {
...
+            Deploy Preview environment
+            ----------------------------------------"""
+            // first substitute environment variables in chart values
+            sh """
+              mv values.yaml values.yaml.tosubst
+              envsubst < values.yaml.tosubst > values.yaml
+            """
+            // second create target namespace (if doesn't exist) and copy secrets to target namespace
+            script {
+              boolean nsExist = sh(returnStatus: true, script: "kubectl get namespace ${PREVIEW_NAMESPACE}") == 0
```
ataillefer, line 677:
```suggestion
              boolean nsExists = sh(returnStatus: true, script: "kubectl get namespace ${PREVIEW_NAMESPACE}") == 0
```

Jenkinsfile
```diff
@@ -632,6 +638,72 @@ pipeline {
...
+              String noCommentOpt = '';
+              if (nsExist) {
+                noCommentOpt = '--no-comment'
+              } else {
+                sh "kubectl create namespace ${PREVIEW_NAMESPACE}"
+              }
+              sh "kubectl create secret generic preview-docker-cfg --namespace=${PREVIEW_NAMESPACE} --from-file=.dockerconfigjson=/home/jenkins/.docker/config.json --type=kubernetes.io/dockerconfigjson --dry-run -o yaml | kubectl apply -f -"
+              // third build and deploy the chart
+              // reference branch have a different preview command. The -n name provided must be lowercase.
```
ataillefer, line 686: Could you be more explicit?
Explain that for the reference branches we cannot use `jx preview` otherwise the preview deployment would be garbage collected.

Jenkinsfile
```diff
@@ -632,6 +638,72 @@ pipeline {
...
+              previewCommand = "jx step helm install --namespace ${PREVIEW_NAMESPACE} -n ${APP_NAME}-${PREVIEW_NAMESPACE.toLowerCase()} ."
```
ataillefer, line 687: Out of curiosity, did you try to use `helm install` directly instead of `jx step helm install`?
It's worth giving it a try to rely as less as possible on the `jx` command, since we are already using a mix of pure `helm` and `jx` commands.

Jenkinsfile
```diff
@@ -632,6 +638,72 @@ pipeline {
...
+              if (isPullRequest()) {
+                // waiting for https://github.com/jenkins-x/jx/issues/5797 to be fixed in order to remove --source-url
+                previewCommand = "jx preview --namespace ${PREVIEW_NAMESPACE} --source-url=https://github.com/nuxeo/nuxeo ${noCommentOpt}"
+              }
+              sh """
+                jx step helm build
```
ataillefer, line 693: Same remark as https://github.com/nuxeo/nuxeo/pull/3856/commits/a56c1daccaf38394bed76fb5471085cf33646b1d#r401422462: would `helm build` work?

Jenkinsfile
```diff
@@ -632,6 +638,72 @@ pipeline {
...
+              if (nsExist) {
+                noCommentOpt = '--no-comment'
+              } else {
+                sh "kubectl create namespace ${PREVIEW_NAMESPACE}"
+              }
+              sh "kubectl create secret generic preview-docker-cfg --namespace=${PREVIEW_NAMESPACE} --from-file=.dockerconfigjson=/home/jenkins/.docker/config.json --type=kubernetes.io/dockerconfigjson --dry-run -o yaml | kubectl apply -f -"
+              // third build and deploy the chart
+              // reference branch have a different preview command. The -n name provided must be lowercase.
+              previewCommand = "jx step helm install --namespace ${PREVIEW_NAMESPACE} -n ${APP_NAME}-${PREVIEW_NAMESPACE.toLowerCase()} ."
+              if (isPullRequest()) {
```
ataillefer, line 688: Usually, we tend to use either a ternary:
```
def previewCommand = isPullRequest() ?
  "jx preview ..."
  : "jx step helm install..."
```
or the traditional form:
```
if (isPullRequest()) {
  ...
} else {
  ...
}
```
It avoids assigning the variable twice and is easier to read.

Jenkinsfile
```diff
@@ -632,6 +638,72 @@ pipeline {
...
+                previewCommand = "jx preview --namespace ${PREVIEW_NAMESPACE} --source-url=https://github.com/nuxeo/nuxeo ${noCommentOpt}"
+              }
+              sh """
+                jx step helm build
+                mkdir target && helm template . --output-dir target
+                ${previewCommand}
+              """
+              if (!isPullRequest()) {
+                // When not using jx preview, we need to expose the nuxeo url by hand
+                url = sh(returnStdout: true, script: "jx get urls -n nuxeo-pr-3856 | grep -oP https://.* | tr -d '\\n'")
```
ataillefer, line 699: I guess `nuxeo-pr-3856` should be `${PREVIEW_NAMESPACE}`?

ataillefer:
Do you really want to commit the .tgz, .lock and `code` files?

Jenkinsfile
NourNuxeo: Yes I tried it first for the reason you mention, not relying on jx when possible:
https://jenkins.platform.dev.nuxeo.com/blue/organizations/jenkins/nuxeo%2Fnuxeo/detail/PR-3856/52/pipeline/

Maybe it's easy to solve. I just tried jx step as a workaround as you suggested in case of and it worked and stuck with it.

Jenkinsfile
NourNuxeo: indeed I thought I fixed this.

Jenkinsfile
NourNuxeo: Maybe this would make the direct helm install work (cf https://github.com/nuxeo/nuxeo/pull/3856#discussion_r401431268), I'm gonna test it

Jenkinsfile
ataillefer: OK great, stick with `jx step helm install` then.

Jenkinsfile
ataillefer: Don't think so, but yes you can give it a try.

NourNuxeo: Indeed `helm build` doesn't exist jx stem helm build wraps some stuff https://github.com/jenkins-x/jx/blob/8fdc3a1182bc3ed6ae09728721b53ea4fa1d6234/pkg/cmd/step/helm/step_helm_build.go that would need further investigation to see how to translate it in helm commands.

I'm sticking with jx step helm install as you suggest https://github.com/nuxeo/nuxeo/pull/3856#discussion_r401443689

ci/helm/preview/values.yaml
kevinleturc: It is also not needed regarding: 
```yaml
preview/scm.ref: $SCM_REF
```
This will make a change in the chart for each push and thus re-depoy the nuxe pod.

Jenkinsfile
ataillefer: :+1: 
Yes, the equivalent is `helm package`.

Jenkinsfile
NourNuxeo: I went for the ternary to avoid declaring an empty string
```

              // third build and deploy the chart
              previewCommand = isPullRequest()?
                // When deploying a pull request, we use jx preview which gc the merged pull requests every 3h
                // waiting for https://github.com/jenkins-x/jx/issues/5797 to be fixed in order to remove --source-url
                "jx preview --namespace ${PREVIEW_NAMESPACE} --source-url=https://github.com/nuxeo/nuxeo ${noCommentOpt}"
                // To avoid jx gc cron job, reference branch previews are deployed by calling helm install directly.
                //The -n name provided must be lowercase.
                : "jx step helm install --namespace ${PREVIEW_NAMESPACE} -n ${APP_NAME}-${PREVIEW_NAMESPACE.toLowerCase()} ."

```

Jenkinsfile
NourNuxeo: yes cf https://github.com/nuxeo/nuxeo/pull/3856#discussion_r401473144

ci/helm/preview/values.yaml
NourNuxeo: removed the whole image value

kevinleturc: `repository` also? We may need it to point to docker registry.

ataillefer: I think we need repository and version!
Otherwise the default values in the nuxeo Helm chart are:
```
repository: docker.packages.nuxeo.com/nuxeo/nuxeo
tag: latest
```
And we want to pull the image previously built and pushed to the internal Docker registry.
Or I missed something :smile:.

NourNuxeo: Back in place!

ci/helm/preview/values.yaml
NourNuxeo: Done here https://jenkins.platform.dev.nuxeo.com/blue/organizations/jenkins/nuxeo%2Fnuxeo/detail/PR-3856/59/pipeline

https://jenkins.platform.dev.nuxeo.com/job/nuxeo/job/nuxeo/job/PR-3856/59/artifact/ci/helm/preview/requirements.lock

Jenkinsfile
```diff
@@ -284,6 +284,12 @@ pipeline {
     CHANGE_BRANCH = "${env.CHANGE_BRANCH != null ? env.CHANGE_BRANCH : BRANCH_NAME}"
     CHANGE_TARGET = "${env.CHANGE_TARGET != null ? env.CHANGE_TARGET : BRANCH_NAME}"
     CONNECT_PREPROD_URL = 'https://nos-preprod-connect.nuxeocloud.com/nuxeo'
+    // Required by the jx preview command: https://github.com/jenkins-x/jx/blob/8fdc3a1182bc3ed6ae09728721b53ea4fa1d6234/pkg/cmd/preview/preview.go#L1007
+    ORG = "nuxeo"
+    // Required by the jx preview command: https://github.com/jenkins-x/jx/blob/8fdc3a1182bc3ed6ae09728721b53ea4fa1d6234/pkg/cmd/preview/preview.go#L1015
+    APP_NAME = "nuxeo"
```
ataillefer, line 290: Since we're not using `jx preview` anymore, can we remove `ORG` and `APP_NAME`?

Jenkinsfile
```diff
@@ -284,6 +284,12 @@ pipeline {
...
+    PREVIEW_NAMESPACE = "$APP_NAME-${BRANCH_NAME.toLowerCase()}"
```
ataillefer, line 291: What do think of having `"nuxeo-preview-${BRANCH_NAME.toLowerCase()}"`?
So, once it's merged, we have `nuxeo-preview-master` as a namespace, clearer to identify than just `nuxeo-master` when looking at the long list of available namespaces in Kubernetes.

Jenkinsfile
```diff
@@ -632,6 +638,67 @@ pipeline {
...
+              if (nsExists) {
+                noCommentOpt = '--no-comment'
+                // Previous preview deployment needs to be scaled to 0 to be replaced correctly
+                sh "kubectl -n ${PREVIEW_NAMESPACE} scale deployment preview --replicas=0"
+              } else {
+                sh "kubectl create namespace ${PREVIEW_NAMESPACE}"
+              }
+              // third build and deploy the chart
+              // To avoid jx gc cron job, reference branch previews are deployed by calling jx step helm install
+              // The -n name provided must be lowercase
```
ataillefer, line 688: I think this comment would be more suitable just before the definition of `PREVIEW_NAMESPACE` where the lowercase is applied, or eventually before  `kubectl create namespace`, because it is really this instruction that requires lowercase.

Jenkinsfile
NourNuxeo: In the end you got those out of the way :p

Jenkinsfile
NourNuxeo: In the end you got those out of the way :p

Jenkinsfile
NourNuxeo: I put it at the declaration. This is the only place where I call toLowerCase() now so I think it belongs there to explain the call

Jenkinsfile
```diff
@@ -632,8 +648,71 @@ pipeline {
...
+            // second create target namespace (if doesn't exist) and copy secrets to target namespace
+            script {
+              boolean nsExists = sh(returnStatus: true, script: "kubectl get namespace ${PREVIEW_NAMESPACE}") == 0
+              String noCommentOpt = '';
+              if (nsExists) {
+                noCommentOpt = '--no-comment'
+                // Previous preview deployment needs to be scaled to 0 to be replaced correctly
+                sh "kubectl -n ${PREVIEW_NAMESPACE} scale deployment nuxeo-preview --replicas=0"
+              } else {
+                sh "kubectl create namespace ${PREVIEW_NAMESPACE}"
```
ataillefer, line 681: I believe that `jx step helm install` creates the namespace if it doesn't exist, so this might be useles..

Jenkinsfile
```diff
@@ -632,8 +648,71 @@ pipeline {
...
+              }
+              // third build and deploy the chart
+              // To avoid jx gc cron job, reference branch previews are deployed by calling jx step helm install
+              sh """
+                jx step helm build
+                mkdir target && helm template . --output-dir target
+                jx step helm install --namespace ${PREVIEW_NAMESPACE} -n ${PREVIEW_NAMESPACE} .
```
ataillefer, line 688: I would suggest to always use the `-` or the `--` syntax for options, at least in the same command.
So: `jx step helm install --namespace xxx --name yyy`.

Jenkinsfile
```diff
@@ -279,11 +279,14 @@ pipeline {
     PUBLIC_DOCKER_REGISTRY = 'docker.packages.nuxeo.com'
     MAVEN_OPTS = "$MAVEN_OPTS -Xms512m -Xmx3072m"
     MAVEN_ARGS = getMavenArgs()
-    VERSION = getVersion()
+    VERSION = "11.x"
     DOCKER_TAG = getDockerTagFrom("${VERSION}")
     CHANGE_BRANCH = "${env.CHANGE_BRANCH != null ? env.CHANGE_BRANCH : BRANCH_NAME}"
     CHANGE_TARGET = "${env.CHANGE_TARGET != null ? env.CHANGE_TARGET : BRANCH_NAME}"
     CONNECT_PREPROD_URL = 'https://nos-preprod-connect.nuxeocloud.com/nuxeo'
+    // jx step helm install's -n option requires the name provided to be lowercase
```
ataillefer, line 287: more precisely the `--name` and `--namespace` options

Jenkinsfile
NourNuxeo: indeed, so it did with my minikube

ci/helm/preview/templates/empty-docker-secret-for-kubernetes-replicator.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: nexus-docker-cfg
+  annotations:
+    replicator.v1.mittwald.de/replicate-from: platform-staging/nexus-docker-cfg
```
ataillefer, line 6:
```suggestion
    replicator.v1.mittwald.de/replicate-from: platform/nexus-docker-cfg
```

Jenkinsfile
```diff
@@ -632,6 +635,63 @@ pipeline {
...
+            Deploy Preview environment
+            ----------------------------------------"""
+            // first substitute environment variables in chart values
+            sh """
+              mv values.yaml values.yaml.tosubst
+              envsubst < values.yaml.tosubst > values.yaml
+            """
+            script {
+              boolean nsExists = sh(returnStatus: true, script: "kubectl get namespace ${PREVIEW_NAMESPACE}") == 0
+              String noCommentOpt = '';
```
kevinleturc, line 659: `noCommentOpt` is unused?

ci/helm/preview/values.yaml
```diff
@@ -0,0 +1,71 @@
...
+    helm.sh/hook-delete-policy: hook-succeeded
+
+nuxeo:
+  fullnameOverride: nuxeo-preview
+  nuxeo:
+    customParams: |-
+      elasticsearch.restClient.socketTimeoutMs=300000
+      elasticsearch.restClient.connectionTimeoutMs=300000
+    podAnnotations:
+      preview/scm.ref: $SCM_REF
```
kevinleturc, line 28: unused?

Jenkinsfile
NourNuxeo: indeed no need without jx preview

PR #3857, opened 2020-03-24
NXP-28790: retention management configs

nmpcunha:


troger:
Why not using a `nuxeo.conf` property and/or a template in the existing Nuxeo Retention Package to switch between both retention modes?
I'm not a fan of adding a new Maven module and a new Package for only one XML contribution.

efge:
After discussion, I agree, let's go back to a simple template containing a .xml.nxftl with a conditional section based on a nuxeo.conf property to activate compliance mode. It will be simpler.

packages/nuxeo-retention-package/src/main/resources/install/templates/retention/nxserver/config/retention-compliance-config.xml.nxftl
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+
+<component name="org.nuxeo.retention.compliance.types">
+
+    <require>org.nuxeo.retention.types</require>
```
troger: Why not in the `if`? If the mode is not `compliance`, there is no need for this component to require something as it will be empty.

packages/nuxeo-retention-package/src/main/resources/install/templates/retention/nxserver/config/retention-compliance-config.xml.nxftl
```diff
@@ -0,0 +1,23 @@
...
+    <extension target="org.nuxeo.ecm.core.schema.TypeService" point="doctype">
+        <facet name="Commentable" enabled="false" />
+        <facet name="Publishable" enabled="false" />
+        <facet name="Versionable" enabled="false" />
+    </extension>
+
+</#if>
+
+</component>
+
```
troger: Empty line to be removed.

packages/nuxeo-retention-package/src/main/resources/install/templates/retention/nxserver/config/retention-compliance-config.xml.nxftl
nmpcunha: My rational here was to avoid this contribution to be loaded before `retention.types` since I wasn't able to understand how this engine works and if there is any chance of that to happen.

packages/nuxeo-retention-package/src/main/resources/install/templates/retention/nuxeo.defaults
```diff
@@ -0,0 +1,5 @@
+## DO NOT EDIT THIS FILE, USE nuxeo.conf  ##
+nuxeo.template.includes=
+retention.target=.
+
+nuxeo.retention.mode=standard
```
troger: Will we have more than 2 modes? If no, I would probably go for a "boolean".

nmpcunha: Not for now.
As far as I was able to understand, we can have "booleans" but at the end of the day they are strings, in fact. This way it will map directly what is on our requirements, but I would also be ok with `nuxeo.retention.compliant` with "true" or "false"

RSalem07: needs a break line at the end of the file.

packages/nuxeo-retention-package/src/main/resources/install/templates/retention/nxserver/config/retention-compliance-config.xml.nxftl
nmpcunha: It is not possible to have that scenario? If not I'll move that line 🙂

troger: You're right, it's better to have the `require`, but it's only needed when the component is not empty, so you can move that require with the other one, inside the `if`. If the component is empty, we don't care if it's loaded before or after `org.nuxeo.retention.types`.

packages/nuxeo-retention-package/src/main/resources/install/templates/retention/nuxeo.defaults
troger: I find it easier to understand/read if it's just "enable or not the compliance mode" :)
Let's add people to the conversation! WDYT @efge @RSalem07 ?

RSalem07: from my point of view, as we have only the compliance mode enable or disabled (standard) we can use boolean, which also avoid string comparaison any any typo error

efge: I'd have chosen `nuxeo.retention.compliance.enabled=true/false`

packages/nuxeo-retention-package/src/main/resources/install/templates/retention/nxserver/config/retention-compliance-config.xml.nxftl
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
+
+<component name="org.nuxeo.retention.compliance.types">
```
efge: Please remove the blank line before `<component` as well.

packages/nuxeo-retention-package/src/main/resources/install/templates/retention/nxserver/config/retention-compliance-config.xml.nxftl
```diff
@@ -0,0 +1,19 @@
...
+<#if "${nuxeo.retention.compliance.enabled}" == "true">
+    <require>org.nuxeo.retention.types</require>
+    <require>org.nuxeo.ecm.core.schema.TypeService</require>
```
efge: This `<require>` is useless, it's implicit in the fact that we use extension points from this component.

nmpcunha: What about loading order as we discussed [above](https://github.com/nuxeo/nuxeo/pull/3857#discussion_r397848731)? Is there any chance that this specific contribution is loaded before `org.nuxeo.retention.types`?

packages/nuxeo-retention-package/src/main/resources/install/templates/retention/nxserver/config/retention-compliance-config.xml.nxftl
nmpcunha: Can remove 👍 
But, just to clarify, it is consistent with the examples provided 😉

packages/nuxeo-retention-package/src/main/resources/install/templates/retention/nxserver/config/retention-compliance-config.xml.nxftl
nelsonsilva: point it to keep `<require>org.nuxeo.retention.types</require>` but drop the `<require>org.nuxeo.ecm.core.schema.TypeService</require>`  since you're already making contributions to it

PR #3862, opened 2020-03-25
NXP-26771: POST "token" returns plain text for JSON response

dbnuxeo:
The AuthenticationTokensObject defines the return type as 'application/json'.

However, the createToken POST endpoint returns an unescaped string (a UUID, it is not valid JSON).

Unable to use the endpoint from web applications because the parsing on the response is broken.

The response sends valid JSON (add quotes to the token) when the Accept header conforms to 'application/json' or is a wildcard.  If text/plain is provided, the value will not be quoted.

Add unit tests.

### DIFF
--- a/nuxeo-services/login/nuxeo-platform-login-token-rest/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/login/tokenauth/AuthenticationTokensObject.java
+++ b/nuxeo-services/login/nuxeo-platform-login-token-rest/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/login/tokenauth/AuthenticationTokensObject.java
@@ -18,18 +18,16 @@
  */
 package org.nuxeo.ecm.restapi.server.jaxrs.login.tokenauth;
 
-import org.nuxeo.ecm.core.api.DocumentModel;
-import org.nuxeo.ecm.core.api.DocumentModelList;
-import org.nuxeo.ecm.core.api.NuxeoPrincipal;
-import org.nuxeo.ecm.tokenauth.io.AuthenticationToken;
-import org.nuxeo.ecm.webengine.model.WebObject;
-import org.nuxeo.ecm.webengine.model.impl.AbstractResource;
-import org.nuxeo.ecm.webengine.model.impl.ResourceTypeImpl;
-import org.nuxeo.ecm.tokenauth.service.TokenAuthenticationService;
-import org.nuxeo.runtime.api.Framework;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
+import javax.ws.rs.HeaderParam;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
@@ -37,10 +35,17 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
-import java.util.Calendar;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
+
+import org.apache.commons.lang3.StringUtils;
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.DocumentModelList;
+import org.nuxeo.ecm.core.api.NuxeoPrincipal;
+import org.nuxeo.ecm.tokenauth.io.AuthenticationToken;
+import org.nuxeo.ecm.tokenauth.service.TokenAuthenticationService;
+import org.nuxeo.ecm.webengine.model.WebObject;
+import org.nuxeo.ecm.webengine.model.impl.AbstractResource;
+import org.nuxeo.ecm.webengine.model.impl.ResourceTypeImpl;
+import org.nuxeo.runtime.api.Framework;
 
 /**
  * Token Object
@@ -65,12 +70,20 @@ public class AuthenticationTokensObject extends AbstractResource<ResourceTypeImp
     }
 
     @POST
+    @Produces({ MediaType.APPLICATION_JSON, MediaType.TEXT_PLAIN })
     public Response createToken(@QueryParam("application") String applicationName,
             @QueryParam("deviceId") String deviceId, @QueryParam("deviceDescription") String deviceDescription,
-            @QueryParam("permission") String permission) {
+            @QueryParam("permission") String permission, @HeaderParam("accept") String acceptType) {
         String username = getCurrentUser().getName();
         String token = service.acquireToken(username, applicationName, deviceId, deviceDescription, permission);
-        return Response.ok(token).status(Response.Status.CREATED).build();
+        String body = token;
+        MediaType cType = MediaType.TEXT_PLAIN_TYPE;
+        List<MediaType> acceptList = accepts(acceptType);
+        if (acceptList.isEmpty() || acceptList.stream().anyMatch(type -> MediaType.APPLICATION_JSON_TYPE.isCompatible(type))) {
+           body = String.format("\"%s\"", token);
+           cType = MediaType.APPLICATION_JSON_TYPE;
+        }
+        return Response.ok(body).type(cType).status(Response.Status.CREATED).build();
     }
 
     @DELETE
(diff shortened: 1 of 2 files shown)


efge:
Originally was #2939

efge:
I'm surprised to see we return `"thetoken"` and not something like `{"token": "thetoken"}` which would be a much more flexible format (and would allow adding additional info later, like maybe an expiration date, etc.)

nuxeo-services/login/nuxeo-platform-login-token-rest/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/login/tokenauth/AuthenticationTokensObject.java
```diff
@@ -65,12 +70,20 @@ protected void initialize(Object... args) {
...
             @QueryParam("deviceId") String deviceId, @QueryParam("deviceDescription") String deviceDescription,
-            @QueryParam("permission") String permission) {
+            @QueryParam("permission") String permission, @HeaderParam("accept") String acceptType) {
         String username = getCurrentUser().getName();
         String token = service.acquireToken(username, applicationName, deviceId, deviceDescription, permission);
-        return Response.ok(token).status(Response.Status.CREATED).build();
+        String body = token;
+        MediaType cType = MediaType.TEXT_PLAIN_TYPE;
+        List<MediaType> acceptList = accepts(acceptType);
+        if (acceptList.isEmpty() || acceptList.stream().anyMatch(type -> MediaType.APPLICATION_JSON_TYPE.isCompatible(type))) {
```
efge: Instead of doing manual parsing/matching of the `Accept:` header, wouldn't it be simpler to have two methods, one with `@Produces(MediaType.APPLICATION_JSON)` and one with `@Produces(MediaType.TEXT_PLAIN)` to let the JAX-RS framework itself do the matching?

nuxeo-services/login/nuxeo-platform-login-token-rest/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/login/tokenauth/AuthenticationTokensObject.java
```diff
@@ -18,29 +18,34 @@
...
-import org.nuxeo.ecm.core.api.DocumentModel;
-import org.nuxeo.ecm.core.api.DocumentModelList;
-import org.nuxeo.ecm.core.api.NuxeoPrincipal;
-import org.nuxeo.ecm.tokenauth.io.AuthenticationToken;
-import org.nuxeo.ecm.webengine.model.WebObject;
-import org.nuxeo.ecm.webengine.model.impl.AbstractResource;
-import org.nuxeo.ecm.webengine.model.impl.ResourceTypeImpl;
-import org.nuxeo.ecm.tokenauth.service.TokenAuthenticationService;
-import org.nuxeo.runtime.api.Framework;
+import java.util.Calendar;
```
efge: Please avoid reorganizing imports in the same commit as the features/fixes being implemented themselves.

dbnuxeo:
I think an acceptable alternative could be to return the value from `asAuthenticationToken` from the JSON context.  The corresponding nuxeo npm javascript library expects a simple string value from the JSON content type.

nuxeo-services/login/nuxeo-platform-login-token-rest/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/login/tokenauth/AuthenticationTokensObject.java
dbnuxeo: Delegated matching to JAX-RS and added "detail" query parameter to return populated JSON object.

nuxeo-services/login/nuxeo-platform-login-token-rest/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/login/tokenauth/AuthenticationTokensObject.java
dbnuxeo: Restored order.

nuxeo-services/login/nuxeo-platform-login-token-rest/src/test/java/org/nuxeo/ecm/restapi/test/AuthenticationTokensTest.java
```diff
@@ -106,16 +107,79 @@ public void itCanRevokeTokens() throws Exception {
...
+        params.put("application", Collections.singletonList("app"));
+        params.put("deviceId", Collections.singletonList("device"));
+        params.put("permission", Collections.singletonList("rw"));
+        params.put("detail", Collections.singletonList("true"));
+        try (CloseableClientResponse response = getResponse(RequestType.POST, "/token", null, params, null, null,
+                new String[] { MediaType.APPLICATION_JSON })) {
+            assertEquals(Response.Status.CREATED.getStatusCode(), response.getStatus());
+            JsonNode respToken = mapper.readTree(response.getEntityInputStream());
+            assertTrue(MediaType.APPLICATION_JSON_TYPE.isCompatible(response.getType()));
+            assertEquals("app", respToken.get("application").textValue());
```
efge: Shouldn't we check `respToken.get("token")` too? It's the main info after all

dbnuxeo: Done

efge: I don't see a change. Did you push on the correct repo?

dbnuxeo: Yes.  The JSON serializes the token value as "id" and not "token".

efge: Ah ok yes `AuthenticationTokenWriter` does that work and uses a different name.

nuxeo-services/login/nuxeo-platform-login-token-rest/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/login/tokenauth/AuthenticationTokensObject.java
```diff
@@ -65,12 +65,32 @@ protected void initialize(Object... args) {
...
+        String username = getCurrentUser().getName();
+        String token = service.acquireToken(username, applicationName, deviceId, deviceDescription, permission);
+        Object body = String.format("\"%s\"", token);
+        if (detail) {
+            // Get the token from the service to populate additional fields like 'creationDate'
+            AuthenticationToken at = getTokens(applicationName).stream().filter(item -> token.equals(item.getToken()))
+                    .findFirst().orElse(null);
+            if (at != null) {
+            	body = at;
+            }
```
kevinleturc: This could be reworked to:
```java
body = getTokens(applicationName).stream()
                                 .filter(item -> token.equals(item.getToken()))
                                 .findFirst()
                                 .orElse(body);
```

dbnuxeo: It could, but the receiver type would have to be Object.  I kept it this way for type readability.

kevinleturc: It could be interesting if we're using the type, this is not the case here. The information it brings can be given by recent IDE on findFirst() line.

For readability too, the current has an extra null check making code strange to read. 
Also, it is not formatted, which will end to something less readable in my opinion:
```java
AuthenticationToken at = getTokens(applicationName).stream()
                                                   .filter(item -> token.equals(item.getToken()))
                                                   .findFirst()
                                                   .orElse(null);
if (at != null) {
    body = at;
}
```

dbnuxeo: Resolved this issue, had to erase the type signature with a map call.

modules/platform/login/nuxeo-platform-login-token-rest/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/login/tokenauth/AuthenticationTokensObject.java
```diff
@@ -65,12 +65,33 @@ protected void initialize(Object... args) {
...
+            @QueryParam("permission") String permission, @QueryParam("detail") boolean detail) {
+        String username = getCurrentUser().getName();
+        String token = service.acquireToken(username, applicationName, deviceId, deviceDescription, permission);
+        Object body = String.format("\"%s\"", token);
+        if (detail) {
+            // Get the token from the service to populate additional fields like 'creationDate'
+            body = getTokens(applicationName).stream()
+                                             .filter(item -> token.equals(item.getToken()))
+                                             .findFirst()
+                                             // Erase AuthenticationToken type signature
```
efge: The comment is a bit distracting when reading the code, especially since it describes what the code does and not why. I'd simply remove it, people interested in understanding why can see in the IDE the error reported if they remove the cast.

dbnuxeo: Removed the comment.

PR #3867, opened 2020-03-25
fix-NXP-28710-NXP-28818-failing-multi-environment-unit-tests

ataillefer:


ataillefer:
For test purpose only, closed in favor of #3870.

PR #3891, opened 2020-04-06
WIP - task-NXP-28619-unit-tests-kafka

ataillefer:


Jenkinsfile
```diff
@@ -369,28 +396,81 @@ pipeline {
...
+                kubectl rollout status statefulset ${TEST_REDIS_RESOURCE} \
+                  --namespace=${testNamespace} \
+                  --timeout=${TEST_DEFAULT_ROLLOUT_STATUS_TIMEOUT}
+              """
+              // wait for Kafka to be ready
+              sh """
+                kubectl rollout status statefulset ${TEST_KAFKA_RESOURCE} \
+                  --namespace=${testNamespace} \
+                  --timeout=${TEST_LONG_ROLLOUT_STATUS_TIMEOUT}
+              """
```
kevinleturc, line 436: Could it be interesting to have a method for the rollouts? something like waitFor(ns, type, name, timeout)? Or without type as we know all kinds of resources and can do a switch in the method, wdyt?

modules/runtime/nuxeo-runtime-stream/src/test/java/org/nuxeo/runtime/stream/RuntimeStreamFeature.java
```diff
@@ -69,6 +74,17 @@
...
     protected String streamType;
 
+    public static void assumeKafkaEnabled() {
+        if ("true".equals(System.getProperty("kafka"))// deprecated since 10.3
+                || STREAM_KAFKA.equals(System.getProperty(STREAM_PROPERTY))) {
+            if (!KafkaUtils.kafkaDetected()) {
+                fail("Kafka is enabled, but no Kafka server found.");
+            }
+        } else {
+            Assume.assumeTrue("Kafka not enabled", false);
```
kevinleturc, line 84: It could be imported statically.

Jenkinsfile
```diff
@@ -369,28 +396,81 @@ pipeline {
...
+
+              echo "runtime unit tests: run Maven"
+              // run unit tests
+              sh """
+                cd modules/runtime
```
troger, line 441: Why replacing `dir('modules/runtime')` by that?

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/event/test/WorkTest.java
```diff
@@ -60,6 +63,16 @@
     @Inject
     protected CoreSession session;
 
+    @Inject
+    protected WorkManagerFeature workManagerFeature;
+
+    // TODO NXP-28931: remove
+    @Before
```
troger, line 70:
```suggestion
    @BeforeClass
```
?

Jenkinsfile
ataillefer: No specific reason, just to avoid one level of indentation and keep the shell scripts aligned.

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/event/test/WorkTest.java
ataillefer: I can't because I'm using the non static member `workManagerFeature`, if I'm not mistaken.

troger: Ha... right.

Jenkinsfile
troger: Well, we should probably stick to one or the other but not both, WDYT?

ataillefer:
T&P OK: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-ataillefer/208/

ataillefer:
Please don't review, will be incuded by https://github.com/nuxeo/nuxeo/pull/3930.

ataillefer:
Superseded by #3930.

PR #3900, opened 2020-04-08
NXP-28900: allow blob provider to access document when reading blob

efge:
T&P ok https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/801/

### DIFF
--- a/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobManager.java
+++ b/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobManager.java
@@ -95,22 +95,28 @@ public interface BlobManager {
     BlobProvider getBlobProvider(Blob blob);
 
     /**
-     * INTERNAL - Gets an {@link InputStream} for the data of a managed blob. Used by internal implementations, regular
-     * callers should call {@link Blob#getStream}.
+     * Gets an {@link InputStream} for the data of a managed blob.
+     * <p>
+     * If the blob is managed this is equivalent to {@link ManagedBlob#getStream()}, otherwise returns {@code null}.
      *
      * @param blob the blob
-     * @return the stream
+     * @return the stream, or {@code null} if the blob is not managed
+     * @deprecated since 11.1, use {@link Blob#getStream()} instead
      */
+    @Deprecated
     InputStream getStream(Blob blob) throws IOException;
 
     /**
-     * INTERNAL - Gets a {@link File} (if one exists) for the data of a managed blob. Used by internal implementations,
-     * regular callers should call {@link Blob#getFile}.
+     * Gets a {@link File} (if one exists) for the data of a managed blob.
+     * <p>
+     * If the blob is managed this is equivalent to {@link ManagedBlob#getFile()}, otherwise returns {@code null}.
      *
      * @param blob the blob
-     * @return the file, or {@code null} if no underlying file is available
+     * @return the file, or {@code null} if no underlying file is available or the blob is not managed
      * @since 11.1
+     * @deprecated since 11.1, use {@link Blob#getFile()} instead
      */
+    @Deprecated
     File getFile(Blob blob);
 
     /**
--- a/modules/core/nuxeo-core-api/src/test/resources/OSGI-INF/dummy-blob-provider.xml
+++ b/modules/core/nuxeo-core-api/src/test/resources/OSGI-INF/dummy-blob-provider.xml
@@ -15,5 +15,12 @@
     <blobprovider name="fromdoc">
       <class>org.nuxeo.ecm.core.blob.DummyBlobProvider</class>
     </blobprovider>
+    <blobprovider name="fromdoc2">
+      <class>org.nuxeo.ecm.core.blob.DummyBlobProvider</class>
+    </blobprovider>
+    <blobprovider name="memwithbyterange">
+      <class>org.nuxeo.ecm.core.blob.InMemoryBlobProvider</class>
+      <property name="allowByteRange">true</property>
+    </blobprovider>
   </extension>
 </component>
--- a/modules/core/nuxeo-core-api/src/test/resources/OSGI-INF/dummy-blob-provider.xml
+++ b/modules/core/nuxeo-core-api/src/test/resources/OSGI-INF/dummy-blob-provider.xml
@@ -12,5 +12,8 @@
     <blobprovider name="dummy">
       <class>org.nuxeo.ecm.core.blob.DummyBlobProvider</class>
     </blobprovider>
+    <blobprovider name="fromdoc">
+      <class>org.nuxeo.ecm.core.blob.DummyBlobProvider</class>
+    </blobprovider>
   </extension>
 </component>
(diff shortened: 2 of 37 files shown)


efge:
Note that https://github.com/nuxeo/nuxeo-jsf-ui/pull/49 has a few related changes

modules/core/nuxeo-core-api/src/test/resources/OSGI-INF/dummy-blob-provider.xml
```diff
@@ -12,5 +12,15 @@
     <blobprovider name="dummy">
       <class>org.nuxeo.ecm.core.blob.DummyBlobProvider</class>
     </blobprovider>
+    <blobprovider name="fromdoc">
+      <class>org.nuxeo.ecm.core.blob.DummyBlobProvider</class>
+    </blobprovider>
+    <blobprovider name="fromdoc2">
+      <class>org.nuxeo.ecm.core.blob.DummyBlobProvider</class>
```
kevinleturc, line 19: Should we configure it as allowing byte range?

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
```diff
@@ -367,15 +373,23 @@ protected void clearBucket() {
...
+        }
         String objectKey;
         String versionId;
-        int seppos = key.indexOf(VER_SEP);
-        if (seppos < 0) {
-            objectKey = key;
-            versionId = null;
-        } else {
+        int seppos;
+        if (useVersion && (seppos = key.indexOf(VER_SEP)) > 0) {
```
kevinleturc, line 387: What do you think about introducing an object for key which does all the needed parsing (byte range and version) and holds methods such as `containsRange`.

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/SessionImpl.java
```diff
@@ -1491,26 +1493,32 @@ public boolean isFulltextStoredInBlob() {
...
-                BlobInfo blobInfo = new BlobInfo();
-                blobInfo.key = fulltext;
-                Blob blob = blobManager.readBlob(blobInfo, getRepositoryName());
-                fulltext = blob.getString();
-                map.put(BINARY_FULLTEXT_MAIN_KEY, fulltext);
-            } catch (IOException e) {
-                throw new PropertyException("Cannot read fulltext blob for doc: " + id, e);
+            if (doc == null) {
+                // could not find doc (shouldn't happen)
+                fulltext = null;
```
kevinleturc, line 1506: It can happen if fulltext is not stored in blob, no?

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManager.java
```diff
@@ -35,14 +35,27 @@
...
+    /**
+     * Reads a {@link Blob} from storage.
+     *
+     * @param blobInfo the blob information
+     * @param doc the document to which this blob belongs
+     * @param xpath the xpath of the blob in the document
+     * @return a managed blob
+     * @since 11.1
+     */
+    Blob readBlob(BlobInfo blobInfo, Document doc, String xpath) throws IOException;
```
kevinleturc, line 47: Why not using BlobInfoContext for the signature?

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobManager.java
```diff
@@ -95,22 +95,28 @@ default BlobProvider getBlobProviderWithNamespace(String id) {
...
-     * INTERNAL - Gets an {@link InputStream} for the data of a managed blob. Used by internal implementations, regular
-     * callers should call {@link Blob#getStream}.
+     * Gets an {@link InputStream} for the data of a managed blob.
+     * <p>
+     * If the blob is managed this is equivalent to {@link ManagedBlob#getStream()}, otherwise returns {@code null}.
      *
      * @param blob the blob
-     * @return the stream
+     * @return the stream, or {@code null} if the blob is not managed
+     * @deprecated since 11.1
```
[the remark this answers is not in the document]
efge: The Javadoc says `If the blob is managed this is equivalent to {@link ManagedBlob#getStream()}, otherwise returns {@code null}` so I think it's enough?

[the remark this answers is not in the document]
efge: But I'll add something anyway ok.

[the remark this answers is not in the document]
efge: Doing `Blob#getStream` directly is the replacement. I'll add details.

modules/core/nuxeo-core-api/src/test/resources/OSGI-INF/dummy-blob-provider.xml
efge: No this one is just a "virtual" blob provider, it doesn't need byte ranges. It's the one the virtual implementation is actually using that needs byte ranges.

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
efge: You mean a helper? In the vast majority of cases the key is just a string without structure. It's only for S3 versions and byte ranges that we need additional semantics but I don't want to overhaul the whole BlobStore API to add an abstraction here, which is why I did this string parsing.

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/SessionImpl.java
efge: No, there's a `if (fulltextDescriptor.getFulltextStoredInBlob()` just before.

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManager.java
efge: For the `DocumentBlobManager` I kept simple explicit APIs. `BlobInfoContext` is currently used at lower level in `BlobProvider` to pass info around, but here I kept something symmetric with the existing `writeBlob(Blob blob, Document doc, String xpath)`. We could refactor this to use `BlobInfoContext` / `BlobContext` yes (although `BlobContext` doesn't currently keep the full doc, just its id). I'll keep that for later.

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
kevinleturc: I meant a helper, more a data object with convenient methods and a static parse method (for instance). We would use only here and where we need to read range and not to use in a method signature as the change would be very big and not relevant in all cases (as you said).

`VER_SEP` could be a blocker for such helper as it is only for S3 and we don't want it in a "generic" thing.

efge:
Updated with details for deprecation in Javadoc, and better details in byte range parsing debug log.

PR #3905, opened 2020-04-09
NXP-28903: allow data uri for images

nelsonsilva:
Data uris are not part of default allowed protocols. This makes them allowed and assumes the policy defines regexes for attribute values (which is the case for us with the legacy antisamy policy in place).


troger:
Missing T&P? 😃

nelsonsilva:
🔵 https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-nsilva/133/

PR #3906, opened 2020-04-09
NXP-27713: explicitly write nuxeo helm chart overriden nuxeo.conf values

NourNuxeo:
The overriden nuxeo.conf are lost, so they need to be written explicitly:
In our case: 
elasticsearch.indexNumberOfReplicas=0


ataillefer:
See https://github.com/nuxeo/nuxeo-helm-chart/pull/8/ for including
```
elasticsearch.restClient.socketTimeoutMs
elasticsearch.restClient.connectionTimeoutMs
```
directly in the configmap template.

PR #3914, opened 2020-04-13
NXP-26872: Overall ZIP CSV Improvements to improve functionality and align with CSV import

eikkaj:
-- name field should be required
-- id field should not be used
-- update should be supported
-- tags should be supported
-- list type should be supported
-- combined archive files per request of nuxeo


modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/CSVZipImporter.java
```diff
@@ -111,47 +118,46 @@ public DocumentModel createOrUpdate(FileImporterContext context) throws IOExcept
...
-                            id = lineValue;
+                        } else if ("name".equalsIgnoreCase(headerValue)) {
+                            name = lineValue;
                         } else {
                             stringValues.put(headerValue, lineValue);
                         }
                     }
 
+                    if (StringUtils.isBlank(name)) {
+                        log.error("Can not create or update doc without a name, skipping line: {}", name);
```
efge, line 140: I would use a `log.warn` here, ERROR seems a bit strong for something about which the server admins can't do anything. Also logging the name is useless as it's blank, maybe log `csvRecord.getRecordNumber()` or `csvRecord.getCharacterPosition()` instead?

eikkaj: sorry I missed that @efge , I made an update for this

modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/CSVZipImporter.java
```diff
@@ -111,47 +118,47 @@ public DocumentModel createOrUpdate(FileImporterContext context) throws IOExcept
...
+                        } else if ("name".equalsIgnoreCase(headerValue)) {
+                            name = lineValue;
                         } else {
                             stringValues.put(headerValue, lineValue);
                         }
                     }
 
+                    if (StringUtils.isBlank(name)) {
+                        log.warn("Can not create or update doc without a name, skipping line: {}",
+                                csvRecord.getRecordNumber());
```
efge, line 141: Not a big deal, as WARN messages will rarely be hidden, and this is not a performance hot spot, but (as some check tools warn about) the better syntax here is to use `csvRecord::getRecordNumber` to delay evaluation until the message is actually required.

modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/CSVZipImporter.java
```diff
@@ -111,47 +118,47 @@ public DocumentModel createOrUpdate(FileImporterContext context) throws IOExcept
...
+                    if (session.exists(new PathRef(targetPath))) {
+                        targetDoc = session.getDocument(new PathRef(targetPath));
+                        updateDoc = true;
                     }
 
                     // create doc if needed
                     if (targetDoc == null) {
                         if (type == null) {
-                            log.error("Can not create doc without a type, skipping line");
+                            log.error("Can not create doc without a type, skipping line: {}", name);
```
efge, line 158: I had to double check because the other "skipping line" message above logs the line number, whereas here we log the line "name" but this is a good enough identifier. Although I guess it wouldn't hurt to also add the line number.

modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/CSVZipImporter.java
```diff
@@ -241,9 +254,16 @@ protected Serializable getFieldValue(Field field, String stringValue, ZipFile zi
...
+        } else if (type.isComplexType()) {
+            boolean returnEarly = true;
+            if (TypeConstants.CONTENT.equals(field.getName().getLocalName())) {
+                returnEarly = false;
+            }
             ZipEntry blobIndex = zip.getEntry(stringValue);
-            if (blobIndex != null) {
+            if (!returnEarly) {
+                returnEarly = blobIndex == null;
+            }
```
efge, line 265: You can just replace these 3 lines with `returnEarly |= blobIndex == null`

modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/CSVZipImporter.java
eikkaj: Cool, I had not thought of this. I made this update now, thanks for pointing it out.

modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/CSVZipImporter.java
eikkaj: I replaced with line number for simplicity and also changed to warn instead of error

modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/CSVZipImporter.java
eikkaj: Nice :) Updated

PR #3916, opened 2020-04-13
NXP-28921: check for null and undefined on search docs properties

mcedica:


addons/nuxeo-platform-spreadsheet/nuxeo-platform-spreadsheet-web/src/main/js/app/app.js
```diff
@@ -142,6 +142,9 @@ function run(baseURL = '/nuxeo', username = null, password = null) {
       for (var k in cv.searchDocument.properties) {
         var v = cv.searchDocument.properties[k];
         // skip empty values
+        if ( v == null || v === undefined ) {
```
nmpcunha, line 145: The comparison can be simply `v == null` or `v == undefined`.
When using `==` and not the strict comparison `===`, it checks both `null` and `undefined`.

We could also merge the both `ifs` 👍

PR #3929, opened 2020-04-17
NXP-28805: review retention automation operations

gcarlin:


modules/platform/nuxeo-retention-core/src/main/java/org/nuxeo/retention/RetentionConstants.java
```diff
@@ -93,6 +93,8 @@
 
     public static final String RULE_ATTACHED_EVENT = "retentionRuleAttached";
 
+    public static final String RETENTION_CATEGORY = "Retention";
+
```
RSalem07, line 97: if the add it first in master please (you did a backport) ignore my comment https://github.com/nuxeo/nuxeo/pull/3931/files#r412219197

PR #3932, opened 2020-04-17
NXP-28939: add the ability to validate an EL

RSalem07:
This PR is in Draft mode to let the involved persons (Florent, Neslon and Nuno) to see how things can be done, to avoid the exception that we have on the retention in the case of EL and event based rule.

I create a specific ticket for adding the validity check: https://jira.nuxeo.com/browse/NXP-28994

T&P In Progress: https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-saouana-10.10-2/4/

As discussed with @nmpcunha we will keep this PR and this improvement. But we will do it on `11.1` only. No need it for `10.10` as we did the split of the properties in https://github.com/nuxeo/nuxeo/pull/3963

### DIFF
--- a/addons/nuxeo-retention/nuxeo-retention-core/src/main/java/org/nuxeo/retention/listeners/RetentionBusinessEventListener.java
+++ b/addons/nuxeo-retention/nuxeo-retention-core/src/main/java/org/nuxeo/retention/listeners/RetentionBusinessEventListener.java
@@ -21,6 +21,9 @@ package org.nuxeo.retention.listeners;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.jboss.el.ExpressionFactoryImpl;
+import org.nuxeo.ecm.core.api.CloseableCoreSession;
+import org.nuxeo.ecm.core.api.CoreInstance;
 import org.nuxeo.ecm.core.api.security.SecurityConstants;
 import org.nuxeo.ecm.core.bulk.BulkService;
 import org.nuxeo.ecm.core.bulk.message.BulkCommand;
@@ -29,6 +32,9 @@ import org.nuxeo.ecm.core.event.EventContext;
 import org.nuxeo.ecm.core.event.EventListener;
 import org.nuxeo.ecm.core.query.sql.NXQL;
 import org.nuxeo.ecm.core.repository.RepositoryService;
+import org.nuxeo.ecm.platform.actions.ELActionContext;
+import org.nuxeo.ecm.platform.el.ExpressionContext;
+import org.nuxeo.ecm.platform.usermanager.UserManager;
 import org.nuxeo.retention.RetentionConstants;
 import org.nuxeo.retention.actions.ProcessRetentionEventAction;
 import org.nuxeo.retention.event.RetentionEventContext;
@@ -60,7 +66,7 @@ public class RetentionBusinessEventListener implements EventListener {
                  .append(RetentionConstants.STARTING_POINT_EVENT_PROP)
                  .append(" = ")
                  .append(NXQL.escapeString(eventName));
-            if (StringUtils.isBlank(eventInput)) {
+            if (StringUtils.isBlank(eventInput) || isExpression(eventInput)) {
                 query.append(" AND ") //
                      .append(RetentionConstants.STARTING_POINT_EXPRESSION_PROP)
                      .append(" IS NULL");
@@ -76,6 +82,19 @@ public class RetentionBusinessEventListener implements EventListener {
                 bulkService.submit(command);
             }
         }
+    }
+
+    protected boolean isExpression(String expression) {
+        UserManager userManager = Framework.getService(UserManager.class);
+        try (CloseableCoreSession session = CoreInstance.openCoreSession(null)) {
+            // we need only to ensure if we can evaluate the EL without any errors, no matter if the EL is evaluated to
+            // true or false. that's why we give the user system and any document
+            ELActionContext ctx = new ELActionContext(new ExpressionContext(), new ExpressionFactoryImpl());
+            ctx.setCurrentPrincipal(userManager.getPrincipal(SecurityConstants.SYSTEM_USERNAME));
+            ctx.setCurrentDocument(session.getRootDocument());
+            return ctx.isValid(expression);
+        }
+
     }
 
 }
--- a/nuxeo-jsf/nuxeo-platform-actions-jsf/src/main/java/org/nuxeo/ecm/platform/actions/jsf/JSFActionContext.java
+++ b/nuxeo-jsf/nuxeo-platform-actions-jsf/src/main/java/org/nuxeo/ecm/platform/actions/jsf/JSFActionContext.java
@@ -25,6 +25,7 @@ import javax.el.ValueExpression;
 import javax.el.VariableMapper;
 import javax.faces.context.FacesContext;
 
+import org.apache.commons.lang3.ClassUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.nuxeo.ecm.core.api.DocumentModel;
 import org.nuxeo.ecm.core.api.NuxeoPrincipal;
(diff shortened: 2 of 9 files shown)


nuxeo-jsf/nuxeo-platform-actions-jsf/src/main/java/org/nuxeo/ecm/platform/actions/jsf/JSFActionContext.java
```diff
@@ -58,15 +59,38 @@ public JSFActionContext(ELContext originalContext, ExpressionFactory expressionF
     @SuppressWarnings("unchecked")
     @Override
     public <T> T evalExpression(String expression, Class<T> expectedType) throws ELException {
+        ValueExpressionDetails veDetails = getValueExpressionDetails(expression, Boolean.class);
```
nmpcunha, line 62: Instead of `Boolean.class` it should be `expectedType`, right?

RSalem07: thanks for the review :)

to keep the backward compatibility and avoid any breaking changes I keep it as it was (initially the passed parameter is not used).

have a look [here](https://github.com/nuxeo/nuxeo/blob/10.10/nuxeo-jsf/nuxeo-platform-actions-jsf/src/main/java/org/nuxeo/ecm/platform/actions/jsf/JSFActionContext.java#L94)

nmpcunha: Fair enough 🙂 
Thanks!

addons/nuxeo-retention/nuxeo-retention-core/src/main/java/org/nuxeo/retention/listeners/RetentionBusinessEventListener.java
```diff
@@ -78,4 +84,20 @@ public void handleEvent(Event event) {
         }
     }
 
+    protected boolean isExpressionLanguage(String expression) {
```
efge, line 87: As in #3963 I'd suggest using simpler names. Here `isExpression` is enough.

addons/nuxeo-retention/nuxeo-retention-core/src/main/java/org/nuxeo/retention/listeners/RetentionBusinessEventListener.java
```diff
@@ -78,4 +84,20 @@ public void handleEvent(Event event) {
...
+        RepositoryService repositoryService = Framework.getService(RepositoryService.class);
+        UserManager userManager = Framework.getService(UserManager.class);
+
+        try (CloseableCoreSession session = CoreInstance.openCoreSession(repositoryService.getRepositoryNames().get(0),
```
efge, line 91: `openCoreSessionSystem(null)` is enough. Although using the default repository (`null`) is a sign that something is wrong, what happens if there are several repositories? Something should be passed as argument or found in the context to decide which one to use.

nuxeo-services/nuxeo-platform-actions-core/src/main/java/org/nuxeo/ecm/platform/actions/ActionContext.java
```diff
@@ -112,4 +112,14 @@
...
 
+    /**
+     * Validates the given expression before evaluated it, which avoid any errors or exception during the evaluation.
+     * <p>
+     * This is designed to be called before {@link #checkCondition(String)}.
+     * 
+     * @param expression the expression to evaluate
+     * @return true if the given expression can be evaluated without exceptions
+     */
+    <T> boolean isValid(String expression);
```
efge, line 123: `<T>` is unused.

nuxeo-services/nuxeo-platform-actions-core/src/main/java/org/nuxeo/ecm/platform/actions/ELActionContext.java
```diff
@@ -58,15 +59,32 @@ public ELActionContext(ELContext originalContext, ExpressionFactory expressionFa
...
+        return (T) valueExpression.getValue(originalContext);
+    }
+
+    @Override
+    public boolean checkCondition(String expression) throws ELException {
+        return Boolean.TRUE.equals(evalExpression(expression, Boolean.class));
+    }
+
+    @Override
+    public <T> boolean isValid(final String expression) {
```
efge, line 72: Drop `final`

nuxeo-jsf/nuxeo-platform-actions-jsf/src/main/java/org/nuxeo/ecm/platform/actions/jsf/JSFActionContext.java
```diff
@@ -91,13 +115,26 @@ public JSFActionContext(ELContext originalContext, ExpressionFactory expressionF
...
+        ValueExpression valueExpression = expressionFactory.createValueExpression(finalContext, expr, expectedType);
+        Class<?> type = valueExpression.getType(originalContext);
+        boolean valid = type != null && ClassUtils.isAssignable(type, expectedType);
+        return new ValueExpressionDetails(finalContext, valueExpression, valid);
     }
 
-    @Override
-    public boolean checkCondition(String expression) throws ELException {
-        return Boolean.TRUE.equals(evalExpression(expression, Boolean.class));
+    protected static class ValueExpressionDetails {
```
efge, line 124: I would also change `ValueExpression` -> `Expression` everywhere (new class and methods)

addons/nuxeo-retention/nuxeo-retention-core/src/main/java/org/nuxeo/retention/listeners/RetentionBusinessEventListener.java
RSalem07: thx, I have a few questions:
- `openCoreSessionSystem(null)` will use the defaut repo. In the case of multi repo, we always have this notion of default ? 

> is a sign that something is wrong, what happens if there are several repositories? Something should be passed as argument or found in the context to decide which one to use.

You mean in the EL context ? If I understand correctly, perhaps I am wrong, but the EL will evaluate the document as a regular pojo and will calls the method on the object, Is it rigt ? if it is the case. Where the session will be used ?

efge: There's always a default repo, this is configurable in repo contribs, and defaults to the "default" repo.
By context I mean whatever context this is being run in, for example operation context, or object context. If you have a session field then use its repository.

addons/nuxeo-retention/nuxeo-retention-core/src/main/java/org/nuxeo/retention/listeners/RetentionBusinessEventListener.java
```diff
@@ -78,4 +84,17 @@ public void handleEvent(Event event) {
         }
     }
 
+    protected boolean isExpression(String expression) {
+        UserManager userManager = Framework.getService(UserManager.class);
+        try (CloseableCoreSession session = CoreInstance.openCoreSession(null)) {
```
efge, line 89: No, I meant `openCoreSessionSystem`, not `openCoreSession`. You want the system user, right?

nuxeo-services/nuxeo-platform-actions-core/src/main/java/org/nuxeo/ecm/platform/actions/ActionContext.java
```diff
@@ -112,4 +112,14 @@
...
 
+    /**
+     * Validates the given expression before evaluated it, which avoid any errors or exception during the evaluation.
+     * <p>
+     * This is designed to be called before {@link #checkCondition(String)}.
+     * 
+     * @param expression the expression to evaluate
+     * @return true if the given expression can be evaluated without exceptions
+     */
+    boolean isValid(String expression);
```
efge, line 123: `@since 11.1` and you need a default implementation for backward compatibility.

addons/nuxeo-retention/nuxeo-retention-core/src/main/java/org/nuxeo/retention/listeners/RetentionBusinessEventListener.java
RSalem07: yes, sorry

addons/nuxeo-retention/nuxeo-retention-core/src/main/java/org/nuxeo/retention/listeners/RetentionBusinessEventListener.java
efge: Given a re-reading of the code, the `isExpression` check depends on the repository, so the query construction must be moved inside the loop over all repositories.

RSalem07:
As discussed with Florent / Nuno. The split of the properties [here](https://github.com/nuxeo/nuxeo/pull/3963) will fix the issue that we have and adding a new improvement to validate the EL is no longer needed, to avoid any breaking changes/ perfs issues

PR #3934, opened 2020-04-17
NXP-28396: Don't hang on stop if nuxeo become a defunct process

kevinleturc:


server/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -1929,6 +1930,12 @@ public void doStop() {
         }
     }
 
+    protected boolean isAliveAndNotZombie(ProcessHandle process) {
+        // nuxeo process could become a defunct process (aka zombie) due to nuxeoctl not reading its exit code
+        // defunct processes loose their command because system just wants to save exit code
```
efge, line 1935: loose -> lose

PR #3939, opened 2020-04-18
NXP-28952: fix retention/legal hold with transactional blob stores

efge:


modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStoreConfiguration.java
```diff
@@ -412,6 +423,33 @@ protected TransferManager createTransferManager() {
...
+        DefaultRetention defaultRetention = rule.getDefaultRetention();
+        if (defaultRetention == null) {
+            return DEFAULT_RETENTION_MODE;
+        }
+        String mode = defaultRetention.getMode();
+        if (mode == null) {
+            return DEFAULT_RETENTION_MODE;
+        }
+        return ObjectLockRetentionMode.valueOf(mode);
+    }
```
kevinleturc, line 451: If Optional creation is not an issue, you could save some lines:
```
return Optional.ofNullable(result.getObjectLockConfiguration())
                .map(ObjectLockConfiguration:: getRule)
                .map(ObjectLockRule:: getDefaultRetention)
                .map(DefaultRetention::getMode)
                .map(ObjectLockRetentionMode::valueOf)
                .orElse(DEFAULT_RETENTION_MODE);
```

efge: Nice, I updated the code.

PR #3947, opened 2020-04-20
NXP-28532: mention failing doc id and property when reindexing ES

NourNuxeo:
Helps with debugging when ES reindexing fails

### DIFF
--- a/modules/platform/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/main/java/org/nuxeo/elasticsearch/io/JsonESDocumentWriter.java
+++ b/modules/platform/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/main/java/org/nuxeo/elasticsearch/io/JsonESDocumentWriter.java
@@ -38,6 +38,7 @@ import org.nuxeo.ecm.automation.core.util.JSONPropertyWriter;
 import org.nuxeo.ecm.core.api.CoreSession;
 import org.nuxeo.ecm.core.api.DocumentModel;
 import org.nuxeo.ecm.core.api.DocumentRef;
+import org.nuxeo.ecm.core.api.NuxeoException;
 import org.nuxeo.ecm.core.api.model.Property;
 import org.nuxeo.ecm.core.api.security.ACE;
 import org.nuxeo.ecm.core.api.security.ACL;
@@ -212,7 +213,12 @@ public class JsonESDocumentWriter {
         }
 
         for (Property p : properties) {
-            writer.writeProperty(jg, p);
+            try {
+                writer.writeProperty(jg, p);
+            } catch (ClassCastException e) {
+                throw new NuxeoException(
+                        String.format("writing JSON property failed on document: %s for property: %s", doc, p), e);
+            }
         }
     }
 


modules/platform/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/main/java/org/nuxeo/elasticsearch/io/JsonESDocumentWriter.java
```diff
@@ -212,7 +217,14 @@ protected static void writeProperties(JsonGenerator jg, DocumentModel doc, Strin
         }
 
         for (Property p : properties) {
-            writer.writeProperty(jg, p);
+            try {
+                writer.writeProperty(jg, p);
+            } catch (ClassCastException e) {
+                var nuxeoException = new NuxeoException(e);
+                nuxeoException.addInfo(
+                        String.format("writing JSON property failed on document: %s for property: %s", doc, p));
```
efge, line 225: You can do directly `throw new NuxeoException(message, e)`, no need for `addInfo` which is only useful if you rethrow the same `NuxeoException` that you caught.

modules/platform/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/main/java/org/nuxeo/elasticsearch/io/JsonESDocumentWriter.java
```diff
@@ -212,7 +217,12 @@ protected static void writeProperties(JsonGenerator jg, DocumentModel doc, Strin
         }
 
         for (Property p : properties) {
-            writer.writeProperty(jg, p);
+            try {
+                writer.writeProperty(jg, p);
+            } catch (ClassCastException e) {
+                throw new NuxeoException(
+                        String.format("writing JSON property failed on document: %s for property: %s", doc, p));
```
efge, line 224: Please keep the original exception as last argument when throwing a new one.

NourNuxeo: Yes sorry about that. Fixed

modules/platform/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/main/java/org/nuxeo/elasticsearch/io/JsonESDocumentWriter.java
```diff
@@ -59,6 +62,8 @@
  */
 public class JsonESDocumentWriter {
 
+    private static final Logger log = LogManager.getLogger(JsonESDocumentWriter.class);
```
kevinleturc, line 65: Doesn't seem used.

NourNuxeo: yes not anymore. Removing 😬

PR #3972, opened 2020-04-24
NXBT-3289: remove buggy empty-docker-secret

NourNuxeo:


Jenkinsfile
```diff
@@ -660,6 +660,10 @@ pipeline {
                 // Previous preview deployment needs to be scaled to 0 to be replaced correctly
                 sh "kubectl -n ${PREVIEW_NAMESPACE} scale deployment nuxeo-preview --replicas=0"
               }
+              sh """kubectl create secret generic kubernetes-docker-cfg \
+                  --namespace=${PREVIEW_NAMESPACE} \
+                  --from-literal=.dockerconfigjson="\$(kubectl --namespace platform get secret kubernetes-docker-cfg -ojsonpath='{.data.\\.dockerconfigjson}' | base64 --decode)" \
+                  --type=kubernetes.io/dockerconfigjson --dry-run -o yaml | kubectl apply -f -"""
```
ataillefer, line 666: why not simply
```
kubectl create secret generic kubernetes-docker-cfg \
                  --namespace=${PREVIEW_NAMESPACE} \
                  --from-literal=.dockerconfigjson="\$(kubectl --namespace platform get secret kubernetes-docker-cfg -ojsonpath='{.data.\\.dockerconfigjson}' | base64 --decode)" \
                  --type=kubernetes.io/dockerconfigjson
```

Jenkinsfile
```diff
@@ -660,6 +660,10 @@ pipeline {
                 // Previous preview deployment needs to be scaled to 0 to be replaced correctly
                 sh "kubectl -n ${PREVIEW_NAMESPACE} scale deployment nuxeo-preview --replicas=0"
               }
+              sh """kubectl create secret generic kubernetes-docker-cfg \
+                  --namespace=${PREVIEW_NAMESPACE} \
+                  --from-literal=.dockerconfigjson="\$(kubectl --namespace platform get secret kubernetes-docker-cfg -ojsonpath='{.data.\\.dockerconfigjson}' | base64 --decode)" \
```
ataillefer, line 665: Could you pleas use a constant for 'platform' as we try to do for kubectl commands.
I'd say, after this [line](https://github.com/nuxeo/nuxeo/blob/fix-NXBT-3289-kubernetes-docker-cfg/Jenkinsfile#L21):
```
kubernetesNamespace = 'platform'
```

Jenkinsfile
NourNuxeo: Not sure it works if I paste some bash code directly in a script directive which takes groovy initially?
I can also try and remove it but I wanted to stay consistent with the all the other kubectl calls accros the same file which are wrapped in an sh""" """" call. 
wdyt ?

Jenkinsfile
NourNuxeo: Changed the single quotes to double quotes on the whole lines where the var is used

Jenkinsfile
troger: @ataillefer is just simplifying the content of the `sh"""`, not saying you should remove the `sh"""` part :)

Jenkinsfile
```diff
@@ -158,7 +159,7 @@ def buildUnitTestStage(env) {
     stage("Run ${env} unit tests") {
       container("maven-${env}") {
         script {
-          setGitHubBuildStatus("platform/utests/${env}", "Unit tests - ${env} environment", 'PENDING')
+          setGitHubBuildStatus("${kubernetesNamespace}/utests/${env}", "Unit tests - ${env} environment", 'PENDING')
```
troger, line 162: I would say this one is not linked to the Kubernetes namespace, it's just a GitHub status name.

NourNuxeo: yes I reverted all those gh status changes

PR #3974, opened 2020-04-27
NXP-28846: replace doctype icons

nmpcunha:
Ongoing T&P: https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-ncunha/43/

Just wondering if the icons change required by @carolinagouveia would have impact on other apps like Drive and/or Mobile? (cc @BoboTiG )


BoboTiG:
There is no impact on the Drive side yet. But it is good to see that PR, as we will need such icons in a near future :)

troger:
It should not have an impact on the Mobile app (I hope), let's test it after the merge :)

BTW, why are we mixing `gif` and `png` in this "refactoring"?

nmpcunha:
> It should not have an impact on the Mobile app (I hope), let's test it after the merge :)
> 
> BTW, why are we mixing `gif` and `png` in this "refactoring"?

The inicial idea was to use `svg` everywhere. But due to the uncertain about the other apps, the decision for now was to map the existing ones (`gif` and `png`). While doing the replacements I tried to map to the most common use case (small -> `gif`, big -> `png`) since it was the format @carolinagouveia provided, for consistency purposes.

BoboTiG:
FTR on Drive, we use only SVG.

troger:
I would say consistency should be all `png` or all `gif`, seems not consistent to me to mix them both 😄

nmpcunha:
> I would say consistency should be all `png` or all `gif`, seems not consistent to me to mix them both 😄

🤷‍♂️ 
Agree with you! 😄  
But I guess we would only be able to get that kind of consistency when we decide to move forward with svg only, or "other" only approach. (cc @carolinagouveia and @bchauvin )

atchertchian:
My 2 cents: it could be worth checking if the [MimetypeIconUpdater listener](https://github.com/nuxeo/nuxeo/blob/10.10/nuxeo-services/nuxeo-platform-filemanager-core-listener/src/main/java/org/nuxeo/ecm/platform/filemanager/core/listener/MimetypeIconUpdater.java) still persists the path to icons, as it did at some point.
That's why we ended up creating a "nuxeo-resources-compat" for https://jira.nuxeo.com/browse/NXP-10928 for instance.

atchertchian:
(see also #3980)

PR #3986, opened 2020-04-29
Fix nxp 28709 log silent failure in automation bulk.run action

NourNuxeo:
setError is called twice.
The second call overrides the first one so it is conditioned.

There is some logic to reconstitute the whole error message. I would rather have changed setError to take the throwable and do this work than just giving it a string.

Then no matter how many times we call setError with the same executionId and throwable, we would always get the same resulting message

### DIFF
--- a/modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
+++ b/modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
@@ -60,6 +60,7 @@ import org.nuxeo.ecm.core.api.NuxeoException;
 import org.nuxeo.ecm.core.api.NuxeoPrincipal;
 import org.nuxeo.ecm.core.transientstore.api.TransientStore;
 import org.nuxeo.ecm.core.transientstore.api.TransientStoreService;
+import org.nuxeo.ecm.platform.web.common.exceptionhandling.ExceptionHelper;
 import org.nuxeo.ecm.platform.web.common.vh.VirtualHostHelper;
 import org.nuxeo.ecm.webengine.model.WebAdapter;
 import org.nuxeo.ecm.webengine.model.exceptions.WebResourceNotFoundException;
@@ -164,7 +165,7 @@ public class AsyncOperationAdapter extends DefaultAdapter {
 
             @Override
             public OperationException onError(OperationException error) {
-                setError(executionId, error.getMessage());
+                setError(executionId, error);
                 return error;
             }
 
@@ -182,7 +183,7 @@ public class AsyncOperationAdapter extends DefaultAdapter {
                     opCtx.setCoreSession(s);
                     service.run(opCtx, opId, xreq.getParams());
                 } catch (OperationException e) {
-                    setError(executionId, e.getMessage());
+                    setError(executionId, e);
                 } finally {
                     LoginComponent.popPrincipal();
                 }
@@ -200,11 +201,11 @@ public class AsyncOperationAdapter extends DefaultAdapter {
     @GET
     @Path("{executionId}/status")
     public Object status(@PathParam("executionId") String executionId) throws IOException, MessagingException {
+        String error = getError(executionId);
+        if (error != null) {
+            throw new NuxeoException(error, HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+        }
         if (isCompleted(executionId)) {
-            String error = getError(executionId);
-            if (error != null) {
-                throw new NuxeoException(error, HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
-            }
             String resURL = String.format("%s/%s", getPath(), executionId);
             return redirect(resURL);
         } else {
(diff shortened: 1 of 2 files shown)


modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
```diff
@@ -200,11 +201,11 @@ public OperationException onError(OperationException error) {
     @GET
     @Path("{executionId}/status")
     public Object status(@PathParam("executionId") String executionId) throws IOException, MessagingException {
+        String error = getError(executionId);
```
bdelbosc, line 204: So now each normal call to /status costs 2 round trips, one to the transient store to check for an error and one to the kv to get the status.

NourNuxeo: The whole problem comes from not checking errors on incomplete async operations. We need to make sure there is no error completed or not.
Not a dream state but 2 calls vs 1 call is still constant complexity-wise 🤷

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
```diff
@@ -164,7 +165,7 @@ public void onOperationExit(Object output) {
 
             @Override
             public OperationException onError(OperationException error) {
-                setError(executionId, error.getMessage());
+                setError(executionId, ExceptionHelper.unwrapException(error).getMessage());
```
troger, line 168: I would have added a new method `setError(String, Throwable)` to handle the `unwrapException` logic (and deprecate the existing `setError` method probably as it won't be used anymore).

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
nelsonsilva: when `setError` is called is also calls `setCompleted`, in which case do we have errors with an execution not marked as completed ?

kevinleturc:
Should we add a unit test?

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
NourNuxeo: when failing while compiling the operation chain, isCompleted is never called. Since it conditions the call to getErrors() we never see the error. So we made the call to getErrors independant and preceding isCompleted(): https://github.com/nuxeo/nuxeo/pull/3986/commits/028ad6c66f4120654dfdaec4b0391cde20220e1f#diff-632d12880ac130d2edc7fa897f2779eaR205-L207

modules/platform/nuxeo-automation/nuxeo-automation-test/src/test/java/org/nuxeo/ecm/automation/server/test/AsyncOperationAdapterTest.java
```diff
@@ -222,6 +222,19 @@ public void testAsyncBulkAction() throws Exception {
...
     }
 
+    @Test
+    public void testFailingBulkAction() throws Exception {
+        String result = async.newRequest(BulkRunAction.ID) //
+                             .set("action", AutomationBulkAction.ACTION_NAME)
+                             .set("query", "SELECT * FROM Folder")
+                             .set("bucketSize", "10")
+                             .set("batchSize", "5")
+                             .set("parameters", MAPPER.writeValueAsString("{}"))
```
kevinleturc, line 232: Can we just put `"{}"` without calling the MAPPER? `writeValueAsString` is used to serialize an object to JSON (or other), here you're giving to it a String. My first thought is: writing a string as string will return the input string. If it's not the case, having the final value in the test will help us to understand what makes it fail and what we want to control/catch.

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
```diff
@@ -289,6 +290,13 @@ protected void enterMethod(String executionId, InvokableMethod method) {
...
+    protected void setError(String executionId, Throwable t) {
+        getTransientStore().putParameter(executionId, TRANSIENT_STORE_ERROR,
+                ExceptionHelper.unwrapException(t).getMessage());
+        setCompleted(executionId);
+    }
+
+    @Deprecated
     protected void setError(String executionId, String error) {
         getTransientStore().putParameter(executionId, TRANSIENT_STORE_ERROR, error);
         setCompleted(executionId);
```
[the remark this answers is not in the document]
NourNuxeo: I was wondering what would happen when the deprecated method is removed. I guess there is some mechanism around that ?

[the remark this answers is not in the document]
kevinleturc: Deprecated methods are not removed automatically. This is us who remove them during a Cleanup / Format commit or a dedicated ticket if there's a lot to remove.

PR #3991, opened 2020-05-01
NXP-29057: don't save toplevel properties that aren't dirty

efge:
T&P ok https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/813/


modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/api/DocumentModelFactory.java
```diff
@@ -192,6 +192,14 @@ public static DataModel createDataModel(Document doc, Schema schema) {
      */
     @SuppressWarnings("deprecation")
     public static DocumentModel writeDocumentModel(DocumentModel docModel, Document doc) {
+        return writeDocumentModel(docModel, doc, false);
+    }
+
+    /**
+     * Writes a document model to a document. Returns the re-read document model.
+     */
```
troger, line 200: Since 11.1.

PR #3994, opened 2020-05-04
Fix nxp 29066 Init of the ACL document root occurs many times [forward port]

RSalem07:
Please review the `10.10` one https://github.com/nuxeo/nuxeo/pull/3995 before.
 
T&P OK: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-saouana-2/11/

### DIFF
--- a/modules/platform/nuxeo-platform-content-template-manager/src/main/java/org/nuxeo/ecm/platform/content/template/factories/SimpleTemplateBasedRootFactory.java
+++ b/modules/platform/nuxeo-platform-content-template-manager/src/main/java/org/nuxeo/ecm/platform/content/template/factories/SimpleTemplateBasedRootFactory.java
@@ -37,6 +37,8 @@ public class SimpleTemplateBasedRootFactory extends SimpleTemplateBasedFactory {
     public void createContentStructure(DocumentModel eventDoc) {
         initSession(eventDoc);
 
+        boolean initRootAcl = shouldInitAcl(eventDoc);
+
         for (TemplateItemDescriptor item : template) {
             if (!shouldCreateDocument(eventDoc.getId(), item.getId(), item.getTypeName())) {
                 continue;
@@ -55,7 +57,9 @@ public class SimpleTemplateBasedRootFactory extends SimpleTemplateBasedFactory {
             session.save();
         }
         // init root ACL if really empty
-        setAcl(acl, eventDoc.getRef());
+        if (initRootAcl) {
+            setAcl(acl, eventDoc.getRef());
+        }
     }
 
     /**
@@ -72,6 +76,29 @@ public class SimpleTemplateBasedRootFactory extends SimpleTemplateBasedFactory {
         try (IterableQueryResult it = session.queryAndFetch(query, NXQL.NXQL)) {
             return !it.iterator().hasNext();
         }
+    }
+
+    /**
+     * Checks if we should init the acl of the root document.
+     * <p>
+     * The root init acl should occur only once, At the beginning where root document doesn't contain any child.
+     *
+     * @return {@code true} if the root document doesn't contain any child of template document type, {@code false}
+     *         otherwise
+     */
+    protected boolean shouldInitAcl(DocumentModel rootDoc) {
+        for (TemplateItemDescriptor item : template) {
+            String query = String.format("SELECT %s FROM Document WHERE %s = '%s' AND %s = '%s'", //
+                    NXQL.ECM_UUID, //
+                    NXQL.ECM_PARENTID, rootDoc.getId(), //
+                    NXQL.ECM_PRIMARYTYPE, item.getTypeName());
+            try (IterableQueryResult it = session.queryAndFetch(query, NXQL.NXQL)) {
+                if (it.iterator().hasNext()) {
+                    return false;
+                }
+            }
+        }
+        return true;
     }
 
 }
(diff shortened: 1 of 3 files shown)


modules/platform/nuxeo-platform-content-template-manager/src/main/java/org/nuxeo/ecm/platform/content/template/factories/SimpleTemplateBasedRootFactory.java
```diff
@@ -37,8 +37,10 @@
     public void createContentStructure(DocumentModel eventDoc) {
         initSession(eventDoc);
 
+        boolean initRootAcl = true;
         for (TemplateItemDescriptor item : template) {
             if (!shouldCreateDocument(eventDoc.getId(), item.getId(), item.getTypeName())) {
+                initRootAcl = false;
```
kevinleturc, line 43: Shouldn't it be the opposite? Start with false and if we found one to create set it to true?

modules/core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSTransactionState.java
```diff
@@ -641,6 +641,10 @@ public void updateReadACLs(Collection<String> docIds) {
      */
     protected void updateDocumentReadAcls(String id) {
         DBSDocumentState docState = getStateForUpdate(id);
+        if(docState == null) {
+            log.trace("No document state found for id: " + id);
+            return;
+        }
```
kevinleturc, line 647: Maybe something like below?
```java
            if (docState == null) {
                if (log.isTraceEnabled()) {
                    log.trace("Cannot fetch document with id: " + id, new Throwable("debug stack trace"));
                }
                return;
            }
```

modules/platform/nuxeo-platform-content-template-manager/src/main/java/org/nuxeo/ecm/platform/content/template/factories/SimpleTemplateBasedRootFactory.java
RSalem07: thx for the point, 
My idea here is to keep the same flow that was before my initial fix see [here](https://github.com/nuxeo/nuxeo/commit/8f84864a3bcdf4190b72409d0ff194929b471f85)

before the fix above, the acl was computed/inited only once (call to `shouldCreateContent` returns true if at least once child exist) this is the reason why we didn't see the problem before.

efge:
This will need a dedicated unit test for the pure `updateDocumentReadAcls` part to show exactly what behavior was being done incorrectly and is being fixed.

Also in the NXP-29048 you should not use GitHub links too ...nuxeo/nuxeo/blob/master... because these will be invalid as soon as we do further commits. Please replace `master` with the current git revision, so that the links are still valid in the future.

modules/platform/nuxeo-platform-content-template-manager/src/main/java/org/nuxeo/ecm/platform/content/template/factories/SimpleTemplateBasedRootFactory.java
kevinleturc: Ok, but the flow has been changed with the fix. Now it makes more sense to me to update root ACLs if a child was created whether to not update them if we didn't create a child, or I'm missing something.

Do you know what happen in such cases? Ie: a child is created and another one is skipped which results in not setting root acls.

RSalem07: I see your point of view.I was thinking the same thing. 

>Do you know what happen in such cases? Ie: a child is created and another one is skipped which results in not setting root acls.

>...Ie: a child is created and another one is skipped

with introduced fix each child is created, but the acl of root is re-computed each time. even they don't change, in the case of MongoDB it will re-compute all acl of the children but without making  a merge which it means that the acl are removed and re-created from the root onces and if I am not wrong we will even loose the acl defined on each document.

kevinleturc: Ok, is it the case of acls or just read calls?

Are you saying the last set acl is not useful?

RSalem07:
@efge 
>This will need a dedicated unit test for the pure updateDocumentReadAcls part to show exactly what behavior was being done incorrectly and is being fixed.

The problem that we have with the introduce fix is that in some case a document is deleted by another thread where the main one try to update its acl. mainly see in a test scenario. I put the whole description in the jira ticket  https://jira.nuxeo.com/browse/NXP-29048 

after a discussion with Kevin we take the choice to avoid the check that I introduced on the existence of a document before the acl set.

efge:
I don't understand how your last comment answers my request about a dedicated non-regression unit test.

RSalem07:
>I don't understand how your last comment answers my request about a dedicated non-regression unit test.

I was thinking that you asking adding test to cover the `updateDocumentReadAcls` and the NPE. But what I said is that I am going to avoid the modification.

bdelbosc:
It is still unclear to me how this PR fixes the random on MongoDB `WorkflowEndpointTest.testRejectTaskInSubWorkflow` which happens a lot.
Is it by removing the need to run `updateReadAcl` a second time during the `@Deploy` of the method avoiding concurrency with the cleaning?
We need to merge this or revert NXP-28958.

RSalem07:
@bdelbosc you can find more on https://github.com/nuxeo/nuxeo/pull/3995#issue-412879995. I updated it description to answer to your interrogation:

>this fix should also fix the failure test on MongoDB, because now once we @Deploy run it will not reinit the Acl for a second time and it will avoid the race condition that occurs.


for your second question about the revert, I am waiting for a second review on `10.10` + T&P (H2 and Mongo) try to merge it asap. and then I will ask for a review for the master one.

sorry for the side effects

PR #3995, opened 2020-05-04
Fix nxp 29066 Init of the ACL document root occurs many times

RSalem07:
T&P: https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-saouana-10.10/129/

T&P MongoDB: https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-saouana-10.10-2/6/

**Context**

Initially this PR was open to fix  [NXP-29048](https://jira.nuxeo.com/browse/NXP-29048) which is a race condition issue that happens only in the test context. But to fix the problem that was initially introduced by the fix [here](https://github.com/nuxeo/nuxeo/commit/8f84864a3bcdf4190b72409d0ff194929b471f85) This new fix allows us to see the problem of the race condition that happens when we add `@Deploy` on the test (see [NXP-29048](https://jira.nuxeo.com/browse/NXP-29048)).

But with the new fix the root acl init will occurs each time, which is not the expected behaviour, we should init the acl only once, as it was before the fix. But we should be able to add new entry in the template.

to recap:
To fix the issue introduced by  [NXP-28958](https://jira.nuxeo.com/browse/NXP-28958) on init the acl many times. I created a new   [NXP-29066](https://jira.nuxeo.com/browse/NXP-29066). 

this fix should also fix the failure test on MongoDB, because now once we `@Deploy` run it will not reinit the Acl for a second time and it will avoid the race condition that occurs.

### DIFF
--- a/nuxeo-features/nuxeo-platform-audit/nuxeo-platform-audit-core/src/test/java/org/nuxeo/ecm/platform/audit/TestSimpleTemplateBasedRootAudit.java
+++ b/nuxeo-features/nuxeo-platform-audit/nuxeo-platform-audit-core/src/test/java/org/nuxeo/ecm/platform/audit/TestSimpleTemplateBasedRootAudit.java
@@ -1,0 +1,103 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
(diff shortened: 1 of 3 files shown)


RSalem07:
@efge as asked I added the test and rework the fix. 

please for the review start by this one as it required for the HF26
thx

nuxeo-features/nuxeo-platform-audit/nuxeo-platform-audit-core/src/test/java/org/nuxeo/ecm/platform/audit/TestSimpleTemplateBasedRootAudit.java
```diff
@@ -0,0 +1,103 @@
...
+    @Inject
+    protected PageProviderService pps;
+
+    @Test
+    @Deploy("org.nuxeo.ecm.platform.audit.tests:test-add-content-template-contrib.xml")
+    public void shouldInitRootAclOnlyOnce() {
+        DocumentModel root = session.getDocument(new PathRef("/"));
+        @SuppressWarnings("unchecked")
+        PageProvider<LogEntry> pp = (PageProvider<LogEntry>) pps.getPageProvider("DOCUMENT_HISTORY_PROVIDER", null, 3l,
+                0l, new HashMap<String, Serializable>(), root);
```
efge, line 75: In the future please use `L` instead of `l` for readability. `l` is very easily confused with `1`.

PR #4001, opened 2020-05-05
NXP-29057: don't save toplevel properties that aren't dirty (except on creation)

efge:
Compared to the last PR (#3991) there is an additional fix for empty lists (https://github.com/nuxeo/nuxeo/pull/4001/files#diff-11b83df5ae25e9402fd3c7bf02b89cceR462-R468) and a corresponding unit test (https://github.com/nuxeo/nuxeo/pull/4001/files#diff-afd96e2e7f9352fbd75b7811996aa34fR427)

PR #4002, opened 2020-05-05
NXP-29057: don't save toplevel properties that aren't dirty (except on creation) (10.10)

efge:
This is a new version of #3992 that had a bug.
Master PR is #4001.


efge:
T&P ok https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume-10.10/95/

PR #4008, opened 2020-05-05
NXP-28798: Gracefully log missing DocumentRoutes

NourNuxeo:


modules/platform/nuxeo-platform-document-routing/nuxeo-routing-core/src/main/java/org/nuxeo/ecm/platform/routing/core/io/TaskWriter.java
```diff
@@ -219,6 +224,9 @@ public void writeEntityBody(Task item, JsonGenerator jg) throws IOException {
 
                 jg.writeEndObject();
             }
+        } catch (DocumentNotFoundException e) {
+            // If the DocumentRoute is not found (deleted for example), the TaskWriter fails.
+            log.error(e);
```
efge, line 229: If you want to log the stack trace you need `log.error(e, e)`.
If you don't, please add a comment saying so because it's surprising to hide it.

NourNuxeo: Yes no special reason to hide the stack! Fixed!

efge:
I asked
> Also please fix the typo in `routs` in the commit message.

But this hasn't been done yet.

NourNuxeo:
now it is!

modules/platform/nuxeo-platform-document-routing/nuxeo-routing-core/src/main/java/org/nuxeo/ecm/platform/routing/core/io/TaskWriter.java
```diff
@@ -219,6 +224,9 @@ public void writeEntityBody(Task item, JsonGenerator jg) throws IOException {
 
                 jg.writeEndObject();
             }
+        } catch (DocumentNotFoundException e) {
+            // If the DocumentRoute is not found (deleted for example), the TaskWriter fails.
+            log.error(e, e);
```
efge, line 229: If a sysadmin can't do much about it and the application can keep functioning, I'd prefer a WARN.

Also where is the exception being thrown exactly? Could you add a full stack trace in the NXP ticket please. I'm asking because the try/catch has a very large scope, which is bad if we can have a smaller one.
From a quick test it seems to appear in `nodeAccessRunner.runUnrestricted()` so the try/catch should be around that, and you should `log.warn("Failed to get workflow instance", e)`

NourNuxeo: Making it a warn and updating the ticket!

Yes indeed, I just catched there because there was already a try clause and because I was certain about the context this is called in (by context I mean while writing a task to json). The `UnrestrictedSessionRunner.runUnrestricted()` is called in so many places I didn't want to catch in it.

I could also make a sub try catch to make sure I only around the `nodeAccessRunner.runUnrestricted();`call in `TaskWriter.writeEntityBody()` line 116. At first I wanted to avoid doing a sub try catch.

efge: Yes I mean around the call to `nodeAccessRunner.runUnrestricted` at line 116. It helps the reader to know where the exception can come from.

NourNuxeo: Done!

modules/platform/nuxeo-platform-document-routing/nuxeo-routing-core/src/main/java/org/nuxeo/ecm/platform/routing/core/io/TaskWriter.java
```diff
@@ -108,7 +113,12 @@ public void writeEntityBody(Task item, JsonGenerator jg) throws IOException {
             if (StringUtils.isNotBlank(workflowInstanceId)) {
                 NodeAccessRunner nodeAccessRunner = new NodeAccessRunner(wrapper.getSession(), workflowInstanceId,
                         nodeId);
-                nodeAccessRunner.runUnrestricted();
+                try {
+                    nodeAccessRunner.runUnrestricted();
```
kevinleturc, line 117: Why not fixing the root cause in `NodeAccessRunner` instead of catching?

NourNuxeo: As we discussed on slack the root cause is a DocumentNotFound exception we can only catch where we need to.

We have identified other issues as well: the NodeAccessRunner is also called by `org.nuxeo.ecm.platform.routing.core.io.TaskCompletionRequestJsonReader.readEntity(JsonNode)`

When we list the tasks, corrupted tasks show, but obviously miss information, which we may want to avoid in the UI.

Here is a sample response with a first corrupted task and a well formed task:
https://jira.nuxeo.com/browse/NXP-28798?focusedCommentId=453928&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-453928

modules/platform/nuxeo-platform-document-routing/nuxeo-routing-core/src/main/java/org/nuxeo/ecm/platform/routing/core/io/TaskWriter.java
NourNuxeo: After moving the catch, now the faulty task contains more info in the response. I have updated the jira ticket with a response containing a corrupted task first and then a wellformed one. Also Kevin has identified some issues we'd like to have your opinion about: https://github.com/nuxeo/nuxeo/pull/4008#discussion_r420618445

modules/platform/nuxeo-platform-document-routing/nuxeo-routing-core/src/main/java/org/nuxeo/ecm/platform/routing/core/io/TaskWriter.java
```diff
@@ -108,7 +113,12 @@ public void writeEntityBody(Task item, JsonGenerator jg) throws IOException {
...
+                } catch (DocumentNotFoundException e) {
+                    // If the DocumentRoute is not found (deleted for example), the TaskWriter fails.
+                    log.warn(e, e);
```
troger, line 120: As @efge said in one of his comments, you should log something meaningful if possible:
```
log.warn("Failed to get workflow instance", e)
```
BTW, do we want the whole stack trace for a WARN? (cc @kevinleturc)

modules/platform/nuxeo-platform-document-routing/nuxeo-routing-core/src/main/java/org/nuxeo/ecm/platform/routing/core/io/TaskWriter.java
```diff
@@ -108,7 +113,12 @@ public void writeEntityBody(Task item, JsonGenerator jg) throws IOException {
             if (StringUtils.isNotBlank(workflowInstanceId)) {
                 NodeAccessRunner nodeAccessRunner = new NodeAccessRunner(wrapper.getSession(), workflowInstanceId,
                         nodeId);
-                nodeAccessRunner.runUnrestricted();
+                try {
+                    nodeAccessRunner.runUnrestricted();
+                } catch (DocumentNotFoundException e) {
+                    // If the DocumentRoute is not found (deleted for example), the TaskWriter fails.
```
troger, line 119: The comment is not accurate, the TaskWriter does not fail anymore with your fix. Keep it simple...
```
// Workflow instance not found
```
I'm not sure it's even needed according to the `DocumentNotFoundException` and the log message?

NourNuxeo: At the begining the catch was at the end of the try resource (way more line below) so I explained why I was catching.
I should have removed it when I made a very local try catch

modules/platform/nuxeo-platform-document-routing/nuxeo-routing-core/src/main/java/org/nuxeo/ecm/platform/routing/core/io/TaskWriter.java
kevinleturc: I would say no as this is a valid error case, I'm wondering if we still want it behind a debug/trace level.

troger: So a simple:
```
log.warn("Failed to get workflow instance");
log.debug(e, e);
```
?

kevinleturc: Looks good 👍

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
```diff
@@ -0,0 +1,33 @@
+import org.apache.logging.log4j.LogManager
```
kevinleturc, line 1: There's still an interrogation about committing this script or not.

troger: Checked with @tmartins ans he is OK committing the script there 😃

troger, line 1: Could you rename the script to match the already existing one (consistency)? Such as `deleteOrphanTasks.groovy`.

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
```diff
@@ -0,0 +1,33 @@
...
+import org.nuxeo.ecm.platform.task.Task
+import org.nuxeo.ecm.core.api.IdRef
+import org.nuxeo.ecm.core.api.ScrollResult
+import org.nuxeo.runtime.transaction.TransactionHelper
+import java.util.HashMap
+
+log = LogManager.getLogger("taskGC.groovy")
```
troger, line 8: Even if it's not mandatory, could you take the habit to declare your vars at least with `def` if you don't want to type them?

```suggestion
def log = LogManager.getLogger("taskGC.groovy")
```

Same for all vars in the script.

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
```diff
@@ -0,0 +1,33 @@
...
+nxql = "SELECT * FROM Document WHERE ecm:mixinType = 'Task'"
+dryRun = true
+session = Context.getCoreSession()
+ScrollResult<String> scrollResult = session.scroll(nxql, 500, 60)
```
troger, line 12: I'm just wondering, is `500` not too much as batch size to avoid transaction timeout? Or could it be increased? Or `500` is just fine :)
@kevinleturc @bdelbosc ?

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
```diff
@@ -0,0 +1,33 @@
...
+documentCount = 0
+
+while (scrollResult.hasResults()) {
+    documentCount += scrollResult.getResults().size()
+    cache = new HashMap<IdRef>()
+    for(id in scrollResult.getResults()) {
+        task = session.getDocument(new IdRef(id)).getAdapter(Task.class)
+        workflowRef = new IdRef(task.getProcessId())
+        if(!cache.computeIfAbsent(workflowRef, session.&exists)) {
+            log.info("removing orphan task " + id)
```
troger, line 23: Can't we use the same as in Java?
```suggestion
            log.info("removing orphan task: {}", id)
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
```diff
@@ -0,0 +1,33 @@
...
+dryRun = true
+session = Context.getCoreSession()
+ScrollResult<String> scrollResult = session.scroll(nxql, 500, 60)
+bucketCount = 0
+documentCount = 0
+
+while (scrollResult.hasResults()) {
+    documentCount += scrollResult.getResults().size()
+    cache = new HashMap<IdRef>()
+    for(id in scrollResult.getResults()) {
```
troger, line 19:
```suggestion
    for (id in scrollResult.getResults()) {
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
```diff
@@ -0,0 +1,33 @@
...
+        task = session.getDocument(new IdRef(id)).getAdapter(Task.class)
+        workflowRef = new IdRef(task.getProcessId())
+        if(!cache.computeIfAbsent(workflowRef, session.&exists)) {
```
troger, line 22:
```suggestion
        if (!cache.computeIfAbsent(workflowRef, session.&exists)) {
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
```diff
@@ -0,0 +1,33 @@
...
+import org.nuxeo.runtime.transaction.TransactionHelper
+import java.util.HashMap
+
+log = LogManager.getLogger("taskGC.groovy")
+nxql = "SELECT * FROM Document WHERE ecm:mixinType = 'Task'"
+dryRun = true
+session = Context.getCoreSession()
+ScrollResult<String> scrollResult = session.scroll(nxql, 500, 60)
+bucketCount = 0
+documentCount = 0
```
troger, line 14: Either use it or just remove those variables.

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
```diff
@@ -0,0 +1,33 @@
...
+
+while (scrollResult.hasResults()) {
+    documentCount += scrollResult.getResults().size()
+    cache = new HashMap<IdRef>()
+    for(id in scrollResult.getResults()) {
+        task = session.getDocument(new IdRef(id)).getAdapter(Task.class)
+        workflowRef = new IdRef(task.getProcessId())
+        if(!cache.computeIfAbsent(workflowRef, session.&exists)) {
+            log.info("removing orphan task " + id)
+            if(!dryRun) {
```
troger, line 24:
```suggestion
            if (!dryRun) {
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
troger, line 23: You are always logging this, dry run or not, it could be misleading while reading the logs :)

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
kevinleturc: `500` is the default we put everywhere, it looks like a good default.

I think we always increase it when it comes to production.

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy
NourNuxeo: Added a removedCount instead of bucketCount
I log at the end:
```log.info("removed {} out of {} tasks", removedCount, documentCount)```

File: server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy (hunk 9 above)
Comment (NourNuxeo) [reply server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy:23]: Then I will do 2 different logs "would remove..."

File: server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy (hunk 12 above)
Comment (troger) [reply server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy]: Name it `taskCount` maybe so? :p

Comment (NourNuxeo) [reply server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/taskGC.groovy]: there you go

File: server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy [hunk 14]
```diff
@@ -0,0 +1,39 @@
...
+            }
+        }
+    }
+    scrollResult = session.scroll(scrollResult.getScrollId())
+    TransactionHelper.commitOrRollbackTransaction()
+    TransactionHelper.startTransaction()
+}
+
+log.info("Finished deleting orphan tasks")
+log.info("removed {} out of {} tasks", removedCount, taskCount)
```
troger, line 39: You probably could merge both logs...
```
log.info("End deleting orphan tasks, removed {} out of {} tasks", ...)
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
```diff
@@ -0,0 +1,39 @@
...
+log.info("Starting deleting orphan tasks...")
+while (scrollResult.hasResults()) {
+    taskCount += scrollResult.getResults().size()
+    cache = new HashMap<IdRef>()
+    for (id in scrollResult.getResults()) {
+        task = session.getDocument(new IdRef(id)).getAdapter(Task.class)
+        workflowRef = new IdRef(task.getProcessId())
+        if (!cache.computeIfAbsent(workflowRef, session.&exists)) {
+            if (dryRun) {
+                log.info("would remove orphan task: {}", id)
```
troger, line 25:
```suggestion
                log.info("Would remove orphan task: {}", id)
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
```diff
@@ -0,0 +1,39 @@
...
+            } else {
+                log.info("removing orphan task: {}", id)
```
troger, line 27:
```suggestion
                log.info("Removing orphan task: {}", id)
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
```diff
@@ -0,0 +1,39 @@
...
+def dryRun = true
+def session = Context.getCoreSession()
+def ScrollResult<String> scrollResult = session.scroll(nxql, 500, 60)
+def removedCount = 0
+def taskCount = 0
+
+log.info("Starting deleting orphan tasks...")
+while (scrollResult.hasResults()) {
+    taskCount += scrollResult.getResults().size()
+    cache = new HashMap<IdRef>()
```
troger, line 19:
```suggestion
    def cache = new HashMap<IdRef>()
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
```diff
@@ -0,0 +1,39 @@
...
+    for (id in scrollResult.getResults()) {
+        task = session.getDocument(new IdRef(id)).getAdapter(Task.class)
```
troger, line 21:
```suggestion
        def task = session.getDocument(new IdRef(id)).getAdapter(Task.class)
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
```diff
@@ -0,0 +1,39 @@
...
+        workflowRef = new IdRef(task.getProcessId())
```
troger, line 22:
```suggestion
        def workflowRef = new IdRef(task.getProcessId())
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
```diff
@@ -0,0 +1,39 @@
+import org.apache.logging.log4j.LogManager
+import org.nuxeo.ecm.platform.task.Task
+import org.nuxeo.ecm.core.api.IdRef
+import org.nuxeo.ecm.core.api.ScrollResult
+import org.nuxeo.runtime.transaction.TransactionHelper
+import java.util.HashMap
+
+def log = LogManager.getLogger("taskGC.groovy")
```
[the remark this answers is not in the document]
NourNuxeo: Yes, since it's a String and doesn't fail, I missed that one !

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
NourNuxeo: Took the opportunity to declare cache in a groovy fashion and spare the Hashmap import

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
```diff
@@ -0,0 +1,37 @@
...
+            } else {
+                log.info("Removing orphan task: {}", id)
+                session.removeDocument(new IdRef(id))
+                removedCount += 1
+            }
+        }
+    }
+    scrollResult = session.scroll(scrollResult.getScrollId())
+    TransactionHelper.commitOrRollbackTransaction()
+    TransactionHelper.startTransaction()
```
troger, line 34: Shouldn't we re-start the transaction *before* doing a new scroll?

NourNuxeo: This is how it's done in `org.nuxeo.ecm.core.BatchFinderWork.work()` lines 88-90
Shouldn't the scroll be a snapshot ? Then it shouldn't be affected by the transaction ?

NourNuxeo: Tested it with a 1 bucket size and 2 tasks. Seems to be ok but in doubt i swapped the lines

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
```diff
@@ -0,0 +1,37 @@
...
+def dryRun = true
+def session = Context.getCoreSession()
+def ScrollResult<String> scrollResult = session.scroll(nxql, 500, 60)
+def removedCount = 0
+def taskCount = 0
+
+log.info("Starting deleting orphan tasks...")
+while (scrollResult.hasResults()) {
+    taskCount += scrollResult.getResults().size()
+    def cache = [:]
```
troger, line 18: Should be outside the `while` as we may want to reuse the cache when scrolling.

NourNuxeo: doh

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
```diff
@@ -0,0 +1,37 @@
...
+def log = LogManager.getLogger("deleteOrphanTasks.groovy")
+def nxql = "SELECT * FROM Document WHERE ecm:mixinType = 'Task'"
+def dryRun = true
+def session = Context.getCoreSession()
+def ScrollResult<String> scrollResult = session.scroll(nxql, 500, 60)
+def removedCount = 0
+def taskCount = 0
+def cache = [:]
+
+log.info("Starting deleting orphan tasks...")
```
ataillefer, line 16:
```suggestion
log.info("Start deleting orphan tasks...")
```

server/nuxeo-nxr-server/src/main/resources/templates/common-base/client/scripts/deleteOrphanTasks.groovy
```diff
@@ -0,0 +1,37 @@
...
+while (scrollResult.hasResults()) {
+    taskCount += scrollResult.getResults().size()
```
ataillefer, line 18: `scrollResult.getResults()` could be in a variable

PR #4015, opened 2020-05-11
NXP-29086: Support proper subclassing of Document JSON writer marshaller

michaelva:


modules/core/nuxeo-core-io/src/main/java/org/nuxeo/ecm/core/io/marshallers/json/document/DocumentPropertyJsonWriter.java
```diff
@@ -98,7 +98,7 @@
      */
     public static final String OMIT_PHANTOM_SECURED_PROPERTY = "omitPhantomSecuredProperty";
 
-    private static final Log log = LogFactory.getLog(DocumentPropertyJsonWriter.class);
+    protected static final Log log = LogFactory.getLog(DocumentPropertyJsonWriter.class);
```
kevinleturc, line 101: The logger should remain private, otherwise, the class extending this one could use it to log with inappropriate fqdn.

michaelva: ok. I reverted it to private.

PR #4027, opened 2020-05-14
NXP-29098: new migration object

NourNuxeo:


modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
```diff
@@ -18,6 +18,10 @@
  */
 package org.nuxeo.runtime.migration;
 
+import java.util.List;
+
+import javax.resource.spi.IllegalStateException;
```
efge, line 23: You shouldn't use this exception from `javax.resource.spi`.

NourNuxeo: I'm a bit lost about what to use as I can't access NuxeoException. I put RuntimeException wdyt ?

efge: `java.lang.IllegalStateException` would be ok. Or use `org.nuxeo.runtime.RuntimeServiceException` if you want a nuxeo-specific exception but not for the signature as it's too generic.

efge: If no caller is really expected to catch the exception and do special treatment, don't do anything in the signature and just throw `RuntimeServiceException` or `IllegalStateException` from the code in question.

NourNuxeo: Thanks!

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/Migration.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public String toString() {
+        String[] stepsArray = steps.entrySet()
+                                   .stream()
+                                   .map(e -> "'" + e.getKey() + "' : " + "'" + e.getValue().getDescription() + "'")
+                                   .toArray(String[]::new);
+        return "Migration: [id: '" + id + "', description: '" + description + "', label: '" + label + "', status: '"
+                + status.getState() + "', steps: " + Arrays.toString(stepsArray) + "]";
+    }
```
kevinleturc, line 79: Too expensive for a `toString` imho.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationConstants.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Contributors:
+ *     Nour AL KOTOB
+ */
+package org.nuxeo.runtime.migration;
+
+/**
+ * @since 11.1
+ */
+public class MigrationConstants {
```
kevinleturc, line 24: + `final`

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationConstants.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public static final String MIGRATION_ENTITY_TYPE = "migration";
+
+    public static final String MIGRATION_ID = "id";
+
+    public static final String MIGRATION_DESCRIPTION = "description";
+
+    public static final String MIGRATION_DESCRIPTION_LABEL = "description.label";
```
kevinleturc, line 32: Isn't a problem for javascript when it comes to reading this property (due to `.`)?

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
```diff
@@ -234,4 +237,28 @@ public long getProgressTotal() {
...
+    /**
+     * @since 11.1
+     */
+    List<Migration> getMigrations();
+
+    /**
+     * @throws IllegalStateException if the migration doesn't have exactly one available step to run.
+     * @since 11.1
+     */
+    void run(String migrationId) throws IllegalArgumentException;
```
kevinleturc, line 254: Is it possible to move it next to `runStep`, ideally before it, in order to gather related methods.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
```diff
@@ -234,4 +237,28 @@ public long getProgressTotal() {
...
+
+    /**
+     * Waits for completion of all migration steps.
+     *
+     * @return {@code true} if the migration executor has no more threads running {@code false} otherwise
+     * @since 11.1
+     */
+    boolean await(Duration duration) throws InterruptedException;
```
kevinleturc, line 262: Not sure we want this API, we don't have migration to wait during tests, and current migration tests don't need it.

Such APIs are a pain when they become available to the world. Here you may want it for tests, but later someone will ask you to make it work in a Nuxeo Cluster... Because it's public on an interface.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -529,4 +533,55 @@ protected void atomic(String id, Consumer<KeyValueStore> consumer) {
...
+
+    @Override
+    public List<Migration> getMigrations() {
+        var descriptors = getDescriptors(XP_CONFIG);
+        return descriptors.stream().map(d -> getMigration(d.getId())).collect(Collectors.toList());
+    }
+
+    @Override
+    public void run(String migrationId) {
+        if(getStatus(migrationId) == null) {
```
kevinleturc, line 551: Format

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -529,4 +533,55 @@ protected void atomic(String id, Consumer<KeyValueStore> consumer) {
...
+            probeAndSetState(migrationId);
+        }
+        MigrationDescriptor descriptor = getDescriptor(XP_CONFIG, migrationId);
+        Set<MigrationStepDescriptor> availableSteps = getAvailableSteps(descriptor);
+        if(availableSteps.size() != 1) {
```
kevinleturc, line 556: Format

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
```diff
@@ -234,4 +237,28 @@ public long getProgressTotal() {
...
+     * @since 11.1
+     */
+    Migration getMigration(String id);
+
+    /**
+     * @since 11.1
+     */
+    List<Migration> getMigrations();
+
+    /**
```
kevinleturc, line 250: Is it possible to add javadoc to explain what this method does?

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -529,4 +533,55 @@ protected void atomic(String id, Consumer<KeyValueStore> consumer) {
...
+    @Override
+    public List<Migration> getMigrations() {
+        var descriptors = getDescriptors(XP_CONFIG);
+        return descriptors.stream().map(d -> getMigration(d.getId())).collect(Collectors.toList());
+    }
+
+    @Override
+    public void run(String migrationId) {
+        if(getStatus(migrationId) == null) {
+            probeAndSetState(migrationId);
```
kevinleturc, line 552: This method is synchronous and could take a lot of time as explained in its Javadoc. This could be problematic when using it through REST.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -529,4 +533,55 @@ protected void atomic(String id, Consumer<KeyValueStore> consumer) {
...
+
+    protected Set<MigrationStepDescriptor> getAvailableSteps(MigrationDescriptor descriptor) {
+        MigrationStatus status = getStatus(descriptor.getId());
+        return descriptor.getSteps()
+                         .entrySet()
+                         .stream()
+                         .filter(e -> e.getValue().getFromState().equals(status.getState()))
+                         .map(Entry::getValue)
+                         .collect(Collectors.toSet());
+    }
```
kevinleturc, line 571:
```java
MigrationStepDescriptor step = descriptor.getSteps().values().stream().filter(v -> status.getState().equals(v.getFromState())).collect(Collectors.reducing((a, b) -> null)).orElseThrow(() -> new IllegalStateException(String.format("Migration %s needs to have exactly one step to run", migrationId)));
```

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -529,4 +533,55 @@ protected void atomic(String id, Consumer<KeyValueStore> consumer) {
...
+        for (;;) {
+            if (executor.getActiveCount() == 0) {
+                return true;
+            }
+            Thread.sleep(100);
+            if (deadline < System.nanoTime()) {
+                return false;
+            }
+        }
+    }
```
kevinleturc, line 585: See my comment on the interface.

For the cluster aspect, if you don't reach the node doing the migration, you won't await I think.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
NourNuxeo: I implemented it because in this new test I need to wait https://github.com/nuxeo/nuxeo/blob/c0c77d054898ec7a59bc35b08a50abf06822c8b5/modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/migration/test/TestMigrationService.java#L72

First when I saw `nextTransaction()` was not helping I did `Thread.sleep(1000)` for the test to pass. Otherwise it will fail. Shall I get back to this solution ?

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
NourNuxeo: Yes that may be why when I run and directly probe, I get an old result if I don't manually wait

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/migration/test/TestMigrationService.java
```diff
@@ -0,0 +1,82 @@
...
+import org.nuxeo.runtime.test.runner.Features;
+import org.nuxeo.runtime.test.runner.FeaturesRunner;
+import org.nuxeo.runtime.test.runner.TransactionalFeature;
+
+/**
+ * @since 11.1
+ */
+@RunWith(FeaturesRunner.class)
+@Features({ CoreFeature.class, MigrationFeature.class, TransactionalFeature.class })
+public class TestMigrationService {
```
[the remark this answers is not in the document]
troger: Because you added `MigrationFeature` to `nuxeo-core-test` probably :p

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/migration/test/TestMigrationService.java
```diff
@@ -0,0 +1,82 @@
...
+    protected TransactionalFeature txFeature;
+
+    @Inject
+    protected MigrationService migrationService;
+
+    @Test
+    public void testGetMigrations() {
+        var migrations = migrationService.getMigrations();
+        assertEquals(2, migrations.size());
+    }
```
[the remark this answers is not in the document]
troger: Agree, the dummy one is here for that.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationConstants.java
troger: 👍 should be `descriptionLabel` as what we do on other writers.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
```diff
@@ -234,4 +237,28 @@ public long getProgressTotal() {
...
 
+    /**
+     * @since 11.1
+     */
+    Migration getMigration(String id);
+
+    /**
+     * @since 11.1
+     */
+    List<Migration> getMigrations();
```
troger, line 248: Missing javadoc, and worth moving them before `MigrationStatus getStatus(String id);`

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
troger: If we don't want to expose it publicly, I would put that method only on the implementation class stating it's for test purposes, and use it in the test.
Does that sound good @kevinleturc ?

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
troger: We agreed to have the `#run` method probing the migration to have only one call to do from the outside. However, like in the `#probeAndSetState` method Javadoc, it should be stated clearly in the `#run` method Javadoc.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -529,4 +533,55 @@ protected void atomic(String id, Consumer<KeyValueStore> consumer) {
         throw new RuntimeException("Cannot lock for write migration: " + id + ", already locked: " + currentLock);
     }
 
+    @Override
+    public Migration getMigration(String id) {
+        MigrationDescriptor descriptor = getDescriptor(XP_CONFIG, id);
+        return new Migration(id, descriptor.getDescription(), descriptor.getDescriptionLabel(), getStatus(id),
+                descriptor.getSteps());
```
troger, line 540: The steps should be filtered to only return the available steps for the given migration, like what is done on the JSF UI (see https://github.com/nuxeo/nuxeo-jsf-ui/blob/master/code/nuxeo-admin-center-jsf/src/main/java/org/nuxeo/ecm/admin/MigrationAdminBean.java#L49 for instance).

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
NourNuxeo: Then I would call the method `get{Only | Single}AvailableStep` ?

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
NourNuxeo: followed https://github.com/nuxeo/nuxeo/pull/4027#discussion_r428223297
The only public method left is `MigrationFeature#initialize` but shouldn't be a problem since it's only used for test

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationConstants.java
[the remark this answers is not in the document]
NourNuxeo: Moving it in the Management REST API project

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
NourNuxeo: MigrationFeature#Initialize calls the await method. If it is protected, then we can't call it

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
NourNuxeo: After inputing it, I have the following problem https://github.com/nuxeo/nuxeo/pull/4027#discussion_r430345263

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
NourNuxeo: there is actually another case where we need to get all available steps: https://github.com/nuxeo/nuxeo/pull/4027#discussion_r428232972
I think staying in a method that doesn't fail if it finds several steps is more flexible. What do you think ?

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
```diff
@@ -234,4 +237,28 @@ public long getProgressTotal() {
...
+    /**
+     * @throws IllegalStateException if the migration doesn't have exactly one available step to run.
+     * @since 11.1
+     */
+    void run(String migrationId) throws IllegalArgumentException;
+
+    /**
+     * Waits for completion of all migration steps.
+     *
+     * @return {@code true} if the migration executor has no more threads running {@code false} otherwise
```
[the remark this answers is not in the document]
NourNuxeo: added some html tags and extra infos: ```/**
     * Waits for completion of all migration steps.<br/>
     * <b>ONLY FOR TEST PURPOSE</b>
     * <ul>
     * <li>Not suitable for use in a cluster environment.</li>
     * <li>Not declared in the API.</li>
     * </ul>
     *
     * @return {@code true} if the migration executor has no more threads running, {@code false} otherwise
     * @since 11.1
     */```

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/migration/test/TestMigrationService.java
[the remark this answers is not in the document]
kevinleturc: Because I wanted to test a real migration, is it the case there?

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
kevinleturc: Sounds good to have a protected method on the implementation only.

Not sure I, the end if we want the waiter, we control the migration start/stop, to discuss.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
kevinleturc: Don't we want to fail if there'are several steps in this case? I'm not sure the returned `Set` will be useful for the other case.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
NourNuxeo: I made the `getAvailableSteps` return a map so I can give it here in the constructor and leverage `getAvailableSteps`

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
NourNuxeo: Now we also return null when the id is not found

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationService.java
NourNuxeo: if it is protected we can't call it in the MigrationFeature#initialize

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
NourNuxeo: Yes we want to fail.
I changed to a map for the other case, then it can fit the argument.

Otherwise I will have to make 2 methods, 1 to get a single step and fail otherwise and another one to get all available steps

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
NourNuxeo: now probeAndRun with javadoc:
```
     /**
     * Probes the migration state if necessary and runs a migration.
     * THIS METHOD MAY TAKE A WHILE as it may call {@link #probeAndSetState(String)}.
     *
     * @since 11.1
     */
```

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/migration/test/TestMigrationService.java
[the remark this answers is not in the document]
NourNuxeo: Works fine now in `nuxeo-runtime-migration` (no more MigrationFeature#await)

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/Migration.java
```diff
@@ -0,0 +1,70 @@
...
+
+    protected String id;
+
+    protected String description;
+
+    protected String label;
+
+    protected MigrationStatus status;
+
+    protected Map<String, MigrationStepDescriptor> steps;
```
troger, line 39: Should be a simple `List`.

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/DummyMigrator.java
```diff
@@ -0,0 +1,43 @@
...
+import org.nuxeo.runtime.transaction.TransactionHelper;
+
+/**
+ * @since 11.1
+ */
+public class DummyMigrator implements Migrator {
+
+    protected String state;
+
+    public void run(String step, MigrationContext migrationContext) {
```
troger, line 32: Missing `@Override`.

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/DummyMigrator.java
```diff
@@ -0,0 +1,43 @@
...
+        TransactionHelper.commitOrRollbackTransaction();
+        TransactionHelper.startTransaction();
```
troger, line 34: Why is it needed?

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/TestMigrationService.java
```diff
@@ -0,0 +1,93 @@
...
+import org.nuxeo.runtime.test.runner.RuntimeFeature;
+
+/**
+ * @since 11.1
+ */
+@RunWith(FeaturesRunner.class)
+@Features(RuntimeFeature.class)
+@Deploy("org.nuxeo.runtime.kv")
+@Deploy("org.nuxeo.runtime.cluster")
+@Deploy("org.nuxeo.runtime.migration:OSGI-INF/migration-service.xml")
```
troger, line 45: You could just deploy `org.nuxeo.runtime.migration`.

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/TestMigrationService.java
```diff
@@ -0,0 +1,93 @@
...
+               .until(() -> migrationService.getMigration(dummy).getStatus().getState().equals("after"));
+        try {
+            migrationService.probeAndRun(dummy);
+            fail("should fail");
+        } catch (java.lang.IllegalStateException e) {
+            assertEquals("Migration dummy-migration needs to have exactly one step to run", e.getMessage());
+        }
+    }
+
+    protected void checkMigration(Migration migration) {
```
troger, line 84:
```suggestion
    protected void assertDummyMigration(Migration actual) {
```

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/TestMigrationService.java
```diff
@@ -0,0 +1,93 @@
...
+    @Test
+    public void testGetMigrations() {
+        var migrations = migrationService.getMigrations();
+        assertEquals(1, migrations.size());
+        checkMigration(migrations.get(0));
+    }
+
+    @Test
+    public void testUnknownMigration() {
+        assertEquals(null, migrationService.getMigration("fake"));
```
troger, line 66:
```suggestion
        assertNull(migrationService.getMigration("fake"));
```

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/TestMigrationService.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Test
+    public void testRunMigration() {
```
troger, line 70:
```suggestion
    public void testProbeAndRunMigration() {
```

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/TestMigrationService.java
```diff
@@ -0,0 +1,93 @@
...
+        String dummy = "dummy-migration";
+        assertEquals("before", migrationService.getMigration(dummy).getStatus().getState());
+        migrationService.probeAndRun(dummy);
+        await().atMost(1, SECONDS)
+               .until(() -> migrationService.getMigration(dummy).getStatus().getState().equals("after"));
```
troger, line 75:
```suggestion
               .until(() -> "after".equals(migrationService.getMigration(dummy).getStatus().getState()));
```

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/TestMigrationService.java
```diff
@@ -0,0 +1,93 @@
...
+        try {
```
troger, line 76: New line to separate both tests (improve the readability :)).

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/TestMigrationService.java
```diff
@@ -0,0 +1,93 @@
...
+            migrationService.probeAndRun(dummy);
+            fail("should fail");
+        } catch (java.lang.IllegalStateException e) {
```
troger, line 79:
```suggestion
        } catch (IllegalStateException e) {
```

modules/runtime/nuxeo-runtime-migration/src/test/resources/OSGI-INF/dummy-migration.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<component name="org.nuxeo.runtime.test.migration.dummy" version="1.0">
```
troger, line 2:
```suggestion
<component name="org.nuxeo.runtime.migration.test.dummy" version="1.0">
```
?

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/TestMigrationService.java
NourNuxeo: that's a rare request from you !

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/DummyMigrator.java
NourNuxeo: remanence of me trying to synch things

modules/runtime/nuxeo-runtime-migration/src/test/resources/OSGI-INF/dummy-migration.xml
NourNuxeo: yes more consistent with bundle symbolic name, project name etc...

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -470,8 +471,7 @@ public void runStep(String id, String step) {
                 log.error("Exception during execution of step: {} for migration: {}", step, id, t);
             }
             // after the migrator is finished, change state, except if shutdown is requested or exception
-            String state = t != null || migrationContext.isShutdownRequested()
-                    ? stepDescr.fromState
+            String state = t != null || migrationContext.isShutdownRequested() ? stepDescr.fromState
```
NourNuxeo, line 474: didn't feel worth a separate commit. Unless ?...

kevinleturc: It's OK to me, no need for a cleanup commit.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -529,4 +532,47 @@ protected void atomic(String id, Consumer<KeyValueStore> consumer) {
...
+            return null;
+        }
+        return new Migration(id, descriptor.getDescription(), descriptor.getDescriptionLabel(), getStatus(id),
+                getAvailableSteps(descriptor));
+    }
+
+    @Override
+    public List<Migration> getMigrations() {
+        var descriptors = getDescriptors(XP_CONFIG);
+        return descriptors.stream().map(d -> getMigration(d.getId())).collect(Collectors.toList());
```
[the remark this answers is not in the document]
kevinleturc: This change has not been pushed.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -529,4 +529,35 @@ protected void atomic(String id, Consumer<KeyValueStore> consumer) {
...
+    @Override
+    public void probeAndRun(String id) {
+        MigrationStatus status = getStatus(id);
+        if (status == null) {
+            probeAndSetState(id);
+        }
+        var steps = Migration.from(getDescriptor(XP_CONFIG, id), status).getSteps();
+        if (steps.size() != 1) {
+            throw new IllegalStateException(String.format(
+                    "Migration %s must have only one runnable step from state: %s", id, status.getState())); // NOSONAR
```
kevinleturc, line 558:
```suggestion
                    "Migration: %s must have only one runnable step from state: %s", id, status.getState())); // NOSONAR
```

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationStep.java
```diff
@@ -0,0 +1,65 @@
...
+        return toState;
+    }
+
+    public MigrationStep(MigrationStepDescriptor descriptor) {
+        this.id = descriptor.getId();
+        this.description = descriptor.getDescription();
+        this.descriptionLabel = descriptor.getDescriptionLabel();
+        this.fromState = descriptor.getFromState();
+        this.toState = descriptor.getToState();
+    }
```
troger, line 64: Maybe worth doing the same thing as for `Migration`, a static `from` and only simple params in the constructor.

NourNuxeo: here you go!

kevinleturc:
I removed myself from reviewers as I'm getting notified every day whereas the PR doesn't seem to be ready (open discussion, change requested, etc)

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -529,4 +530,36 @@ protected void atomic(String id, Consumer<KeyValueStore> consumer) {
...
+    /**
+     * @implNote Runs a migration if it has exactly one available step to run.
+     */
+    @Override
+    public void probeAndRun(String id) {
+        MigrationStatus status = getStatus(id);
+        if (status == null) {
+            probeAndSetState(id);
+        }
+        var steps = Migration.from(getDescriptor(XP_CONFIG, id), status).getSteps();
```
kevinleturc, line 557: Shouldn't you get status in case you probe the state?

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/TestMigrationService.java
```diff
@@ -0,0 +1,116 @@
...
+    public void testProbeAndRunMigration() {
+        // Migration with 2 sequential steps: before to after, after to reallyAfter
+        String dummyMigration = "dummy-migration";
+        assertEquals("before", migrationService.getMigration(dummyMigration).getStatus().getState());
+        migrationService.probeAndRun(dummyMigration);
+        await().atMost(1, SECONDS)
+               .until(() -> "after".equals(migrationService.getMigration(dummyMigration).getStatus().getState()));
+        migrationService.probeAndRun(dummyMigration);
+        await().atMost(1, SECONDS)
+        .until(() -> "reallyAfter".equals(migrationService.getMigration(dummyMigration).getStatus().getState()));
```
kevinleturc, line 79: Format

modules/runtime/nuxeo-runtime-migration/src/test/java/org/nuxeo/runtime/migration/TestMigrationService.java
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            migrationService.probeAndRun(dummyMigration);
+            fail("should fail");
```
kevinleturc, line 82: Please a meaningful comment or leave it empty.

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
NourNuxeo: The name was changed to probeAndRun later but it was just defined as run and meant to do so even if the state is unknown (so we probe then).
I think it was renamed probeAndRun to make the call to probeAndSetState more explicit as it has this warning:
> * THIS METHOD MAY TAKE A WHILE as it needs to get information from persistent storage.

@troger might remember better than me

troger: @kevinleturc is pointing that if `status` is null, you call `probeAndSetState(id);` but never get back the updated status, so status is still `null`.

kevinleturc: Exactly

NourNuxeo: I made a systematic call to `probeAndSetState` then `getStatus`

troger: No no... why probing if there is no need to probe?

```java
MigrationStatus status = getStatus(id);
if (status == null) {
    probeAndSetState(id);
    status = getStatus(id);
}
...
```

It is stated in your method's Javadoc:
> Probes the migration state if necessary and runs a migration.

NourNuxeo: alright

PR #4042, opened 2020-05-15
NXP-29118: fix Redis SSL support

fmunch:
SSL support for Redis (non HA) does not seem to be complete. The [ping attempt](https://github.com/nuxeo/nuxeo/blob/v11.1.24/modules/runtime/nuxeo-runtime-redis/src/main/java/org/nuxeo/ecm/core/redis/RedisServerDescriptor.java#L100) uses SSL as expected, but the [pool creation](https://github.com/nuxeo/nuxeo/blob/v11.1.24/modules/runtime/nuxeo-runtime-redis/src/main/java/org/nuxeo/ecm/core/redis/RedisServerDescriptor.java#L113) ignores SSL.

This issues prevents the use of Azure Redis Cache for example (unless SSL is disabled server-side, of course).

A backport to 10.10 would be appreciated, I can open another PR on that branch if needed.


efge:
Thanks, I created [NXP-29118 ](https://jira.nuxeo.com/browse/NXP-29118), could you update the commit with the ticket number?

efge:
Yes we can backport, please provide a 10.10 PR.

fmunch:
Commit message updated.

efge:
Thanks

PR #4065, opened 2020-05-25
NXP-29138: add an attribute to bypass the security check in file importer

pleday:
[ticket: NXP-29138]

PR #4066, opened 2020-05-25
NXP-29138: add an attribute to bypass the security check in file importer

pleday:
…mporter

[ticket: NXP-29138]

### DIFF
--- a/modules/platform/nuxeo-platform-filemanager-api/src/main/java/org/nuxeo/ecm/platform/filemanager/api/FileImporterContext.java
+++ b/modules/platform/nuxeo-platform-filemanager-api/src/main/java/org/nuxeo/ecm/platform/filemanager/api/FileImporterContext.java
@@ -46,6 +46,8 @@ public class FileImporterContext {
 
     protected boolean persistDocument;
 
+    protected boolean byPassSecurityCheck;
+
     public static Builder builder(CoreSession session, Blob blob, String parentPath) {
         return new Builder(session, blob, parentPath);
     }
@@ -59,6 +61,7 @@ public class FileImporterContext {
         excludeOneToMany = builder.excludeOneToMany;
         fileName = StringUtils.defaultIfEmpty(builder.fileName, blob.getFilename());
         persistDocument = builder.persistDocument;
+        byPassSecurityCheck = builder.byPassSecurityCheck;
     }
 
     public CoreSession getSession() {
@@ -93,6 +96,13 @@ public class FileImporterContext {
         return persistDocument;
     }
 
+    /**
+     * @since 11.1
+     */
+    public boolean isByPassSecurityCheck() {
+        return byPassSecurityCheck;
+    }
+
     public static class Builder {
 
         protected final CoreSession session;
@@ -110,6 +120,8 @@ public class FileImporterContext {
         protected boolean excludeOneToMany;
 
         protected boolean persistDocument = true;
+
+        protected boolean byPassSecurityCheck;
 
         public Builder(CoreSession session, Blob blob, String parentPath) {
             this.session = session;
@@ -165,6 +177,17 @@ public class FileImporterContext {
          */
         public Builder persistDocument(boolean persistDocument) {
             this.persistDocument = persistDocument;
+            return this;
+        }
+
+        /**
+         * Whether to bypass the security check.
+         * <p>
+         * Defaults to {@code false}.
+         * @since 11.1
+         */
+        public Builder byPassSecurityCheck(boolean byPassSecurityCheck) {
+            this.byPassSecurityCheck = byPassSecurityCheck;
             return this;
         }
 
--- a/modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/AbstractFileImporter.java
+++ b/modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/AbstractFileImporter.java
@@ -214,7 +214,10 @@ public abstract class AbstractFileImporter implements FileImporter {
         if (targetDocType == null) {
             targetDocType = getDefaultDocType();
         }
-        doSecurityCheck(session, path, targetDocType);
+
+        if (!context.isByPassSecurityCheck()) {
+            doSecurityCheck(session, path, targetDocType);
+        }
 
         Blob blob = context.getBlob();
         String filename = FileManagerUtils.fetchFileName(context.getFileName());


efge:
Same comments as for the 10.10 branch (#4065)

troger:
There is kinda the same check done in `FileManagerService#defaultCreateFolder`, don't we want the same thing there?

efge:
> There is kinda the same check done in `FileManagerService#defaultCreateFolder`, don't we want the same thing there?

Yes that's a good point

pleday:
> There is kinda the same check done in `FileManagerService#defaultCreateFolder`, don't we want the same thing there?

correct me if i am wrong but this attribute sets if the folder checks the "allowed sub types".
Our problem is different, we want our folders to always check the "allowed sub types" if defined but some of our doctypes can create document children with a server treatment but we cannot do this if the folder parent has "allowed sub types". 
Example :
- the user creates a folder with allowed sub types "doctype1" so this folder is only for this kind of documents
- the doctype1 can create children of type picture or file with a server API method

now we can't do this because of parent sub types restriction
the pr i did allowes server methods to bypass the "allowed sub types"

troger:
You may want the same thing when creating new folder through the `FileManager`. 
Assuming the default folder importer creates folders with `FooFolder` type, but you wan't to create it under a `Folder` type that allows only `Folder` as sub types, you could use `byPassSecurityCheck` to create it anyway.

pleday:
After a too long pause, i resume the issue by reformulate what we want by copying a translated version of the original ticket :

> My understanding of the "local configuration" proposed by Nuxeo, is to guide the user in the documents created from a directory, this guidance is ergonomic. The creation possibilities are described via the 'accepted children / contianer type' of the standard documents. Then for each directory, it is possible to reduce these possibilities according to the local configuration.

> Now from the 'code' point of view, it is quite possible to create a document in any directory. This possibility is not constrained by the local configuration, nor by the "accepted children / conainer type". This is done using the standard nuxeo API. The user just needs to have the necessary ACLs.

> We use this principle to create technical documents storing information without the user being aware of it. As the documents are technical, they should not be visible in the creation window.

> In the case of this ticket, the user adds attachments, and we store them as child and typed documents (video, image, file).
 

> The service provided by Nuxeo AbstractFileImporter perfectly meets our needs, only the fact that it blocks the creation according to the local configuration is a problem for us.

> I think there should be no link between the two: the nuxeo APIs should allow all creations in all directories, and only the HMI should be impacted by the locale configuration.

> Our initial proposal is to add an option in the FileImporterContext context to disable the control over the local configuration.


Example :
- creation of a folder with subtype restriction on doctype1
- in this folder, creation of a doctype1 document (this doctype is folderish)
- on this document, with a backend treatment, we try to create a sub document with doctype2
--> with the inheritance of the restriction on the folder, the treatment failes because the folder accepts only doctype1 and not doctype2

To create the sub document, we use this java code :
`FileImporterContext context = FileImporterContext.builder(session, blob, parentFilePath)
                    .overwrite(override)
                    .persistDocument(true)
                    .build();
            documentFromBlob = fileManager.createOrUpdateDocument(context);`

I hope, i'm clear on my explanation.

ataillefer:
Hi @pleday,

Thanks for the explanation and sorry for the delay on our side.
We're going to take into account this pull request, complete it with a non regression test and merge it soon, as the related [JIRA issue](https://jira.nuxeo.com/browse/NXP-29138) is planned for the current iteration.

Concerning @troger's [comment](https://github.com/nuxeo/nuxeo/pull/4066#issuecomment-635421578) about `FileManagerService#defaultCreateFolder`, forget about it, it's not trivial to implement with the current code base.

About your suggested changes, we would do it in a slightly different way.
Instead of bypassing the whole code executed by `AbstractFileImporter#doSecurityCheck`, we would only bypass the part checking the subtypes, something like:
```
    /**
     * @since 11.3
     */
    protected void doSecurityCheck(CoreSession documentManager, String path, String typeName,
            FileImporterContext context) {
        // always perform the security checks
        PathRef containerRef = new PathRef(path);
        if (!documentManager.hasPermission(containerRef, READ_PROPERTIES)
                || !documentManager.hasPermission(containerRef, ADD_CHILDREN)) {
            throw new DocumentSecurityException("Not enough rights to create folder");
        }

        // perform the allowed subtypes check, unless bypassed
        if (!context.isBypassSubTypeCheck()) {
            TypeManager typeService = Framework.getService(TypeManager.class);
            DocumentModel container = documentManager.getDocument(containerRef);
            Type containerType = typeService.getType(container.getType());
            if (containerType == null) {
                return;
            }

            if (!typeService.isAllowedSubType(typeName, container.getType(), container)) {
                throw new NuxeoException(String.format("Cannot create document of type %s in container with type %s",
                        typeName, containerType.getId()));
            }
        }
    }
```
This would allow you to bypass the **subtype** constraint, without bypassing the actual **security** check on permissions to add a child.
Would this be OK for your functional need?

pleday:
Thank you for your answer
One thing i don't said is that the users who launch the backend treatment has limited right so we have to skip the permissions check too.

ataillefer:
OK so it's a wider need then what seemed to be expressed initially.
Then, the following sentences are unclear to us:

> now we can't do this because of parent sub types restriction

> with the inheritance of the restriction on the folder, the treatment failes because the folder accepts only doctype1 and not doctype2

They suggest that the document creation would fail because of the subtype check:
```
        if (!typeService.isAllowedSubType(typeName, container.getType(), container)) {
            throw new NuxeoException(String.format("Cannot create document of type %s in container with type %s",
                    typeName, containerType.getId()));
        }
```
If you say that

> the users who launch the backend treatment has limited right so we have to skip the permissions check too.

then we would expect the following code to fail:
```
        if (!documentManager.hasPermission(containerRef, READ_PROPERTIES)
                || !documentManager.hasPermission(containerRef, ADD_CHILDREN)) {
            throw new DocumentSecurityException("Not enough rights to create folder");
        }
```
Do you have an extract of the stack trace to see which exception is actually thrown?

ataillefer:
We're also interested in the exact permissions granted to the users who launch the backend treatment, if you can provide them, thanks.

ataillefer:
Hi,

We've agreed internally on allowing to bypass the allowed subtype check only, using the following code:
```
FileImporterContext.builder(...)
        .bypassAllowedSubtypeCheck(true)
        .build()
```

About the security aspect, since you're using Java to invoke the document creation server-side, you can use `doPrivileged(CoreSession session, Function<CoreSession, R> function)` to run the code as a system user:
```
        documentFromBlob = CoreInstance.doPrivileged(session, s -> {
            FileImporterContext context = FileImporterContext.builder(s, blob, parentFilePath)
                                                             .overwrite(override)
                                                             .persistDocument(true)
                                                             .bypassAllowedSubtypeCheck(true)
                                                             .build();
            try {
                return fileManager.createOrUpdateDocument(context);
            } catch (IOException e) {
                throw new NuxeoException(e);
            }
        });
```
Note that the actual user triggering the code will be referenced as a creator in the dublincore metadata and the audit logs.

Closing the current pull request in favor of #4291 since we'retaking care of it.

pleday:
@ataillefer Thank you for your answer and for taking care of it, it's all good for us. Can you tell me an approximative delivery date for this code ?

ataillefer:
@pleday you're welcome.

About delivery dates, can you please check with the support team via the JIRA issue tracker?

PR #4072, opened 2020-05-25
NXP-29067: add the command line name to the created tmp directory

troger:
This would be useful when debugging to know which command line
has created the tmp directory

### DIFF
--- a/modules/platform/nuxeo-platform-convert/src/test/java/org/nuxeo/ecm/platform/convert/tests/TestCommandLineConverter.java
+++ b/modules/platform/nuxeo-platform-convert/src/test/java/org/nuxeo/ecm/platform/convert/tests/TestCommandLineConverter.java
@@ -21,7 +21,9 @@ package org.nuxeo.ecm.platform.convert.tests;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
+import java.io.File;
 import java.io.Serializable;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


modules/platform/nuxeo-platform-convert/src/test/java/org/nuxeo/ecm/platform/convert/tests/TestCommandLineConverter.java
```diff
@@ -56,6 +58,11 @@ public void testCommandLineConverter() throws Exception {
 
         Blob mainBlob = result.getBlob();
         assertEquals("hello.png", mainBlob.getFilename());
+
+        // NXP-29067: make sure the parent path is prefixed with the command line name
+        FileBlob fileBlob = ((FileBlob) mainBlob);
```
efge, line 63: Extra unneeded set of parentheses, and could be inlined in the line below anyway.

modules/platform/nuxeo-platform-convert/src/test/java/org/nuxeo/ecm/platform/convert/tests/TestCommandLineConverter.java
```diff
@@ -56,6 +58,11 @@ public void testCommandLineConverter() throws Exception {
...
+        String parentPath = fileBlob.getFile().getParent();
```
efge, line 64: Actually `getFile()` is part of the `Blob` interface so there's no need for a cast at all — unless you want to verify that it's actually a `FileBlob`.

modules/platform/nuxeo-platform-convert/src/test/java/org/nuxeo/ecm/platform/convert/tests/TestCommandLineConverter.java
troger: Code does not exist anymore 👍

PR #4073, opened 2020-05-25
NXP-29137: improve KeyValueStore.compareAndSet when inserting a new value

efge:


modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/jdbc/dialect/Dialect.java
```diff
@@ -1015,6 +1015,20 @@ public Array createArrayOf(int type, Object[] elements, Connection connection) t
     public abstract String getUpsertSql(List<Column> columns, List<Serializable> values, List<Column> outColumns,
             List<Serializable> outValues);
 
+    /**
+     * Gets the SQL string for an INSERT ... ON CONFLICT DO NOTHING expression.
```
bdelbosc, line 1019: on conflict of what? doing nothing means returns null SQL or skip the column or override?

PR #4074, opened 2020-05-25
NXP-29137: improve KeyValueStore.compareAndSet when inserting a new value (10.10)

efge:


efge:
T&P ok https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume-10.10/104/

PR #4075, opened 2020-05-26
NXP-28956: upgrade to H2 1.4.200

efge:


efge:
FYI I still need to do further commits on this because the new version of H2 reports more concurrent updates and we have to deal with them in different places.

efge:
T&P ok https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/845/

PR #4101, opened 2020-06-02
NXP-28758 / NXP-28370: version security

efge:
[NXP-28758](https://jira.nuxeo.com/browse/NXP-28758): Allow ACLs on versions
[NXP-28370](https://jira.nuxeo.com/browse/NXP-28370): Make version Read permission depend on live doc ReadVersion


efge:
More commits to come but I want a job to run to check the status.

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/model/BaseSession.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    /**
+     * Gets the ACP for the document (without any inheritance).
+     *
+     * @param doc the document
+     * @return the ACP
+     */
+    public abstract ACP getACP(Document doc);
```
NourNuxeo, line 139: it seems it was protected when it was in the implementations, is it worth a @since 11.1 ?

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/model/BaseSession.java
```diff
@@ -0,0 +1,173 @@
...
+                // TODO avoid duplicates
+                curAcl.addAll(acl);
+            } else {
+                newAcp.addACL(acl);
+            }
+        }
+        return newAcp;
+    }
+
+}
```
NourNuxeo, line 211: missing new line

NourNuxeo:
Sorry didn't see it's a draft, Only checked the first commit. I'll leave it there then.

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql-extensions/src/main/java/org/nuxeo/ecm/core/storage/sql/db/EmbeddedFunctions.java
```diff
@@ -132,82 +143,93 @@ public static boolean isAccessAllowed(Serializable id, Set<String> principals, S
...
+        }
+    }
+
+    protected static class RowInfo {
+
+        public Serializable parentId;
+
+        public boolean isVersion;
+
+        public Serializable versionableId;
```
NourNuxeo, line 196: maybe make the members final as they are meant to be set once per instantiation ?

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql-extensions/src/main/java/org/nuxeo/ecm/core/storage/sql/db/EmbeddedFunctions.java
```diff
@@ -132,82 +143,93 @@ public static boolean isAccessAllowed(Serializable id, Set<String> principals, S
...
+            Set<String> permissions) throws SQLException {
+        psAcl.setObject(1, id);
+        try (ResultSet rs = psAcl.executeQuery()) {
+            while (rs.next()) {
+                boolean grant = rs.getShort(1) != 0;
+                String permission = rs.getString(2);
+                String user = rs.getString(3);
+                if (principals.contains(user) && permissions.contains(permission)) {
+                    return grant;
+                }
```
NourNuxeo, line 230: Less clarity but maybe this would avoid getting a boolean before knowing if it's aimed at the user regarding the permission ?
I tried adding a comment not to lose too much readability:
```suggestion
                String permission = rs.getString(2);
                String user = rs.getString(3);
                if (principals.contains(user) && permissions.contains(permission)) {
                    return rs.getShort(1) != 0; // the access grant boolean value
                }
```

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql-extensions/src/main/java/org/nuxeo/ecm/core/storage/sql/db/H2Functions.java
```diff
@@ -55,15 +55,34 @@ public static boolean isInTreeString(Connection conn, String id, String baseId)
...
+     */
     public static boolean isAccessAllowedString(Connection conn, String id, String principals, String permissions)
             throws SQLException {
         return isAccessAllowed(conn, id, split(principals), split(permissions));
     }
 
+    /**
+     * @since 11.1
+     */
+    public static boolean isAccessAllowed2(Connection conn, String id, String principals, String permissions,
```
NourNuxeo, line 77: Since the signature is not the same maybe it could get rid off the 2 suffix ?
if the args are not splittable and end in single object collections the name would still be relevant.
Or is it because of the alias creation you need a distinct name ?
it seems you can define the arguments types` as in java.lang.Integer.parseInt(java.lang.String, int)`: https://www.h2database.com/html/commands.html#create_alias
Not sure the result would be pretty with a `Connection`, 3 `String` and 1 `boolean` though...

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql-extensions/src/main/java/org/nuxeo/ecm/core/storage/sql/db/EmbeddedFunctions.java
efge: That would mean writing a constructor, and not being able to update the fields independently which is how this is used today. So no, that would be much more verbosity.

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql-extensions/src/main/java/org/nuxeo/ecm/core/storage/sql/db/EmbeddedFunctions.java
efge: I prefer the clarity of getting all the values first, then acting on them. Also, this is just code that was moved.

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql-extensions/src/main/java/org/nuxeo/ecm/core/storage/sql/db/H2Functions.java
efge: Given that this is purely internal I prefer having a short and obviously unambiguous alias definition, so I won't change it.

efge:
Ready for review

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/model/BaseSession.java
```diff
@@ -0,0 +1,250 @@
...
+        public static VersionAclMode getConfiguration() {
+            if (!Framework.isInitialized()) {
+                // unit tests
+                return ENABLED;
+            }
+            ConfigurationService configurationService = Framework.getService(ConfigurationService.class);
+            Optional<String> val = configurationService.getString(VERSION_ACL_DISABLED_PROP);
+            if (!val.isPresent()) {
+                return ENABLED;
+            }
```
kevinleturc, line 81:
```suggestion
            String val = configurationService.getString(VERSION_ACL_DISABLED_PROP).orElse("false");
```
?

kevinleturc (approved):
Cool to have this ACP code factorized 💪.

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/model/BaseSession.java
efge: 👍

kevinleturc (approved):

troger (approved):

PR #4105, opened 2020-06-03
NXP-27786: rename DocumentMultivaluedProperty.addItem to make it easi…

michaelva:
…er to find in studio and make it consistant with the name of the opposite remove operation

PR #4108, opened 2020-06-03
NXP-27786: fix test after operation id change

atchertchian:
To fix tests after merge of #4105.

Note i would have expected new id to be "Document.AddEntryToMultivaluedProperty" instead of lower case to "to".


efge:
+1 on the `To` I don't know how we missed that

PR #4115, opened 2020-06-04
NXP-29154 Improve feedback on explorer actions

atchertchian:
T&P in progress at https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-atchertchian-master-2/98/, passed ok :heavy_check_mark:


modules/platform/nuxeo-apidoc-server/nuxeo-apidoc-webengine/src/main/java/org/nuxeo/apidoc/browse/Distribution.java
```diff
@@ -358,21 +335,27 @@ public Object doSaveExtended() throws NamingException, NotSupportedException, Sy
...
+            return getView("savedKO").arg("message", e.getMessage()).arg("source", source);
         }
         log.info("Snapshot saved.");
+
         if (tx != null && startedTx) {
             tx.commit();
         }
-        return getView("saved");
+        String redirectUrl = getContext().getBaseURL() + getPath();
+        log.debug("Path => " + redirectUrl);
```
kevinleturc, line 357: You can use log4j2 format feature to save some resources.
```suggestion
        log.debug("Path => {}",  redirectUrl);
```

Otherwise, we tend to use a colon follow by a space and directly is value, for example `The document: {docId} has been removed`. I don't know if we want to follow such pattern here?

modules/platform/nuxeo-apidoc-server/nuxeo-apidoc-webengine/src/main/resources/skin/resources/script/jquery.utils.js
```diff
@@ -0,0 +1,7 @@
+(function ($) {
+
+  $.fn.clickButton = function (element) {
+    element.classList.add('clicked');
+  }
+
+})(jQuery);
```
kevinleturc, line 7: It misses a new line at the end of this file.

modules/platform/nuxeo-apidoc-server/nuxeo-apidoc-webengine/src/main/java/org/nuxeo/apidoc/browse/Distribution.java
atchertchian: Indeed this is old logging, I only moved that line when factoring out duplicated code, and did not pay much attention to it, but i do not see the point of computing the redirectUrl just for logging since we're not redirecting to it anyway.
So i will just remove it :+1:

PR #4116, opened 2020-06-04
Feature NXP-29190 add profile mechanism

kevinleturc:
Open in draft to get the JX statuses + early review if any.


modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
```diff
@@ -112,7 +113,13 @@
     /** @since 11.1 */
     public static final String NUXEO_ENVIRONMENT = System.getenv("NUXEO_ENVIRONMENT");
 
-    /** @since 6.0 */
+    /** @since 11.1 */
+    public static final String NUXEO_PROFILES = System.getenv().getOrDefault("NUXEO_PROFILES", EMPTY);
```
efge, line 117: I'd use `""` instead of `EMPTY`, it's shorter and doesn't require a dependency 😄

kevinleturc:
T&P: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-kleturc-2/201/

kevinleturc:
T&P: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-kleturc-2/202/

kevinleturc:
T&P: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-kleturc-2/203/

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/info/ConfigurationInfo.java
```diff
@@ -19,6 +19,8 @@
 
 package org.nuxeo.launcher.info;
 
+import org.apache.commons.collections.CollectionUtils;
```
troger, line 22: Unused import?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
```diff
@@ -2032,10 +2049,28 @@ protected String getJavaOptsString() {
...
+    protected String getEnvironment(String key, String defaultValue) {
+        return Objects.requireNonNullElse(getEnvironment(key), defaultValue);
+    }
+
+    /**
+     * @return the nuxeo.defaults file for current {@code NUXEO_ENVIRONMENT}
+     * @since 11.1
+     */
+    protected String getNuxeoEnvironmentConfName() {
+        return String.format(NUXEO_ENVIRONMENT_CONF_FORMAT, getEnvironment(NUXEO_ENVIRONMENT));
```
NourNuxeo, line 2074: wouldn't you want to use the other signature of getEnvironment() with a default ?
The signature with a default is used in the public setProperties() method (line 1677) so it may be more consistent ?
The documentation says this signature (without a default) is only for tests. But it seems to be used in several internal methods. Maybe it's outdated ?
> * @apiNote exists to be overridden by tests

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
```diff
@@ -141,8 +143,9 @@ public String getContextName() {
      */
     protected void parseAndCopy(Properties config) throws IOException, TemplateException, ConfigurationException {
         // FilenameFilter for excluding "nuxeo.defaults" files from copy
+        String nuxeoEnvironmentConf = generator.getNuxeoEnvironmentConfName();
```
NourNuxeo, line 146: this is only used once line 148, is this for clarity ?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
```diff
@@ -614,8 +617,11 @@ public InstanceInfo getInfo(String clid, List<LocalPackage> pkgs) {
             nxInstance.packages.add(info);
             pkgTemplates.addAll(info.templates);
         }
-        // templates
         nxInstance.config = new ConfigurationInfo();
+        // profiles
+        nxInstance.config.profiles.addAll(
+                Arrays.asList(generator.getEnvironment(NUXEO_PROFILES, "").split(TEMPLATE_SEPARATOR)));
```
NourNuxeo, line 623:
```suggestion
                List.of(generator.getEnvironment(NUXEO_PROFILES, "").split(TEMPLATE_SEPARATOR)));
```

modules/runtime/nuxeo-launcher-commons/src/test/java/org/nuxeo/launcher/config/ConfigurationGeneratorTest.java
```diff
@@ -531,4 +543,22 @@ public void testCheckEncoding() throws Exception {
         }
     }
 
+    @Test
+    public void testIncludeProfile() {
+        String profileToTest = "testprofile";
+        assertFalse("Profile should not be included", isTemplateIncluded(profileToTest));
+        assertNotEquals("true", configGenerator.getUserConfig().getProperty("nuxeo.profile.added.by.test"));
```
NourNuxeo, line 550: no more `Boolean.TRUE` ?

modules/runtime/nuxeo-launcher-commons/src/test/java/org/nuxeo/launcher/config/ConfigurationGeneratorTest.java
```diff
@@ -531,4 +543,22 @@ public void testCheckEncoding() throws Exception {
...
+
+        env.put(NUXEO_PROFILES, profileToTest);
+
+        configGenerator.init(true);
+
+        assertTrue("Profile should be included", isTemplateIncluded(profileToTest));
+        assertEquals("true", configGenerator.getUserConfig().getProperty("nuxeo.profile.added.by.test"));
```
NourNuxeo, line 557: same question about Boolean.TRUE

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
kevinleturc: There're several things in your comment.

>wouldn't you want to use the other signature of getEnvironment() with a default ?

Which default do you want to use?

>The signature with a default is used in the public setProperties() method (line 1677) so it may be more consistent?

Because I use the `String#isBlank` method and so I can't have null.

>The documentation says this signature (without a default) is only for tests. But it seems to be used in several internal methods. Maybe it's outdated?
@APinote exists to be overridden by tests

The sentence doesn't say it's used by tests, it says that the method exists in order to let let test override it to provide a mock (see the tests).

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
kevinleturc: What do you mean?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
kevinleturc: Not in this case, we get an array so it makes sense to use `Arrays` API.

modules/runtime/nuxeo-launcher-commons/src/test/java/org/nuxeo/launcher/config/ConfigurationGeneratorTest.java
kevinleturc: Check the type, we get String.

modules/runtime/nuxeo-launcher-commons/src/test/java/org/nuxeo/launcher/config/ConfigurationGeneratorTest.java
kevinleturc: Same answer.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
NourNuxeo: The variable is defined here and seems to be used only once right after on line 148 so I would have just made a call to `generator.getNuxeoEnvironmentConfName();` there instead of defining a variable. Unless you did this to make the call shorter and clearer on line 148 ?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
NourNuxeo: 1- I assumed this is to avoid null so I would have given `""` like line 1677 (like you explain in your second point
2- yes
3 - Ok thanks for the explenation !

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
kevinleturc: If I correctly remember, the line you're talking about is a function and it is called several times. That's the reason we get the variable before, otherwise, we will call it several times.

Note: for future reviews, comment on line(s) you want to talk about.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
kevinleturc: Probably, but this is not needed here. Furthermore, if the code evolves, I prefer to have `nuxeo.null` which clearly shows there's something wrong than `nuxeo.`.

efge:
Could NXP-29190 be updated (upgrade notes) to add a synthetic view of the new functionality?

PR #4120, opened 2020-06-05
task-NXP-28937-release-pipeline

ataillefer:


ci/Jenkinsfiles/build.groovy
```diff
@@ -396,8 +396,7 @@ pipeline {
           ----------------------------------------
           """
           sh """
-            git add .
-            git commit -m "Release ${VERSION}"
+            git commit -a -m "Release ${VERSION}"
```
NourNuxeo, line 399: is this because since there is no new untracked file to be added any way at this point that you privilledge `-a` over `.` (and to have a one-liner ?)

ataillefer: Yes for both reasons

ci/Jenkinsfiles/build.groovy
```diff
@@ -643,7 +642,8 @@ pipeline {
             git config credential.helper store
 
             # Git tag
-            jx step tag -v ${VERSION}
+            git tag -a v${VERSION} -m "Release ${VERSION}"
+            git push origin v${VERSION}
```
NourNuxeo, line 646: The original jx step tag adds a --force (-f) to the `git tag` call. I guess it's unecessary ? 
And calling the git command manually spares you commiting again uselessly ?

ataillefer: > The original jx step tag adds a --force (-f) to the git tag call. I guess it's unnecessary ?

Trying to replace an exiting tag should never happen and I think we prefer an explicit error if ever it were the case.

> And calling the git command manually spares you committing again uselessly ?

Yes it prevents from having an unnecessary extra empty commit, see for instance https://github.com/nuxeo/nuxeo/commit/90b1aaa37df5afb8610a8e89ce580817835aecad and
image

And its one less use of `jx` :smile:.

NourNuxeo: Thanks for explenations !

PR #4136, opened 2020-06-10
NXP-29243/NXP-29244: better doc XML export

efge:


efge:
T&P ok https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume/850/

PR #4146, opened 2020-06-13
NXP-29252: Fix EasyShare Expired mail template

ThibArg:
There was no unit test sending mail notification, I did not add one.
But I did test on Nuxeo 11 Snapshot :> all good, I do have my correct title (like "Nuxeo]EasyShare folder My EasyShare Test has expired")


efge:
@ThibArg the NXP number is wrong, could you update your commit message please? Also `Expired` -> `expired` in the message probably. And we usually don't use an initial capital so `Fix` -> `fix`

ThibArg:
Hope I did it. used git commit --amend, then  I have to merge something etc. etc.

efge:
> Hope I did it. used git commit --amend, then I have to merge something etc. etc.

?  As of now you haven't pushed anything

ThibArg:
Ouch true. I was so sure I did it... Sorry. Done.

efge:
But you now have 3 commits. There should be a single commit.

PR #4148, opened 2020-06-13
NXP-29256: fix UserMapperBasedResolver boolean parsing

efge:


modules/platform/login/nuxeo-platform-login-saml2/src/test/java/org/nuxeo/ecm/platform/auth/saml/SAMLAuthenticatorWithMapperTest.java
```diff
@@ -161,6 +162,7 @@ protected HttpServletRequest getMockRequest(String messageFile, String method, S
         when(request.getRequestURL()).thenReturn(new StringBuffer(url));
         when(request.getAttribute("javax.servlet.request.X509Certificate")).thenReturn(null);
         when(request.isSecure()).thenReturn(false);
+        when(request.getLocale()).thenReturn(Locale.ENGLISH);
```
ataillefer, line 165: Why is that needed, is it elated to the current fix?
Also, could the line below be removed with the cleanup commit?

efge: It's needed yes, otherwise when testing the case where we don't create the user then fetch it to check, the underlying code attempts to return an i18n error message and does a NPE when fetching the locale from the request (which is supposed to always exist).

For the removal of the commented-out line I'll postpone, the cleanup commit was just for whitespace and this whole code needs to be rewritten anyway for OpenSAML 3 in the medium term...

PR #4153, opened 2020-06-15
NXP-29252: fix EasyShare Expired mail template (#4146)

ThibArg:
I did test the mail subject was correct by deploying the built jar in a 10.10 and creating an expired EasyShareFolder.


efge:
T&P in progress

efge:
T&P ok https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume-10.10/115/

PR #4161, opened 2020-06-17
NXP-29266: upgrade Apache httpcomponents

efge:


efge:
The current PR fails because httpclient 4.5.12 changed the processing of 308 to force it to be a redirect code, and some of our tests don't expect that (we use 308 for BatchUpload return codes to mean something else, following Google's old and non-standard "Resume Incomplete").

efge:
Postponing this upgrade to 4.5.12.

PR #4162, opened 2020-06-17
NXP-29266: upgrade Apache httpcomponents (10.10)

efge:


efge:
Not ready for review, see comments in https://github.com/nuxeo/nuxeo/pull/4161#issuecomment-645715409

PR #4166, opened 2020-06-18
NXP-29194: allow to pass a Connect URL when running a Nuxeo Docker image

troger:


docker/README.md
```diff
@@ -194,6 +195,16 @@ For instance, to run a container with a registered Nuxeo instance:
 docker run -it -p 8080:8080 -e NUXEO_CLID=<NUXEO_CLID> nuxeo/slim:latest
 ```
 
+#### NUXEO_CONNECT_URL
+
+The value of `NUXEO_CONNECT_URL` is appended to `nuxeo.conf` at startup.
```
bdelbosc, line 200: maybe this is the option that is appended (key and value) and not just the value?

troger: What about:
> Configure the Connect URL with the `NUXEO_CONNECT_URL` value in `nuxeo.conf` at startup
?

ataillefer: Like `NUXEO_PACKAGES` I would say:
> `NUXEO_CONNECT_URL` allows to configure the Connect URL. It is used for instance to fetch the Nuxeo packages .

docker/README.md
```diff
@@ -170,6 +170,7 @@ Currently, there are three environment variables that are taken into account by
```
ataillefer, line 170: > Currently, there are three environment variables that are taken into account by a Nuxeo image:

Let's stop counting :)

> Currently, these are the environment variables that are taken into account by a Nuxeo image:

PR #4170, opened 2020-06-20
NXP-29287: don't log WARN if S3 bucket prefix does not end with slash

efge:


ataillefer:
Don't you want to also update `S3DirectBatchHandler`?

PR #4171, opened 2020-06-20
NXP-29287: don't log WARN if S3 bucket prefix does not end with slash (10.10)

efge:


efge:
T&P ok https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-fguillaume-10.10/119/

ataillefer:
See https://github.com/nuxeo/nuxeo/pull/4170#issuecomment-647385278.

PR #4214, opened 2020-07-15
NXP-29390: remove trashed documents from default retention search (10.10)

gcarlin:
T&P In Progress:
https://qa2.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-gcarlin-10.10/7/

### DIFF
--- a/addons/nuxeo-retention/nuxeo-retention-core/src/main/resources/OSGI-INF/retention-pageproviders.xml
+++ b/addons/nuxeo-retention/nuxeo-retention-core/src/main/resources/OSGI-INF/retention-pageproviders.xml
@@ -103,6 +103,9 @@
         <quickFilter name="exceptTrashed">
           <clause>ecm:isTrashed = 0</clause>
         </quickFilter>
+        <quickFilter name="includeTrashed">
+          <clause>ecm:isTrashed = 1</clause>
+        </quickFilter>
       </quickFilters>
       <pageSize>20</pageSize>
     </genericPageProvider>


addons/nuxeo-retention/nuxeo-retention-core/src/main/resources/OSGI-INF/retention-pageproviders.xml
```diff
@@ -100,8 +101,8 @@
         <quickFilter name="hasNotLegalHold">
           <clause>ecm:hasLegalHold = 0</clause>
         </quickFilter>
-        <quickFilter name="exceptTrashed">
-          <clause>ecm:isTrashed = 0</clause>
+        <quickFilter name="includeTrashed">
```
troger, line 104: Isn't it breaking the backward compatibility? It changes the quick filters returned by the REST API

gcarlin: For the moment, we have only one client using the addon, and they use the default UI. I can let the original quickfilter and hardcode the expected filters (and removing the exceptTrashed one) in the nuxeo-result element (https://github.com/nuxeo/nuxeo-web-ui/blob/master/elements/nuxeo-results/nuxeo-results.js#L275). What would be the best option?

nmpcunha: Guess it could be an option, indeed.

But, if we are sure about the impact (only one client and being a "private" addon), would it be that bad to perform the change? (cc @troger )
Of course,  stating that on the "Upgrade Notes" and double checking our impact assessment.

troger: If you know what you are doing and the impacts, I'm OK, just wanted to mention it :)

nmpcunha: TBH, I don't know 🤷‍♂️  😅 
But @gcarlin's input may be a "game changer". Anyway, the other option (safer one, keeping the existing quickFilter) is also viable.

gcarlin: Using "hardcoded" quickfilters on the nuxeo-results element also implies that a new quickfilter won't be displayed automatically. That said, I would opt for the safe option (never know if anyone else is using it).

nmpcunha: But in that case we would not be "hardcoding" since it is a parameter and the users can change it if they want! Or am I seeing it a wrong way?

gcarlin: Yes yes, but I guess that's the only part in Web UI we explicitly indicate which quickfilters to display (so you might not guess you need also to update the result element)

addons/nuxeo-retention/nuxeo-retention-core/src/main/resources/OSGI-INF/retention-pageproviders.xml
```diff
@@ -100,8 +101,8 @@
...
+          <clause>ecm:isTrashed = 1</clause>
```
nelsonsilva, line 105: With `ecm:isTrashed = 0` in the fixed part we'll end up with `ecm:isTrashed = 0 AND ecm:isTrashed = 1` => no results, no?

gcarlin: Indeed! So maybe adding a quickfilter is enough, and we do not modify the fixed part of the query.

addons/nuxeo-retention/nuxeo-retention-core/src/main/resources/OSGI-INF/retention-pageproviders.xml
gcarlin: So, according to nelson note (https://github.com/nuxeo/nuxeo/pull/4214#discussion_r455803626), we will only add an additional quickfilter

addons/nuxeo-retention/nuxeo-retention-core/src/main/resources/OSGI-INF/retention-pageproviders.xml
nelsonsilva: Initial requirement as described in the ticket was "Default retention search shouldn't list trashed documents" which this doesn't do and we already had a quickfilter to exclude trashed documents so effectively you're adding a new quick filter to list trashed documents under retention (which I thought you shouldn't be able to have in the first place, no?)

gcarlin: You can have documents under retention located in the trash (if compliance mode is not explicitly activated). So the original requirement is to get non trashed documents in the listing. But, as you noted correctly, if we have a fixed part excluding trashed document, we wouldn't be able to make any operation on these documents (extending retention period / apply a legal hold / attach a retention rule). So we would be loosing a default feature (even if we agree it is strange to perform retention actions on trashed documents)

nelsonsilva: But my point is we already have a quick filter to show only non trashed documents.... is there a new requirement to show only trashed now ?

gcarlin: Not really, so well considered, we can keep the original  code. The idea would be to get the quickfilter `exceptTrashed` selected by default.

PR #4230, opened 2020-07-29
NXP-27709: remove nuxeo-jsf-ui profile; simplify GitHub statuses

troger:


troger:
Submitted from fork...

PR #4231, opened 2020-07-29
NXP-29470: add one retry for utests and ftests + a global timeout

troger:


troger:
Submitted from fork...

PR #4233, opened 2020-07-29
impr NXP 27709 remove nuxeo jsf ui

troger:


ci/Jenkinsfiles/build.groovy
```diff
@@ -57,7 +57,7 @@ def isPullRequest() {
 }
 
 String getVersion() {
-  return isPullRequest() ? getPullRequestVersion() : getReleaseVersion()
+  return isPullRequest() ? readMavenPom().getVersion() : getReleaseVersion()
```
kevinleturc, line 60: Wouldn't it be a problem for docker images? I'm talking about collision on our Docker registry.

troger: You're right... I knew we missed something @ataillefer 😂

PR #4236, opened 2020-08-03
NXP-29319: allow using any file key in s3

nelsonsilva:
T&P in progress https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-nsilva/163

### DIFF
--- a/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/KeyStrategyManaged.java
+++ b/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/KeyStrategyManaged.java
@@ -41,7 +41,8 @@ public class KeyStrategyManaged implements KeyStrategy {
     @Override
     public BlobWriteContext getBlobWriteContext(BlobContext blobContext)  {
         if (blobContext.blob instanceof ManagedBlob) {
-            String key = ((ManagedBlob) blobContext.blob).getKey();
+            // strip version id if part of file key
+            String key = ((ManagedBlob) blobContext.blob).getKey().split(String.valueOf(VER_SEP))[0];
             return new BlobWriteContext(blobContext, null, () -> key, this);
         } else {
             return strategy.getBlobWriteContext(blobContext);
--- a/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/KeyStrategyManaged.java
+++ b/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/KeyStrategyManaged.java
@@ -1,0 +1,64 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/KeyStrategyManaged.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    public String getDigestFromKey(String key) {
+        return null;
+    }
+
+    @Override
+    public BlobWriteContext getBlobWriteContext(BlobContext blobContext)  {
+        if (blobContext.blob instanceof ManagedBlob) {
+            String key = ((ManagedBlob) blobContext.blob).getKey();
```
efge, line 44: Here we need to strip the `@versionid` if one is present. This needs `S3BlobStore.VER_SEP` which should probably be moved to `AbstractBlobStore` and made public. (This is a bit dirty because versions are S3-specific at the moment and here we're in more generic code but at least it'll work and we can refactor later.)

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
```diff
@@ -312,13 +320,25 @@ public boolean completeUpload(String batchId, String fileIndex, BatchFileInfo fi
         try {
             batch.addFile(fileIndex, blob, blob.getFilename(), blob.getMimeType());
         } catch (NuxeoException e) {
-            amazonS3.deleteObject(bucket, bucketKey);
+            amazonS3.deleteObject(bucket, fileKey);
```
efge, line 323: It's likely that in the presence of versions this will be incorrect too, the `fileKey` will contain a `@versionid` but we want to do a delete of a specific version (-> parsing and use of `amazonS3.deleteVersion`, see `S3BlobStore.deleteBlob`) ...

nelsonsilva: The `fileKey` here is coming from the client so I don't think we'll ever have a `@versionId`

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/KeyStrategyManaged.java
nelsonsilva: Moved `VER_SEP` to `KeyStrategy` (not sure how you feel about constants in interfaces but it makes for clean code).

packages/nuxeo-amazon-s3-package/src/main/resources/install/templates/s3binaries/nuxeo.defaults
```diff
@@ -5,6 +5,7 @@ s3binaries.target=.
 # mandatory
 nuxeo.s3storage.bucket=
 nuxeo.core.binarymanager=org.nuxeo.ecm.core.storage.sql.S3BinaryManager
+nuxeo.core.blobstore.keyStrategy=managed
```
kevinleturc, line 8: Is this configuration change is compatible with an upgrade? ie: it will be transparent for the users?

nelsonsilva: switching from `digest` to `managed` should work fine as we're basically just trusting the client and letting them use any key. We could leave it as `digest` by default but then uploads > 5GBs won't work... not until [NXP-29517|https://jira.nuxeo.com/browse/NXP-29517] is done. OTOH we can't easily migrate back from `managed` to `digest` 🤔  so perhaps we should indeed leave it as `digest` for now

PR #4237, opened 2020-08-04
Fix nxp 29237 nuxeo drive mongodb

NourNuxeo:


modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/test/java/org/nuxeo/drive/test/MongoDBAuditFeature.java
```diff
@@ -0,0 +1,36 @@
...
+import org.nuxeo.drive.mongodb.TestMongoDBAuditChangeFinder;
+import org.nuxeo.ecm.automation.test.AutomationFeature;
+import org.nuxeo.ecm.platform.audit.AuditFeature;
+import org.nuxeo.runtime.test.runner.Deploy;
+import org.nuxeo.runtime.test.runner.Features;
+import org.nuxeo.runtime.test.runner.RunnerFeature;
+
+/**
+ * MongoDB audit feature for nuxeo drive. This Feature is necessary for the deployment annotations to be taken into
+ * account by {@link TestMongoDBAuditChangeFinder}.
```
ataillefer, line 28: `MongoDB audit feature for Nuxeo Drive.` is enough I think

modules/platform/nuxeo-platform-audit/nuxeo-platform-audit-mongodb/src/test/java/org/nuxeo/mongodb/audit/MongoDBAuditFeature.java
```diff
@@ -43,6 +45,18 @@
...
 
+    @Override
+    public void initialize(FeaturesRunner runner) {
+        runner.getFeature(TransactionalFeature.class).addWaiter(duration -> {
+            // Wait for audit completion
+            long before = System.currentTimeMillis();
+            if (!Framework.getService(AuditLogger.class).await(duration.toMillis(), TimeUnit.MILLISECONDS)) {
+                return false;
+            }
+            return true;
```
ataillefer, line 56: Simply
```
Framework.getService(AuditLogger.class).await(duration.toMillis(), TimeUnit.MILLISECONDS)
```

Sonar should warn you about
> Replace this if-then-else statement by a single return statement.

modules/platform/nuxeo-platform-audit/nuxeo-platform-audit-mongodb/src/test/java/org/nuxeo/mongodb/audit/MongoDBAuditFeature.java
```diff
@@ -43,6 +45,18 @@
 @Features({ MongoDBFeature.class, CoreFeature.class })
 public class MongoDBAuditFeature implements RunnerFeature {
 
+    @Override
+    public void initialize(FeaturesRunner runner) {
+        runner.getFeature(TransactionalFeature.class).addWaiter(duration -> {
+            // Wait for audit completion
+            long before = System.currentTimeMillis();
```
ataillefer, line 52: Unused variable.
Sonar should warn you about
> Remove this unused "before" local variable.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/test/java/org/nuxeo/drive/test/NuxeoDriveMongoDBAuditFeature.java
```diff
@@ -0,0 +1,51 @@
...
+public class NuxeoDriveMongoDBAuditFeature implements RunnerFeature {
+
+    @Override
+    public void initialize(FeaturesRunner runner) {
+        runner.getFeature(TransactionalFeature.class)
+              .addWaiter(duration -> {
+                  // Wait for audit completion
+                  return Framework.getService(AuditLogger.class).await(duration.toMillis(), TimeUnit.MILLISECONDS);
+              });
+    }
```
ataillefer, line 50: You don't need this since it will now be handled by the `MongoDBAuditFeature`.
Then, of course, we'll have to validate that this change in `MongoDBAuditFeature` doesn't break other tests using it.
Though the T&P will validate it, you can already run locally the 6 tests using it in `nuxeo-audit-storage-directory` and `nuxeo-platform-audit-mongodb`.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+      <artifactId>nuxeo-search-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.nuxeo.ecm.platform</groupId>
+      <artifactId>nuxeo-platform-tag-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+</project>
```
ataillefer, line 183: There seems to be a lot of unused dependencies here.
Probably a copy-paste from another drive POM.
In this case, it's probably more relevant to start with a clean POM with no dependencies and add them as they're needed.

Don't hesitate to run `mvn -nsu dependency:analyze`, which currently shows:
```[INFO] --- maven-dependency-plugin:3.1.0:analyze (default-cli) @ nuxeo-drive-mongodb ---
[WARNING] Used undeclared dependencies found:
[WARNING]    org.nuxeo.ecm.core:nuxeo-core-query:jar:11.2-SNAPSHOT:compile
[WARNING]    org.mongodb:mongo-java-driver:jar:3.12.1:compile
[WARNING] Unused declared dependencies found:
[WARNING]    org.apache.commons:commons-lang3:jar:3.9:compile
[WARNING]    org.nuxeo.runtime:nuxeo-runtime-jtajca:jar:11.2-SNAPSHOT:compile
[WARNING]    org.nuxeo.ecm.core:nuxeo-core-event:jar:11.2-SNAPSHOT:compile
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-audit-core:jar:11.2-SNAPSHOT:compile
[WARNING]    org.nuxeo.ecm.automation:nuxeo-automation-core:jar:11.2-SNAPSHOT:compile
[WARNING]    org.nuxeo.ecm:nuxeo-drive-operations:jar:11.2-SNAPSHOT:compile
[WARNING]    javax.inject:javax.inject:jar:1:test
[WARNING]    org.nuxeo.ecm.core:nuxeo-core-management-jtajca:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.core:nuxeo-core-management-jtajca:test-jar:tests:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-test:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-userworkspace-types:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-userworkspace-core:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-collections-core:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-collections-core:test-jar:tests:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-relations-api:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-uidgen-core:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.core:nuxeo-core-cache:test-jar:tests:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.core:nuxeo-core-redis:test-jar:tests:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-admin-center-core:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-webapp-types:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-search-core:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-tag-core:jar:11.2-SNAPSHOT:test
```
So, many dependencies to remove and some to add.
Note, some dependencies required by the unit tests might not be detected since they're implicitly used by XML contributions, so you'll have to add them back if needed.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/test/java/org/nuxeo/drive/test/MongoDBAuditFeature.java
NourNuxeo: since we changed the name how about only keeping the since ?

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/pom.xml
NourNuxeo: Yes I copied nuxeo-drive-elasticsearch.
Didn't know about that dependency:analyze thing. Looks very interesting !

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/test/java/org/nuxeo/drive/test/NuxeoDriveMongoDBAuditFeature.java
NourNuxeo: org.nuxeo.audit.storage.impl.TestRestorationFromDirectoryAuditStorage is affected by the initialize overriding:
The test takes 400 seconds instead of 40 and gives a lot of errors while trying and retrying. Sometimes eclipse marks it as failing sometimes succeeding. Here the exceptions I get:
```
2020-08-05 15:05:50,208 [audit/writerPool-00,in:1,inCheckpoint:0,out:0,lastRead:1596632627011,lastTimer:1596632626942,wm:209273831663206400,loop:504,record] WARN  [AbstractBatchComputation] Computation: audit/writer fails to process batch of 1 records, last record: audit-audit-00:+79366700662784, retrying ...
com.mongodb.MongoBulkWriteException: Bulk write operation error on server localhost:27017. Write errors: [BulkWriteError{index=0, code=11000, message='E11000 duplicate key error collection: unittests.audit index: _id_ dup key: { : 7 }', details={}}]. 
	at com.mongodb.connection.BulkWriteBatchCombiner.getError(BulkWriteBatchCombiner.java:173) ~[mongo-java-driver-3.12.1.jar:?]
```
Other tests are fine

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/test/java/org/nuxeo/drive/test/MongoDBAuditFeature.java
ataillefer: Yes, its name is explicit :+1:

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/test/java/org/nuxeo/drive/test/NuxeoDriveMongoDBAuditFeature.java
ataillefer: OK, strange. So you can keep th code in the `NuxeoDriveMongoDBAuditFeature`.
Hoping the Drive tests will pass.

NourNuxeo: Not yet 😭

NourNuxeo:
Also need to add ataillefer's  blacklist patch and test on a real server

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/pom.xml
NourNuxeo: Some jars are wrongly detected as you warned but it has slimmed down a lot:
```
[WARNING] Unused declared dependencies found:
[WARNING]    org.nuxeo.runtime:nuxeo-runtime-mongodb:test-jar:tests:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-userworkspace-types:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-userworkspace-core:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.core:nuxeo-core-cache:test-jar:tests:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-search-core:jar:11.2-SNAPSHOT:test
[WARNING]    org.nuxeo.ecm.platform:nuxeo-platform-tag-core:jar:11.2-SNAPSHOT:test
```
and no more 
```[WARNING] Used undeclared dependencies found:```

**Comment by NourNuxeo** [comment]
🟢 T&P Ok: https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-nalkotob/182/

File: modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java [hunk 6]
```diff
@@ -0,0 +1,232 @@
...
+                log.error("couldn't read query end template", e);
+            }
+        }
+        queryBuilder.append(queryEnd);
+
+        Bson filter = auditService.buildFilter(queryBuilder.toString(), params);
+        var sorts = List.of(Sorts.ascending("repositoryId"), Sorts.descending("eventDate"));
+        var order = Sorts.orderBy(sorts);
+        FindIterable<Document> iterable = auditCollection.find(filter).sort(order);
+        List<LogEntry> entries = StreamSupport.stream(iterable.spliterator(), false)
```
kevinleturc, line 184: `FindIterable#spliterator` is backed by the default method in `Iterable` which leverages `Iterable#iterator`, or in this case `MongoIterableImpl#iterator` which returns a `MongoCursor` which needs to be closed. You have an example of what to do in [MongoDBConnection](https://github.com/nuxeo/nuxeo/blob/927293a3f326fbcc8d8662ecee35ff6a2d5c8601/modules/core/nuxeo-core-storage-mongodb/src/main/java/org/nuxeo/ecm/core/storage/mongodb/MongoDBConnection.java#L698).

Otherwise, you could use the MongoDB API there, something like:

```java
List<LogEntry> entries = auditCollection.find(filter).sort(order).map(MongoDBAuditEntryReader::read).into(new ArrayList<>());
```

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,232 @@
...
+        for (LogEntry entry : entries) {
+            ExtendedInfo impactedUserInfo = entry.getExtendedInfos().get("impactedUserName");
+            if (impactedUserInfo != null && !principalName.equals(impactedUserInfo.getValue(String.class))) {
+                // ignore event that only impact other users
+                continue;
+            }
+            log.debug("Change detected: {}", entry);
+            postFilteredEntries.add(entry);
+        }
+        return postFilteredEntries;
```
kevinleturc, line 201: This could be expressed in the MongoDB query?

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/test/java/org/nuxeo/drive/test/NuxeoDriveMongoDBAuditFeature.java
```diff
@@ -0,0 +1,44 @@
...
+@Features({ AutomationFeature.class, AuditFeature.class, MongoDBAuditFeature.class })
+@Deploy("org.nuxeo.drive.mongodb")
+public class NuxeoDriveMongoDBAuditFeature implements RunnerFeature {
+
+    @Override
+    public void initialize(FeaturesRunner runner) {
+        runner.getFeature(TransactionalFeature.class).addWaiter(duration -> {
+            return Framework.getService(AuditLogger.class).await(duration.toMillis(), TimeUnit.MILLISECONDS);
+        });
+    }
```
kevinleturc, line 42: Is it really needed? `AuditFeature` already declare a waiter for this.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/pom.xml
```diff
@@ -0,0 +1,114 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.nuxeo.ecm</groupId>
+    <artifactId>nuxeo-drive-parent</artifactId>
+    <version>11.2-SNAPSHOT</version>
```
efge, line 6: Now 11.3-SNAPSHOT of course

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,232 @@
...
+            Set<String> collectionSyncRootMemberIds, long lowerBound, long upperBound, int limit) {
+        MongoDBAuditBackend auditService = (MongoDBAuditBackend) Framework.getService(AuditReader.class);
+        MongoCollection<Document> auditCollection = auditService.getAuditCollection();
+        StringBuilder queryBuilder = new StringBuilder();
+        Map<String, Object> params = new HashMap<>();
+        params.put("repositoryId", session.getRepositoryName());
+        params.put("lowerBound", lowerBound);
+        params.put("upperBound", upperBound);
+
+        if(queryStart == null) {
```
efge, line 147: Space before `(`

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,232 @@
...
+        // second conditional filter
+        if (collectionSyncRootMemberIds != null && !collectionSyncRootMemberIds.isEmpty()) {
+            queryBuilder.append(",\n");
+            queryBuilder.append(getCollectionSyncRootFilteringClause(collectionSyncRootMemberIds, params) + "\n");
+        }
+        // if the first conditional filter was set, then an $or array has been opened and needs to be closed
+        if (!activeRoots.getPaths().isEmpty()) {
+            queryBuilder.append("]}\n");
+        }
+        if(queryEnd == null) {
```
efge, line 171: Space before `(`

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,232 @@
...
+        Map<String, Object> params = new HashMap<>();
+        params.put("repositoryId", session.getRepositoryName());
+        params.put("lowerBound", lowerBound);
+        params.put("upperBound", upperBound);
+
+        if(queryStart == null) {
+            try (InputStream is = getClass().getClassLoader().getResourceAsStream("OSGI-INF/queryStart.json")) {
+                queryStart = IOUtils.toString(is, "UTF-8");
+            } catch (IOException e) {
+                log.error("couldn't read query start template", e);
```
efge, line 151: Then rethrow wrapped in `NuxeoException`, we can't do anything more.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,232 @@
...
+            }
+        }
+        queryBuilder.append(queryStart);
```
efge, line 154: I would very much prefer if we had only one file with the whole JSON query and in the middle a special pattern, maybe like `__BODY__` (or something else that makes sense), that gets replaced by what we want. It will make the file much more readable.

But actually after reading more of the code all this string replacement to in the end get a BSON object is really unnatural, it would be better to directly create a BSON object (using `new Document()` etc). This applies also to methods `getCurrentRootFilteringClause` and `getCollectionSyncRootFilteringClause` which should work at the BSON/Document level, not strings.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,232 @@
...
+            try (InputStream is = getClass().getClassLoader().getResourceAsStream("OSGI-INF/queryEnd.json")) {
+                queryEnd = IOUtils.toString(is, "UTF-8");
+            } catch (IOException e) {
+                log.error("couldn't read query end template", e);
+            }
+        }
+        queryBuilder.append(queryEnd);
+
+        Bson filter = auditService.buildFilter(queryBuilder.toString(), params);
+        var sorts = List.of(Sorts.ascending("repositoryId"), Sorts.descending("eventDate"));
```
efge, line 181: If this has to be backported to 10.10 you'd bette not use `var`, or be ready to update the backport for Java 8.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
efge: Could be yes, that's not an option we had in JPA but here if the underlying structure of `impactedUserName` can be queried directly it'll be more efficient.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/resources/OSGI-INF/queryStart.json
```diff
@@ -0,0 +1,49 @@
+{
+  "$and": [
+    {
+      "repositoryId": "test"
```
efge, line 4: This seems wrong, it won't work for real repositories outside of tests.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
NourNuxeo: solved with https://github.com/nuxeo/nuxeo/pull/4237#discussion_r468950144

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
NourNuxeo: To be continued

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,187 @@
...
+                and(or(and(eq("category", "eventDocumentCategory"),
+                        in("eventId", "documentCreated", "documentModified", "documentMoved", "documentCreatedByCopy",
+                                "documentRestored", "addedToCollection", "documentProxyPublished", "documentLocked",
+                                "documentUnlocked", "documentUntrashed")),
+                        and(eq("category", "eventLifeCycleCategory"), eq("eventId", "lifecycle_transition_event"),
+                                ne("docLifeCycle", "deleted"))),
+                        or(regex("docPath", getRootsRegex(activeRoots.getPaths())),
+                                in("docUUID", collectionSyncRootMemberIds))),
+                and(eq("category", "NuxeoDrive"), ne("eventId", "rootUnregistered"))), gt("_id", lowerBound),
+                lte("_id", upperBound));
```
efge, line 151: This is hard to read, you can't understand what's ANDde or ORed with what. You should use intermediate variables with meaningful names to construct the global filter step by step.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,195 @@
...
+ *
+ * @since 11.3
+ */
+public class MongoDBAuditChangeFinder extends AuditChangeFinder {
+
+    private static final Logger log = LogManager.getLogger(MongoDBAuditChangeFinder.class);
+
+    protected String queryStart;
+
+    protected String queryEnd;
```
kevinleturc, line 91: Both unused.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,195 @@
...
+        Bson sessionRepository = eq("repositoryId", session.getRepositoryName());
+        Bson driveCategory = eq("category", "NuxeoDrive");
+        Bson notRootUnregistered = ne("eventId", "rootUnregistered");
+        Bson isGeneral = and(or(documentEvent, lifeCycleEvent), or(inRootPath, inRootId));
+        Bson isDrive = and(driveCategory, notRootUnregistered);
+        Bson idRange = and(gt("_id", lowerBound), lte("_id", upperBound));
+
+        // Build final filter
+        Bson filter = and(sessionRepository, or(isGeneral,
+                isDrive), idRange);
```
kevinleturc, line 159: Format

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,195 @@
...
+
+        BsonDocument bsonDocument = filter.toBsonDocument(Document.class, auditCollection.getCodecRegistry());
+        log.debug("Query on MongoDB-Audit: {}", bsonDocument.toJson());
```
kevinleturc, line 162: The query evaluation should be done in a lazy way:
```suggestion
        log.debug("Query on MongoDB-Audit: {}", () -> filter.toBsonDocument(Document.class, auditCollection.getCodecRegistry()).toJson());
```

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,195 @@
...
+
+        Bson order = Sorts.orderBy(Sorts.ascending("repositoryId"), Sorts.descending("eventDate"));
```
kevinleturc, line 164: What do you think about importing statically `orderBy`, `ascending` and `descending` like you do for `Filters`' method?

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,195 @@
...
+            }
+            log.debug("Change detected: {}", entry);
+            postFilteredEntries.add(entry);
+        }
+        return postFilteredEntries;
+    }
+
+    protected String getRootsRegex(Set<String> roots) {
+        if(roots.isEmpty()) {
+            return "";
```
kevinleturc, line 188: If the `roots` is empty, I think we shouldn't append the filter otherwise it will result in something like below and I don't know what MongoDB does on such thing:
```
{ "docPath": { "$regex": // } }
```

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+        StringBuilder rootPathClause = new StringBuilder("^");
+        rootPathClause.append(String.join("|^", roots));
+        return rootPathClause.toString();
```
kevinleturc, line 192:
```suggestion
        return "^" + String.join("|^", rootPathClause);
```
You don't need the `StringBuilder` here. As it is a simple append (no if or loop around), the compilator will produce more or less the same byte code.

Note: `+` is replaced by StringBuilder during compilation.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
NourNuxeo: Totally forgot that aspect

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
efge: Yes `+` is better. Note that in recent JDKs there are even better internal strategies than StringBuilder that are used to implement `+` on a fixed set of arguments.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
efge: +1. This can be refactored into a method that builds the whole filter for `inRootPath` or `or(inRootPath, inRootId)` actually.

NourNuxeo: made it conditional now with a call to `addRoots`
Made sure not to add extra useless ORs and ANDs on the way

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,198 @@
...
+        if (!rootPaths.isEmpty()) {
+            rootFilters.add(regex("docPath", "^" + String.join("|^", rootPaths)));
+        }
+        if (!collectionSyncRootMemberIds.isEmpty()) {
+            rootFilters.add(in("docUUID", collectionSyncRootMemberIds));
+        }
+        if (rootFilters.size() > 1) {
+            return and(baseFilter, or(rootFilters));
+        }
+        return and(baseFilter, rootFilters.get(0));
```
efge, line 195: This should be clearer:

          List<Bson> rootFilters = new ArrayList<>();
          if (!rootPaths.isEmpty()) {
              rootFilters.add(regex("docPath", "^" + String.join("|^", rootPaths)));
          }
          if (!collectionSyncRootMemberIds.isEmpty()) {
              rootFilters.add(in("docUUID", collectionSyncRootMemberIds));
          }
          if (rootFilters.isEmpty()) {
              return baseFilter;
          } else {
              rootFilters.add(baseFilter);
              return and(rootFilters);
          }

NourNuxeo: Doesn't cut the unit tests because ORs between rootFilters are missing.
If you don't mind making an or(list) with a single element in list (ANDs are only added when necessary in the final query so I guess ORs as well) we can do the following:
```
        if (!rootPaths.isEmpty() && !collectionSyncRootMemberIds.isEmpty()) {
            return and(baseFilter);
        }
        List<Bson> rootFilters = new ArrayList<>();
        if (!rootPaths.isEmpty()) {
            rootFilters.add(regex("docPath", "^" + String.join("|^", rootPaths)));
        }
        if (!collectionSyncRootMemberIds.isEmpty()) {
            rootFilters.add(in("docUUID", collectionSyncRootMemberIds));
        }
        return and(baseFilter, or(rootFilters));
```

efge: Hm yeah I misread the original code. But it's ugly to test twice for `rootPaths.isEmpty` and `collectionSyncRootMemberIds.isEmpty()`. Just change my code a bit to do at the end:

          return and(baseFilter, or(rootFilters));

NourNuxeo:
🟢 T&P OK: https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-nalkotob/191/

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
NourNuxeo: It seems to me we never fall into continue here because the impactedUserName property does not exist.
None of the existing tests covers this post filtering aspect which seems to not be working to me.
I Added this to the mongo query (using the extended.user property in place of `impactedUserName`) and it seems test get broken when filtering really works.

efge: This was added by 6b8901e8954b2ec7ce9ffb4ac9035f3e7ba44acc but indeed I can't find any use

NourNuxeo: How about removing this whole chunk of code since it has never been of use and since nobody complained about wrong behavior?

efge: Ok with me. @kevinleturc ?

kevinleturc: Ok too 👍

troger: Seems it's set here: https://github.com/nuxeo/nuxeo/blob/master/modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/main/java/org/nuxeo/drive/listener/NuxeoDriveFileSystemDeletionListener.java#L249

efge: Indeed and it's set by https://github.com/nuxeo/nuxeo/blob/11ef171/modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/main/java/org/nuxeo/drive/service/impl/NuxeoDriveManagerImpl.java#L313 and `fireEvent` is called with a specific userName.

So I retract my comment and we should keep this code.

efge (changes requested):
After the last comments, we should keep the code that checks `impactedUserName`

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
NourNuxeo: Instead of keeping the code. How about adding this filter at the end of the root `AND`
```
Bson isUser = and(or(exists("extended.impactedUser", true), eq("extended.user", session.getPrincipal().getName())));
```
Then a test covering that. I see some logs are generated by System and some are generated by Administrator so it should be easy to test.

NourNuxeo:
🟢 T&P Ok: https://qa.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-nalkotob/194/
This takes into consideration : https://github.com/nuxeo/nuxeo/pull/4237#discussion_r472819218

efge:
I'd like to see a test that explicitly checks the `impactedUser` filtering stuff (i.e., a test that fails without the new filter, and passes when the filter is there).

efge (changes requested):
I'd like to see a test that explicitly checks the `impactedUser` filtering stuff (i.e., a test that fails without the new filter, and passes when the filter is there).

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
NourNuxeo: While working more on the subject I have noticed my debug queries are always printed twice. I think this is why:
https://github.com/nuxeo/nuxeo/blob/master/modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/main/java/org/nuxeo/drive/service/impl/AuditChangeFinder.java#L101-L116

We query once the log entries and if there is a change in the drive roots, we update them and do the query again.

The check done after the first query could be done as a simple query rather than post filtering the log entries.
Wdyt ?

modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/test/java/org/nuxeo/drive/fixtures/AuditChangeFinderTestSuite.java
```diff
@@ -79,6 +79,35 @@
...
     protected TrashService trashService;
 
+    /** @since 11.3 */
+    @Test
+    public void testImpactedUser() {
+        try {
+            nuxeoDriveManager.registerSynchronizationRoot(session.getPrincipal(), folder2, session);
+        } finally {
+            txFeature.nextTransaction();
+        }
```
efge, line 89: Not sure why try/finally is needed here

modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/test/java/org/nuxeo/drive/fixtures/AuditChangeFinderTestSuite.java
```diff
@@ -79,6 +79,35 @@
...
+        // We need to save the lower bound id
+        long replayFrom = lastEventLogId;
+        List<FileSystemItemChange> changes = getChanges();
+        assertEquals(2, changes.size());
+
+        // So we can query the same id range again
+        lastEventLogId = replayFrom;
+        NuxeoPrincipalImpl foo = new NuxeoPrincipalImpl("foo");
+        List<FileSystemItemChange> changes2 = getChanges(foo);
+        assertEquals(0, changes2.size());
```
NourNuxeo, line 100:
```suggestion
        List<FileSystemItemChange> changes2 = getChanges(foo);
        assertEquals(0, getChanges(foo).size());
```

efge (approved):

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/test/java/org/nuxeo/drive/test/NuxeoDriveMongoDBAuditFeature.java
```diff
@@ -0,0 +1,31 @@
...
+import org.nuxeo.mongodb.audit.MongoDBAuditFeature;
+import org.nuxeo.runtime.test.runner.Deploy;
+import org.nuxeo.runtime.test.runner.Features;
+import org.nuxeo.runtime.test.runner.RunnerFeature;
+
+/** @since 11.3 */
+@Features({ AutomationFeature.class, AuditFeature.class, MongoDBAuditFeature.class })
+@Deploy("org.nuxeo.drive.mongodb")
+public class NuxeoDriveMongoDBAuditFeature implements RunnerFeature {
+
```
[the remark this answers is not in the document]
kevinleturc: Sure :+1:

modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/test/java/org/nuxeo/drive/fixtures/AuditChangeFinderClusteringEnabledTestSuite.java
```diff
@@ -34,14 +34,12 @@
...
 import org.nuxeo.ecm.core.api.DocumentModel;
 import org.nuxeo.ecm.core.api.impl.blob.StringBlob;
-import org.nuxeo.runtime.test.runner.Deploy;
 
 /**
  * Tests the {@link FileSystemChangeFinder} on a repository with clustering enabled.
  *
  * @since 8.2
  */
-@Deploy("org.nuxeo.drive.core:OSGI-INF/test-nuxeodrive-clustering-enabled-repo-contrib.xml")
```
ataillefer, line 44: This shouldn't be removed, otherwise clustering isn't enabled for this test suite.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,179 @@
...
+    private static final Logger log = LogManager.getLogger(MongoDBAuditChangeFinder.class);
+
+    @Override
+    public long getUpperBound() {
+        RepositoryManager repositoryManager = Framework.getService(RepositoryManager.class);
+        return getUpperBound(new HashSet<>(repositoryManager.getRepositoryNames()));
+    }
+
+    /**
+     * Returns the last available log id in the audit index considering events older than the last clustering
```
ataillefer, line 97:
```suggestion
     * Returns the last available log id in the audit collection considering events older than the last clustering
```

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,179 @@
...
+     * invalidation date if clustering is enabled for at least one of the given repositories. This is to make sure the
+     * {@code DocumentModel} further fetched from the session using the audit entry doc id is fresh.
+     */
+    @Override
+    public long getUpperBound(Set<String> repositoryNames) {
+        long clusteringDelay = getClusteringDelay(repositoryNames);
+        MongoDBAuditBackend auditService = (MongoDBAuditBackend) Framework.getService(AuditReader.class);
+        QueryBuilder queryBuilder = new AuditQueryBuilder().predicate(
+                Predicates.in(LOG_REPOSITORY_ID, repositoryNames));
```
ataillefer, line 106: What's the semantics of this predicate?
I think we can assume that any audit log entry has a repository id among the existing ones.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,179 @@
...
+        if (clusteringDelay > -1) {
+            // Double the delay in case of overlapping, see https://jira.nuxeo.com/browse/NXP-14826
+            long lastClusteringInvalidationDate = System.currentTimeMillis() - 2 * clusteringDelay;
+            queryBuilder.and(Predicates.lt(LOG_LOG_DATE, lastClusteringInvalidationDate));
+        }
+        queryBuilder.order(OrderByExprs.desc(LOG_ID));
+        queryBuilder.limit(1);
+        List<LogEntry> entries = auditService.queryLogs(queryBuilder);
```
ataillefer, line 114: Have you checked the impact of such an ordered query on a huge collection?

I'm referring to the crucial [optimization](https://github.com/nuxeo/nuxeo/blob/master/modules/platform/nuxeo-drive-server/nuxeo-drive-elasticsearch/src/main/java/org/nuxeo/drive/elasticsearch/ESAuditChangeFinder.java#L251) that was done for the Elasticsearch change finder in the context of https://jira.nuxeo.com/browse/NXP-24232..

ataillefer (changes requested):
Among the requested changes, great work for this non trivial MongoDB implementation!

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
NourNuxeo: https://github.com/nuxeo/nuxeo/blob/master/modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/main/java/org/nuxeo/drive/service/impl/AuditChangeFinder.java#L69 queries the logs.
If it finds NuxeoDrive events it invalidates the roots cache and does the same query again:
https://github.com/nuxeo/nuxeo/blob/master/modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/main/java/org/nuxeo/drive/service/impl/AuditChangeFinder.java#L101-L117

I have factorized things so invalidation stays unique but instead of doing that for loop I proposed making a mongo query. The change has been rejected because doing an extra query has been deemed too expensive while doing the same query twice just relies on costless cached results (efge will be able to explain more). However If you want to see, I have stashed it.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,179 @@
...
+
+        if (entries.isEmpty()) {
+            if (clusteringDelay > -1) {
+                // Check for existing entries without the clustering invalidation date filter to not return -1 in this
+                // case and make sure the lower bound of the next call to NuxeoDriveManager#getChangeSummary will be >=
+                // 0
+                if (auditService.getAuditCollection().estimatedDocumentCount() == 0) {
```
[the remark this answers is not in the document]
NourNuxeo: thanks, it happens my sonar was broken and I never noticed...(the secured storage bug again)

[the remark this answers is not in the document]
NourNuxeo: strangely sonar is working now but this rule doesn't show (I can provoke other rules to test it works)

modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/test/java/org/nuxeo/drive/fixtures/AuditChangeFinderTestSuite.java
```diff
@@ -79,6 +79,31 @@
...
+        NuxeoPrincipalImpl foo = new NuxeoPrincipalImpl("foo");
+        assertEquals(0, getChanges(foo).size());
+
+        // Now we check the found logs:
+        FileSystemItemChange change = changes.get(0);
+        assertEquals(folder2.getId(), change.getDocUuid());
+        assertEquals("rootRegistered", change.getEventId());
+        change = changes.get(1);
+        assertEquals(folder2.getId(), change.getDocUuid());
+        assertEquals("documentCreated", change.getEventId());
```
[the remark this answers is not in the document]
NourNuxeo: I have definitely missed something because I was pretty sure it worked but well...obviously...
however I'm not sure I understood your point.
The mongo filter linked (https://github.com/nuxeo/nuxeo/pull/4237/files#diff-d4ec341d2c4e8ec9ac9ff5775f2f006bR152) is not equivalent to the SQL clause: https://github.com/nuxeo/nuxeo/blob/master/modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/main/java/org/nuxeo/drive/service/impl/AuditChangeFinder.java#L350

The mongo filter that corresponds to it is https://github.com/nuxeo/nuxeo/pull/4237/files#diff-d4ec341d2c4e8ec9ac9ff5775f2f006bR157-R158 and is always "ANDed" at the end of the top-level filter array: https://github.com/nuxeo/nuxeo/pull/4237/files#diff-d4ec341d2c4e8ec9ac9ff5775f2f006bR161

Also when registering a root, NuxeoDrive events are fired as well, not only when deleting.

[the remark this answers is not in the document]
NourNuxeo: Replaying the tests it seems to work correctly:
"foo" can't see the logs where impacted user is "Administrator" in both SQL and MongoDB tests
if I do the same query as foo with Administrator, I see the logs again, so it means the bounds are correct.

There is still a big issue with the clustering test going on. And it seems to me the mongo driver has a problem with dates and timestamps handling.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
NourNuxeo: This is my way of avoiding giving an AND() predicate with an empty array when clustering is disabled. Mongo would complain about it.
This is because the AuditQueryBuilder will init with an empty AND()

So I tell it to search all log repositories (if I guessEngineered correctly ?)
in(LOG_REPOSITORY_ID, repositoryNames));
means the field repositoryId's value is in the list of names repositoryNames.
So if I search all the logs that are in ["foo", "bar", "baz"]
if I get a log with {"repositoryId": "foo"} I will take it, same for other entries of the repositoryNames list "bar" and "baz"

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
NourNuxeo: No, I didn't think about that aspect. I need to check that thanks !

modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/test/java/org/nuxeo/drive/fixtures/AuditChangeFinderClusteringEnabledTestSuite.java
NourNuxeo: back in place, now the problem is the following:
the query transferred to MongoDB is in the form 
```
{"repositoryId": {"$in": ["test"]}, "logDate": {"$lt": {"$date": 1598265956818}}}
```
which doesn't work consistently

If I send the same date in the mongo shell as an ISODate
```
{"repositoryId": {"$in": ["test"]}, "logDate": {"$lt": ISODate("2020-08-24T12:45:56.818556+02:00")}}
```
it will work.

**Now the really inconsistent part:**
The date sent is given by the formula: now - 2 * clusteringDelay as milliseconds 🔴 
If I give the date without substracting 2 * clusteringDelay as milliseconds it will work 🟢
If I give the date without substracting 2 * clusteringDelay as **NANOSeconds** it will work 🟢 

**Now more inconsistency:**
I have faced the same problem with upper and lower bound:
The query sent by the java driver was of the form:
{"lowerbound": {"$gt": {"$numberLong": "0"}}} 🟢 
The same query copied in the mongo shell would fail 🔴 
in the mongoShell I had to write: {"lowerbound": {"$gt": 0}} 🟢

modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/test/java/org/nuxeo/drive/fixtures/AuditChangeFinderTestSuite.java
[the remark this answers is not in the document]
ataillefer: Sorry about the links, they must've moved with the PR evolving.

The point is, whatever implementation you take, SQL or Mongo, remove the code that handles the impacted user name in the XXChangefinder and the test will still pass => the test doesn't test what it should.

By debugging, you'll see that the reason why the _foo_ user gets an empty list of changes is not what you think, it's because it simply doesn't have access to `folder2` (no Read ACL). Indeed, the audit entry relevant to this root registration by _Administrator_ is returned by the audit query for _foo_ but is then filtered in `AuditChangeFinder` by the following code:
```
                if (!session.exists(docRef)) {
                    log.debug("Document {} ({}) doesn't exist, not adding entry to the change summary.",
                            entry::getDocPath, () -> docRef);
                    // Deleted or non accessible documents are mapped to
                    // deleted file system items in a separate event: no need to
                    // try to propagate this event.
                    continue;
                }
```
Look at when the `impactedUserName` property is added to the virtual audit event by [NuxeoDriveFileSystemDeletionListener](https://github.com/nuxeo/nuxeo/blob/master/modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/main/java/org/nuxeo/drive/listener/NuxeoDriveFileSystemDeletionListener.java).

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
ataillefer: Oh! So it's a hack for a missing feature (no predicates) in `MongoDBAuditBackend`.
Might be interesting to have a new API in `MongoDBAuditBackend` such as:
```
public List<LogEntry> queryLogs(OrderByList orders, long offset, protected long limit)
```
@efge WDYT?

efge: I'm not sure I understand what the issue is here or why we need a new API. If clustering is disabled then there is no clustering delay stuff and nothing to query about repositories.

NourNuxeo: If I pass an empty predicate here is what happens
```
com.mongodb.MongoQueryException: Query failed with error code 2 and error message '$and/$or/$nor must be a nonempty array' on server localhost:27017
	at com.mongodb.operation.FindOperation$1.call(FindOperation.java:735)
	at com.mongodb.operation.FindOperation$1.call(FindOperation.java:725)
	at com.mongodb.operation.OperationHelper.withReadConnectionSource(OperationHelper.java:463)
	at com.mongodb.operation.FindOperation.execute(FindOperation.java:725)
	at com.mongodb.operation.FindOperation.execute(FindOperation.java:89)
	at com.mongodb.client.internal.MongoClientDelegate$DelegateOperationExecutor.execute(MongoClientDelegate.java:196)
	at com.mongodb.client.internal.MongoIterableImpl.execute(MongoIterableImpl.java:143)
	at com.mongodb.client.internal.MongoIterableImpl.iterator(MongoIterableImpl.java:92)
	at com.mongodb.client.internal.MongoIterableImpl.iterator(MongoIterableImpl.java:39)
	at java.base/java.lang.Iterable.spliterator(Iterable.java:101)
	at org.nuxeo.mongodb.audit.MongoDBAuditBackend.buildLogEntries(MongoDBAuditBackend.java:371)
	at org.nuxeo.mongodb.audit.MongoDBAuditBackend.queryLogs(MongoDBAuditBackend.java:186)
	at org.nuxeo.drive.mongodb.MongoDBAuditChangeFinder.getUpperBound(MongoDBAuditChangeFinder.java:119)
```

ataillefer: This is because the [QueryBuilder](https://github.com/nuxeo/nuxeo/blob/master/modules/core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/sql/model/QueryBuilder.java#L47) is initialized with and AND filter and, in [MongoDBAuditBackend](https://github.com/nuxeo/nuxeo/blob/master/modules/platform/nuxeo-platform-audit/nuxeo-platform-audit-mongodb/src/main/java/org/nuxeo/mongodb/audit/MongoDBAuditBackend.java#L178), we always apply the filters.

So, currently, there's no way to query the audit logs without any predicate through a `QueryBuilder`.

efge: > So, currently, there's no way to query the audit logs without any predicate through a `QueryBuilder`.

Yes because that's never been a use case. As a quick workaround if the `repositoryNames` is empty let's just replace the predicate with something like `gt(eventDate, new Date(0))` that should match everything and let us move forward, we'll refine it later.

efge (changes requested):

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/test/java/org/nuxeo/drive/mongodb/TestMongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Runs the {@link AbstractChangeFinderTestCase} implementations using the {@link MongoDBAuditChangeFinder}.
+ *
+ * @since 11.3
+ */
+@RunWith(ContributableFeaturesRunner.class)
+@Features(NuxeoDriveMongoDBAuditFeature.class)
+@SuiteClasses({ AuditChangeFinderClusteringEnabledTestSuite.class,
+        })
```
ataillefer, line 37: Don't forget to add back the other suites!

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,184 @@
...
+    public long getUpperBound(Set<String> repositoryNames) {
+        long clusteringDelay = getClusteringDelay(repositoryNames);
+        MongoDBAuditBackend auditService = (MongoDBAuditBackend) Framework.getService(AuditReader.class);
+        QueryBuilder queryBuilder = new AuditQueryBuilder().predicate(
+                Predicates.in(LOG_REPOSITORY_ID, repositoryNames));
+        if (clusteringDelay > -1) {
+            // Double the delay in case of overlapping, see https://jira.nuxeo.com/browse/NXP-14826
+            ZonedDateTime lastClusteringInvalidationDate = ZonedDateTime.now()
+                                                                        .minus(2 * clusteringDelay, ChronoUnit.MILLIS);
+            log.error("preparing upperBound millis: {}", lastClusteringInvalidationDate);
```
ataillefer, line 114: Don't forget to remove or change the log level.

modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/test/java/org/nuxeo/drive/fixtures/AuditChangeFinderTestSuite.java
```diff
@@ -79,6 +79,25 @@
     @Inject
     protected TrashService trashService;
 
+    /** @since 11.3 */
+    @Test
+    public void testImpactedUser() {
+        log.trace("Register a sync root for Administrator");
+        nuxeoDriveManager.registerSynchronizationRoot(session.getPrincipal(), folder1, session);
+        txFeature.nextTransaction();
```
ataillefer, line 87: You could have the newline here, as suggested initially, to separate the actions from the expected check, as in the other tests.

modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/test/java/org/nuxeo/drive/fixtures/AuditChangeFinderTestSuite.java
```diff
@@ -79,6 +79,25 @@
...
+        // Check changes, expecting 2:
+        // - rootRegistered for folder1
+        // - documentCreated for folder1
+        List<FileSystemItemChange> changes = getChanges();
+        assertEquals(2, changes.size());
```
ataillefer, line 92: Please add a newline.

modules/platform/nuxeo-drive-server/nuxeo-drive-core/src/test/java/org/nuxeo/drive/fixtures/AuditChangeFinderTestSuite.java
```diff
@@ -79,6 +79,25 @@
...
+        log.trace("Unregister the sync root for Administrator");
+        nuxeoDriveManager.unregisterSynchronizationRoot(session.getPrincipal(), folder1, session);
+        txFeature.nextTransaction();
```
ataillefer, line 95: Please add a newline.

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,181 @@
...
+    public long getUpperBound(Set<String> repositoryNames) {
+        long clusteringDelay = getClusteringDelay(repositoryNames);
+        MongoDBAuditBackend auditService = (MongoDBAuditBackend) Framework.getService(AuditReader.class);
+        QueryBuilder queryBuilder = new AuditQueryBuilder().predicate(
+                Predicates.gt("eventDate", new Date(0)));
+        if (clusteringDelay > -1) {
+            // Double the delay in case of overlapping, see https://jira.nuxeo.com/browse/NXP-14826
+            ZonedDateTime lastClusteringInvalidationDate = ZonedDateTime.now()
+                                                                        .minus(2 * clusteringDelay, ChronoUnit.MILLIS);
+            log.debug("preparing upperBound millis: {}", lastClusteringInvalidationDate);
```
ataillefer, line 113: I'd say either remove or be more precise, think about the admin who will check the logs.
The message could simply be: "Last clustering invalidation date in milliseconds for the upper bound check: {}"

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,181 @@
...
+import org.nuxeo.mongodb.audit.MongoDBAuditBackend;
+import org.nuxeo.mongodb.audit.MongoDBAuditEntryReader;
+import org.nuxeo.runtime.api.Framework;
+
+import com.mongodb.client.MongoCollection;
+
+/**
+ * Override the JPA audit based change finder to execute query in BSON.
+ * <p>
+ * The structure of the query executed by the {@link AuditChangeFinder} is:
```
efge, line 65: Actually this whole comment and the query examples (which are JPA-specific) are not useful in a Javadoc-level comment. And I suspect they're not up to date anymore anyway

efge (changes requested):

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
NourNuxeo: It happens the reverted index is kinda built in. This explains why one can sort or desc _id easily.
Tested it with 10 millions documents anyway and the result was instant.

ataillefer (approved):

ataillefer:
Good job, thanks.
Now let's test this  in a server configured with mongodb-audit and Nuxeo Drive!

NourNuxeo:
> Good job, thanks.
> Now let's test this in a server configured with mongodb-audit and Nuxeo Drive!

Thanks for your patience and help!

modules/platform/nuxeo-drive-server/nuxeo-drive-mongodb/src/main/java/org/nuxeo/drive/mongodb/MongoDBAuditChangeFinder.java
```diff
@@ -0,0 +1,156 @@
...
+     * Returns the last available log id in the audit collection considering events older than the last clustering
+     * invalidation date if clustering is enabled for at least one of the given repositories. This is to make sure the
+     * {@code DocumentModel} further fetched from the session using the audit entry doc id is fresh.
+     */
+    @Override
+    public long getUpperBound(Set<String> repositoryNames) {
+        long clusteringDelay = getClusteringDelay(repositoryNames);
+        MongoDBAuditBackend auditService = (MongoDBAuditBackend) Framework.getService(AuditReader.class);
+        QueryBuilder queryBuilder = new AuditQueryBuilder().predicate(
+                Predicates.gt("eventDate", new Date(0)));
```
efge, line 83: Could you please add a comment explaining why we do this, because we don't have APIs to do a query with no predicate at all? Like `// TODO remove this dummy predicate once we can query with no predicate at all`

efge (changes requested):
One last small change

efge (approved):

ataillefer (approved):

PR #4238, opened 2020-08-05
NXP-29498: Relax OAuth2 redirect URI constraint in dev mode

kevinleturc:


kevinleturc:
I added a unit test to check that everything except empty is valid when dev mode is on.

PR #4239, opened 2020-08-05
NXP-29499: fix error management on invalid operation/chain/codec regi…

atchertchian:
…stration

T&P: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-atchertchian-master/262/

PR #4241, opened 2020-08-06
NXP-29499: fix error management on invalid operation/chain/codec (lazy check on chains)

atchertchian:
T&P check at https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-atchertchian-master-3/13/

### DIFF
--- a/modules/platform/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/AutomationComponent.java
+++ b/modules/platform/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/AutomationComponent.java
@@ -38,7 +38,7 @@ import org.nuxeo.ecm.automation.AutomationFilter;
 import org.nuxeo.ecm.automation.AutomationService;
 import org.nuxeo.ecm.automation.ChainException;
 import org.nuxeo.ecm.automation.OperationException;
-import org.nuxeo.ecm.automation.OperationType;
+import org.nuxeo.ecm.automation.OperationParameters;
 import org.nuxeo.ecm.automation.TypeAdapter;
 import org.nuxeo.ecm.automation.context.ContextHelperDescriptor;
 import org.nuxeo.ecm.automation.context.ContextHelperRegistry;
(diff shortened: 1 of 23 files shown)


atchertchian:
Changed the strategy compared to #4239 to avoid to many changes: requirement is not needed anymore with this check at startup.
Note the important change here is that the chain is registered to the service, while this was prevented with a direct check (but in strict mode, startup will still fail).

modules/platform/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/AutomationComponent.java
```diff
@@ -135,21 +136,25 @@ public void registerContribution(Object contribution, String extensionPoint, Com
...
             }
+            Class<?> type;
             try {
-                service.putOperation(opc.type, opc.replace, contributor.getName().toString(), widgetDefinitionList);
+                type = Class.forName(opc.type);
+            } catch (ClassNotFoundException e) {
+                throw new IllegalArgumentException("Invalid operation class '" + opc.type + "': class not found.");
+            }
+            try {
+                service.putOperation(type, opc.replace, contributor.getName().toString(), widgetDefinitionList);
```
kevinleturc, line 146: What do you think about merging the two try/catch? Something like below:

```java
            try {
                Class<?> type = Class.forName(opc.type);
                service.putOperation(type, opc.replace, contributor.getName().toString(), widgetDefinitionList);
            } catch (ClassNotFoundException e) {
                throw new IllegalArgumentException("Invalid operation class '" + opc.type + "': class not found.");
            } catch (OperationException e) {
                throw new RuntimeException(e);
            }
```

atchertchian: will let the T&P go through like this first, and will make the change after

atchertchian:
Rebased to handle above feedback + reworked the last commit message

modules/platform/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/AutomationComponent.java
```diff
@@ -135,21 +136,23 @@ public void registerContribution(Object contribution, String extensionPoint, Com
                     widgetDefinitionList.add(widgetDescriptor.getWidgetDefinition());
                 }
             }
+            Class<?> type;
```
kevinleturc, line 139: Not sure you need this declaration outside the try/catch?

atchertchian: indeed

atchertchian:
Rebasing to replace "since 11.2" references by "since 11.3" because of release, and ignoring misleading "pending" statuses linked to changes on the checks.

PR #4249, opened 2020-08-10
NXP-29504: runtime startup errors/warnings

atchertchian:


modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/DefaultComponent.java
```diff
@@ -63,6 +65,10 @@ public void deactivate(ComponentContext context) {
         setModifiedNow();
     }
 
+    protected void addRuntimeMessage(Level level, String message) {
+        Framework.getRuntime().getMessageHandler().addMessage(level, message, ComponentManagerImpl.class.getName());
```
kevinleturc, line 69: Shouldn't be getClass instead of ComponentManagerImpl.class?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
```diff
@@ -86,56 +89,43 @@ public void beforeDeactivation(ComponentManager mgr) {
...
+    public List<String> getMessageStrings(Predicate<RuntimeMessage> predicate) {
+        return getMessages(predicate).stream()
+                                     .map(RuntimeMessage::getMessage)
+                                     .collect(Collectors.collectingAndThen(Collectors.toList(),
+                                             Collections::unmodifiableList));
+    }
 
+    @Override
+    public void clear(Predicate<RuntimeMessage> predicate) {
+        final Predicate<RuntimeMessage> p = predicate == null ? m -> true : predicate;
```
kevinleturc, line 127: Can you introduce a dedicated clear method without argument and do not allow null as argument for this one?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
```diff
@@ -86,56 +89,43 @@ public void beforeDeactivation(ComponentManager mgr) {
...
 
-        RUNNING,
+    protected Predicate<RuntimeMessage> getFinalPredicate(Predicate<RuntimeMessage> givenPredicate) {
+        return givenPredicate == null ? m -> true : givenPredicate;
+    }
 
-        STOPPING,
+    @Override
+    public List<RuntimeMessage> getMessages(Predicate<RuntimeMessage> predicate) {
+        final Predicate<RuntimeMessage> p = predicate == null ? m -> true : predicate;
```
kevinleturc, line 111: Can you remove the null argument possibility?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
```diff
@@ -86,56 +89,43 @@ public void beforeDeactivation(ComponentManager mgr) {
...
+        return messages.stream()
+                       .filter(msg -> p.test(msg))
```
kevinleturc, line 113: Just p doesn't work?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
```diff
@@ -86,56 +89,43 @@ public void beforeDeactivation(ComponentManager mgr) {
...
-    /**
-     * @since 9.10
-     */
-    protected enum ComponentManagerStep {
-
-        ACTIVATING,
-
-        STARTING,
+    @Override
+    public void addMessage(Level level, String message, String source) {
```
kevinleturc, line 101: Does it make sense to have instead/in addition an api with Class<? > as source type?

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationTest.java
```diff
@@ -47,18 +49,47 @@ public void testContributions() throws Exception {
         // do nothing
     }
 
+    protected void checkDupe(boolean detected) {
+        List<String> errors = Framework.getRuntime().getMessageHandler().getErrors();
+        if (detected) {
+            assertEquals(List.of("Duplicate component name: service:my.comp2"), errors);
+            System.err.println(Framework.getRuntime().getMessageHandler().getMessages(null).get(0).getSource());
```
kevinleturc, line 56: Forgotten System.err?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/DefaultComponent.java
atchertchian: These are still registration errors, so if we want to cleanup them up on hot reload, they should be linked to the ComponentManager marker

atchertchian:
> Great work, I'm wondering if we should import statically all Level to have something a bit more readable IMHO. What do you think?

I don't have a strong opinion on how we do imports, it's just aesthetics right?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
atchertchian: I hesitated to avoid piling up new API (since nobody would call that clear() method) but i can add it if you fin it better

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
atchertchian: See above comment (same reason)

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
atchertchian: Indeed, got confused by error when i tried to redefine the predicate parameter directly

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
atchertchian: Instead makes sens given the usage :+1:

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationTest.java
atchertchian: Indeed thanks!

atchertchian:
Rebasing after 11.3 release and merge of #4241, introducing on more usage to be replaced.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
atchertchian: Removed the null possibility

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
atchertchian: Removed the null possibility

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
atchertchian: Will check the DeploymentPreprocessor use case (only usage based on a String for now)

atchertchian: Made the changed and used the addMessage(RuntimeMessage) api for the DeploymentPreprocessor use case (as nuxeo-runtime cannot depend on nuxeo-runtime-deploy module without producing a dependency cycle)

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
```diff
@@ -0,0 +1,86 @@
...
+    protected final String message;
+
+    protected final String source;
+
+    public RuntimeMessage(ComponentManagerStep step, Level level, String message) {
+        this(step, level, message, null);
+    }
+
+    public RuntimeMessage(ComponentManagerStep step, Level level, String message, String source) {
+        super();
```
troger, line 42: Not needed.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public String getSource() {
+        return source;
+    }
+
+    protected enum ComponentManagerStep {
+
+        /**
+         * Pseudo state when listener has not been called yet.
```
troger, line 68: Which listener?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
```diff
@@ -0,0 +1,86 @@
...
+         *
+         * @since 11.3
```
troger, line 70: Not needed, the whole class is `@since 11.3`.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
```diff
@@ -41,12 +48,51 @@
...
      */
+    @Deprecated
     void addError(String message);
 
     /**
      * @return an unmodifiable {@link List} of error messages
      */
     List<String> getErrors();
 
+    void addMessage(RuntimeMessage message);
```
troger, line 63: Missing Javadoc/since.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
```diff
@@ -41,12 +48,51 @@
...
+
+    /**
+     * Adds the following message.
+     * <p>
+     * The source is a free string identifier allowing to distinguish between messages (and to clear them selectively if
```
troger, line 68: It's not a `free string identifier` :)

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
```diff
@@ -41,12 +48,51 @@
...
+     * Returns all messages, filtered by following predicate.
+     * <p>
+     * Predicate can be null, in which case no filtering will be done.
+     *
+     * @since 11.3
+     */
+    List<RuntimeMessage> getMessages(Predicate<RuntimeMessage> predicate);
+
+    /**
+     * Returns all messages strings, filtered by following predicate.
```
troger, line 85: Predicate could be `null`, like for `#getMessages`?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
```diff
@@ -41,12 +48,51 @@
...
+     *
+     * @since 11.3
+     */
+    List<String> getMessageStrings(Predicate<RuntimeMessage> predicate);
+
+    /**
+     * Clears messages filtered by following predicate.
```
troger, line 92: What if `predicate` is `null`?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
```diff
@@ -86,56 +89,37 @@ public void beforeDeactivation(ComponentManager mgr) {
...
+    @Override
+    public void addMessage(Level level, String message, Class<?> source) {
+        addMessage(new RuntimeMessage(currentStep, level, message, source == null ? null : source.getName()));
+    }
 
-        STOPPING,
+    @Override
+    public List<RuntimeMessage> getMessages(Predicate<RuntimeMessage> predicate) {
+        return messages.stream()
+                       .filter(msg -> predicate.test(msg))
```
troger, line 108:
```suggestion
                       .filter(predicate)
```
?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
```diff
@@ -86,56 +89,37 @@ public void beforeDeactivation(ComponentManager mgr) {
...
+    public List<String> getMessageStrings(Predicate<RuntimeMessage> predicate) {
+        return getMessages(predicate).stream()
+                                     .map(RuntimeMessage::getMessage)
+                                     .collect(Collectors.collectingAndThen(Collectors.toList(),
+                                             Collections::unmodifiableList));
+    }
 
+    @Override
+    public void clear(Predicate<RuntimeMessage> predicate) {
+        messages.removeIf(msg -> predicate.test(msg));
```
troger, line 122:
```suggestion
        messages.removeIf(predicate);
```

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java
```diff
@@ -501,7 +496,9 @@ public synchronized void unregisterServices(RegistrationInfo ri) {
 
     protected static void handleError(String message, Exception e) {
         log.error(message, e);
-        Framework.getRuntime().getMessageHandler().addWarning(message);
+        Framework.getRuntime()
+                 .getMessageHandler()
+                 .addMessage(Level.SEVERE, message, ComponentManagerImpl.class);
```
troger, line 501: Why not using `#getClass()` like in other changes?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
```diff
@@ -342,7 +343,9 @@ protected ComponentInstance createComponentInstance() {
             String msg = "Failed to instantiate component: " + implementation;
             log.error(msg, e);
             msg += " (" + e.toString() + ')';
-            Framework.getRuntime().getMessageHandler().addError(msg);
+            Framework.getRuntime()
+                     .getMessageHandler()
+                     .addMessage(Level.SEVERE, msg, ComponentManagerImpl.class);
```
troger, line 348: Seems not clear to me why we are using `ComponentManagerImpl.class` here instead of `#getClass()` or `RegistrationInfoImpl.java`.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
atchertchian: Needed for the DeploymentPreprocessor use case, see comments on resolved conversation above

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
atchertchian: RuntimeMessageHandler is a runtime listener. Maybe we don't need this info anymore, wdyt @kevinleturc ?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
atchertchian: Yes but still a new behaviour compared to previous states, anyway i can remove it.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
atchertchian: Forgot to update the javadoc after latest changes indeed

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
atchertchian: Not anymore, see previous comments from Kevin: i removed this possibility

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
atchertchian: see previous comments

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java
atchertchian: because it's a static method

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
atchertchian: because we need to clear all these in case of hotreload/tests, but i can reference both classes here if you prefer: https://github.com/nuxeo/nuxeo/blob/fix-NXP-29504-runtime-startup-errors/modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java#L899

note i'll need to rework this anyway for https://jira.nuxeo.com/browse/NXP-28948, i was trying to merge minimal changes before refactoring the DescriptorRegistry api to handle things more finely (if we want to).

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
troger: 🤔 
I was just talking about the `super();`, not the whole constructor (?)

atchertchian: ah ok

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
kevinleturc: We can get rid of the listener aspect there indeed.

The new state `INITIALIZING` represents the ComponentManager state before it starts its activation and it is only reachable just after the ComponentManager instantiation right?

We can reword the comment in this way 👍 .

atchertchian:
> I'm not a fan of exposing `java.util.logging.Level` everywhere... we end up with `java.util.logging` _and_ `org.apache.logging.log4j` in classes.
> The use of `java.util.logging.Level` should be an implementation detail and we should expose our own "runtime log level" with levels we really use (I can see `Level.SEVERE` and `Level.WARNING` only).
> 
> As we deprecate the uses of `RuntimeMessageHandler#addWarning` and `RuntimeMessageHandler#addError`, I would have expected that we also deprecate the uses of `RuntimeMessageHandler#getWarnings` and `RuntimeMessageHandler#getErrors` to be consistent.
> They both could be replaced with something like `RuntimeMessageHandler#getMessages(level)`.

So would it be ok to define a custom level for this, and undeprecate addError and addWarning?
Or deprecated getWarnings and getErrors too instead?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
atchertchian: If we don't need this to be a listener, i'll get rid of the whole "step" information anyway. So which is it? Do you think we'll need to know at which step the message has been added?
My understanding is that this info was added to handle dupe messages in case of hotreload, but the strategy is not working anyway, that's why i added the notion of source instead.

kevinleturc: >If we don't need this to be a listener, i'll get rid of the whole "step" information anyway. So which is it? Do you think we'll need to know at which step the message has been added?

We don't really need to know precisely the step, in fact we need to know if it happens during ACTIVATING + STARTING, or before RUNNING in order to not clean them before displaying them.

The source approach is interesting but it requires to manipulate the stack of messages outside of the handler which is not great to handle all usages (runtime/hotreload/test) and requires additional API for this. In addition, see one of my comments, it requires to put `ComponentManagerImpl` as a source for all `Component`. Is it really source? We won't be able to use this new field to know which component reported the message, or am I missing something?

atchertchian: I cannot finely know which component provided the registration without changing the DescriptorRegistry api, that's why i limited changes here by cleaning up all messages from the ComponentManager source, even if i'm not happy with these changes. We're not relying on the step at all anymore: calls to #applyStash is the current criterion i'm using.

I chose to go this way for now because at least we're not hiding important problemes at startup anymore: it fixes the duplicate component name detection at startup.

Now if we want sthg finer, i'll postpone all this work until i get a go to fix the registration in general (as current code is nonsense): this will take time and bring more questions/issues, so this may not happen any time soon.

troger:
> So would it be ok to define a custom level for this, and undeprecate addError and addWarning?
> Or deprecated getWarnings and getErrors too instead?

If we only have error and warning but no other level, I don't really see the point of having a generic method `addMessage(level, ...)`. We could want to stick to `addError` and `addWarning` for simplicity.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
troger: So this need to be removed, right?
https://github.com/nuxeo/nuxeo/pull/4249/files#diff-ea74bf978f35cfe86fc57284c8fc0567R78

atchertchian: Yes, thanks, i will review javadoc once we agree on the rest.

atchertchian:
> > So would it be ok to define a custom level for this, and undeprecate addError and addWarning?
> > Or deprecated getWarnings and getErrors too instead?
> 
> If we only have error and warning but no other level, I don't really see the point of having a generic method `addMessage(level, ...)`. We could want to stick to `addError` and `addWarning` for simplicity.

I'll need to track the originating component/bundle/contrib etc for https://jira.nuxeo.com/browse/NXP-28948, so i went for a generic method to avoid doubling all future API.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/DefaultComponent.java
```diff
@@ -63,6 +65,10 @@ public void deactivate(ComponentContext context) {
         setModifiedNow();
     }
 
+    protected void addRuntimeMessage(Level level, String message) {
```
efge, line 68: I'm good with using `java.util.logging.Level` in this API, as it's a JDK class.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
```diff
@@ -86,56 +89,37 @@ public void beforeDeactivation(ComponentManager mgr) {
...
+    public void addMessage(Level level, String message, Class<?> source) {
+        addMessage(new RuntimeMessage(currentStep, level, message, source == null ? null : source.getName()));
+    }
 
-        STOPPING,
+    @Override
+    public List<RuntimeMessage> getMessages(Predicate<RuntimeMessage> predicate) {
+        return messages.stream()
+                       .filter(msg -> predicate.test(msg))
+                       .collect(Collectors.collectingAndThen(Collectors.toList(), Collections::unmodifiableList));
```
efge, line 109: I know it's a refactoring of pre-existing code, but FYI since Java 10 you can directly do `.collect(Collectors.toUnmodifiableList())`

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/DefaultComponent.java
troger: My issue is that we are mixing 2 logging API (even if it's just for a LogLevel...) in all classes using this API - I don't find it clear.

kevinleturc:
After discussions on slack, we may have found a way to fix the problem without introducing the source of the message which leaves more room to introduce it later for the feature aspect only (needed in [NXP-28948](https://jira.nuxeo.com/browse/NXP-28948)) as it won't be used to fix this.
You can find the patch in this [gist](https://gist.github.com/kevinleturc/31999b8fed9c7f3d8c1343cb77dc1311).

Also, I agree with @troger about `addError`/`addWarning` and `getErrors`/`getWarnings` APIs, deprecation of one should produce deprecation of the other. 
Furthermore, exposing the `java.util.logging.Level` outside of this class is not really a good thing imho. Mainly because we want warnings and errors taxonomy whereas `Level` has `SEVERE` and not `ERROR`, this leads to inconsistency from a usage point of view. The other unused `Level` could also add incomprehension.

atchertchian:
I will rework the PR to take into account @kevinleturc's workaround instead of introducing a source in the message.
I will still try to get rid of the legacy API addError/addWarning because i will need to rework it later anyway, but i will make it consistent to accommodate everyone's feedback.
Please take into account that more changes are needed (and hopefully) coming up around this: i'm still trying to fix a (major, imo) bug before addressing larger issues and needed improvements.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/DefaultComponent.java
atchertchian: source removed --> resolve

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
atchertchian: relying on RUNNING state to clear messages fixed it --> cleaned up the step on messages, but kept the listener logics to be able to clear all messages when *previous* step was RUNNING

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
atchertchian: source removed -> resolve

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/DefaultComponent.java
atchertchian: Introduced a specific level for this, to still avoid duplicating error-related and warning-related api, and hopefully keeping clear semantics

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandlerImpl.java
atchertchian: :+1:

atchertchian:
Please tell if you're ok with these changes (and if so, i'll also replace deprecated getErrors and getWarnings methods usage)

efge:
> Please tell if you're ok with these changes (and if so, i'll also replace deprecated getErrors and getWarnings methods usage)

Ok with these changes

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
```diff
@@ -29,24 +33,64 @@
...
+     * @since 11.3
+     */
+    List<RuntimeMessage> getRuntimeMessages(Predicate<RuntimeMessage> predicate);
+
+    /**
+     * Returns all messages strings, filtered by following predicate.
+     *
+     * @since 11.3
+     */
+    List<String> getMessages(Predicate<RuntimeMessage> predicate);
```
troger, line 87: Worth adding a convenient method `#getMessages(Level)` to ease the replacement of `#getErrors` and `#getWarnings` methods? Make sens to me as we have `#addMessage(Level, String)`.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
```diff
@@ -29,24 +33,64 @@
...
+     * Adds the following message.
+     *
+     * @since 11.3
+     */
+    void addMessage(Level level, String message);
+
+    /**
+     * Returns all messages, filtered by following predicate.
+     * <p>
+     * Predicate can be null, in which case no filtering will be done.
```
troger, line 76: Cannot be `null`, right? 🙃

kevinleturc, line 76: Former javadoc comment I think (?)

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
```diff
@@ -0,0 +1,70 @@
...
+
+        STARTING,
+
+        RUNNING,
+
+        STOPPING,
+
+        DEACTIVATING
+
+    }
```
kevinleturc, line 68: Thanks for the `RuntimeMessage#step` cleanup, I didn't see any usage except the `RuntimeMessageHandlerImpl`, does it worth to leave this class over there?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
```diff
@@ -29,24 +33,64 @@
...
+     * @since 11.3
+     */
+    List<String> getMessages(Predicate<RuntimeMessage> predicate);
+
+    /**
+     * Clears messages filtered by following predicate.
+     *
+     * @since 11.3
+     */
+    void clear(Predicate<RuntimeMessage> predicate);
```
troger, line 94: Seems this method is not used anymore? Do we need it (right now)?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessage.java
atchertchian: Will move it to `RuntimeMessageHandlerImpl` :+1:

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
atchertchian: Removed all api that was not needed right-away

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
atchertchian: Added this api for convenience

atchertchian:
Please re-review as i made some additional changes (fixed some bugs revealed by additional unit tests).
I set aside some invalid behaviours on invalid XML but created https://jira.nuxeo.com/browse/NXP-29547 for them (linked to xmap behaviour).
I also included the trivial fix for https://jira.nuxeo.com/browse/NXP-28914 (Log failure exception in BackingServiceConfigurator).

atchertchian:
(A few force-pushes to handle new lines at end of files...)

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationMockFeature.java
```diff
@@ -0,0 +1,64 @@
...
+import org.nuxeo.runtime.api.Framework;
+import org.nuxeo.runtime.test.runner.FeaturesRunner;
+import org.nuxeo.runtime.test.runner.RunnerFeature;
+
+import com.google.inject.Binder;
+
+/**
+ * @since 11.3
+ */
+public class RuntimeInitializationMockFeature implements RunnerFeature {
```
troger, line 35:
```suggestion
public class RuntimeInitializationErrorMockFeature implements RunnerFeature {
```
Maybe? To make it clear it adds/generates errors.

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationPreprocessorTest.java
```diff
@@ -0,0 +1,43 @@
...
+@RunWith(FeaturesRunner.class)
+@Features({ RuntimeInitializationMockFeature.class, RuntimeFeature.class })
+public class RuntimeInitializationPreprocessorTest {
+
+    @Inject
+    protected HotDeployer hotDeployer;
+
+    @Test
+    public void testDeploymentPreprocessorErrors() throws Exception {
+        // messages have been cleared up by now, see actual test in RuntimeInitializationMockFeature
```
troger, line 40: Worth checking there are no messages?

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationPreprocessorTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.nuxeo.runtime.test.runner.FeaturesRunner;
+import org.nuxeo.runtime.test.runner.HotDeployer;
+import org.nuxeo.runtime.test.runner.RuntimeFeature;
+
+@RunWith(FeaturesRunner.class)
+@Features({ RuntimeInitializationMockFeature.class, RuntimeFeature.class })
+public class RuntimeInitializationPreprocessorTest {
+
+    @Inject
+    protected HotDeployer hotDeployer;
```
troger, line 36: Not used.

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationTest.java
```diff
@@ -47,18 +50,111 @@ public void testContributions() throws Exception {
...
+
+    @Test
+    @Deploy("org.nuxeo.runtime.test.tests:invalid-component.xml")
+    public void testInvalidComponent() throws Exception {
+        assertEquals(List.of(
+                "Bad extension declaration (no target attribute specified) on component 'service:invalid.comp'",
+                "Warning: target extension point 'xp' of 'invalid.comp' is unknown. Check your extension in component service:invalid.comp",
+                "Warning: target extension point 'null' of 'invalid.comp' is unknown. Check your extension in component service:invalid.comp"),
+                Framework.getRuntime().getMessageHandler().getMessages(Level.ERROR));
+        assertEquals(List.of(), Framework.getRuntime().getMessageHandler().getMessages(Level.WARNING));
```
troger, line 112:
```suggestion
        assertTrue(Framework.getRuntime().getMessageHandler().getMessages(Level.WARNING).isEmpty());
```

Same for others `assertEquals(List.of(), ...)`

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationTest.java
```diff
@@ -47,18 +50,111 @@ public void testContributions() throws Exception {
...
-        }
-        assertFalse("An exception should have been raised.", success);
+        checkDupe(true);
+    }
+
+    @Test
+    @Deploy("org.nuxeo.runtime.test.tests:MyComp1.xml")
+    @Deploy("org.nuxeo.runtime.test.tests:MyComp2.xml")
+    @Deploy("org.nuxeo.runtime.test.tests:MyComp2.xml")
+    public void testContributionsWithDuplicateComponentSameFile() throws Exception {
```
troger, line 74: Seems that all the `throws Exception` in methods not using the `hotDeployer` can be removed.

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationPreprocessorTest.java
kevinleturc, line 36: Unused.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
```diff
@@ -27,26 +31,57 @@
...
      * Warning messages don't block server startup.
+     *
+     * @deprecated since 11.3, use {@link #addMessage(Level, String)} instead.
      */
+    @Deprecated
     void addWarning(String message);
 
     /**
      * @return an unmodifiable {@link List} of warning messages
+     * @deprecated since 11.3, use {@link #getMessages(Predicate)} instead
```
troger, line 65: I would have linked to `#getMessages(Level)`.

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationPreprocessorTest.java
kevinleturc, line 40: What about putting the feature as a static class here? It will improve readability.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/RuntimeMessageHandler.java
```diff
@@ -27,26 +31,57 @@
...
      * Error messages block server startup in strict mode.
+     *
+     * @deprecated since 11.3, use {@link #addMessage(Level, String)} instead.
      */
+    @Deprecated
     void addError(String message);
 
     /**
      * @return an unmodifiable {@link List} of error messages
+     * @deprecated since 11.3, use {@link #getMessages(Predicate)} instead
```
troger, line 82: I would have linked to `#getMessages(Level)`.

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationTestComponent.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    protected void actOn(String action) {
+        if (!action.equals(actionTypeOn)) {
+            return;
+        }
+        if ("fail".equals(actionType) && !"activate".equals(actionTypeOn)) { // error at activation will bubble up
+                                                                             // and will not be caught by message
+                                                                             // handler --> skip it
+            throw new RuntimeException("Fail on " + actionTypeOn);
```
kevinleturc, line 61:
```suggestion
        // error at activation will bubble up and will not be caught by message handler --> skip it
        if ("fail".equals(actionType) && !"activate".equals(actionTypeOn)) { 
            throw new RuntimeException("Fail on " + actionTypeOn);
```

modules/runtime/nuxeo-runtime/src/main/resources/OSGI-INF/ConfigurationService.xml
```diff
@@ -37,4 +37,4 @@
     <object class="org.nuxeo.runtime.services.config.ConfigurationPropertyDescriptor" />
   </extension-point>
 
-</component>
\ No newline at end of file
+</component>
```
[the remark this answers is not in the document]
atchertchian: Pushed by mistake, already fixed (did not cleanup anything that was modified later on with actual code changes)

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
```diff
@@ -342,8 +343,8 @@ protected ComponentInstance createComponentInstance() {
             String msg = "Failed to instantiate component: " + implementation;
             log.error(msg, e);
             msg += " (" + e.toString() + ')';
-            Framework.getRuntime().getMessageHandler().addError(msg);
-            throw e;
+            Framework.getRuntime().getMessageHandler().addMessage(Level.ERROR, msg);
+            return null;
```
kevinleturc, line 347: I don't like `null`,  they are often error prone. What do you think about moving the content of this method in the `activate` method?
As it we drop the null thing + we have the full error management (with the early return) at the same place.
Wdyt?

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationPreprocessorTest.java
atchertchian: Was indeed worth checking: some messages have been ignored, some not.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
atchertchian: Ok but it so, it appears more consistent to catch an error at component activation too...
That will change the behaviour described in the RuntimeInitializationTestComponent behaviour (see "error at activation will bubble up and will not be caught by message handler" in my test comments).

We would have in the final RegistrationInfo#activate method the following process:
- try/catch on component instantiation (add error to runtime if so and return)
- try/catch on activation (new behaviour: add error to runtime if so and return instead of bubbling up the error)
- try/catch on each of the extensions contributions
- try/catch on each of the pending contributions

WDYT?

atchertchian: I will include the change so that you can see what it looks like, i can revert it if we're not happy with it or there are unexpected impacts.

atchertchian: (pushed)

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/RuntimeInitializationPreprocessorTest.java
atchertchian: I reverted that change as maven test fail on FeaturesLoader usage of the static inner class:

```
java.lang.IllegalAccessException: class org.nuxeo.runtime.test.runner.FeaturesLoader$Holder cannot access a member of class org.nuxeo.runtime.RuntimeInitializationPreprocessorTest$ErrorFeature with modifiers "public"
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:361)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:591)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:481)
	at org.nuxeo.runtime.test.runner.FeaturesLoader$Holder.<init>(FeaturesLoader.java:54)
	at org.nuxeo.runtime.test.runner.FeaturesLoader.loadFeature(FeaturesLoader.java:124)
	at org.nuxeo.runtime.test.runner.FeaturesLoader.loadFeatures(FeaturesLoader.java:100)
	at org.nuxeo.runtime.test.runner.FeaturesRunner.<init>(FeaturesRunner.java:106)
```

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
kevinleturc: I like it 👍

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
```diff
@@ -574,15 +579,24 @@ public String getImplementation() {
...
-                        "Bad extension declaration (no target attribute specified) on component 's%'", getName());
-                Framework.getRuntime().getMessageHandler().addMessage(Level.WARNING, msg);
+                        "Bad extension declaration (no target attribute specified) on component '%s'", getName());
+                Framework.getRuntime().getMessageHandler().addMessage(Level.ERROR, msg);
                 continue;
+            } else {
+                validExtensions.add(xt);
             }
         }
+        return validExtensions.toArray(Extension[]::new);
```
efge, line 599: Nice, I didn't realize this had been added in Java 11, it's indeed cleaner-looking than `.toArray(new Extension[0])`

PR #4250, opened 2020-08-10
NXP-29515: fix pipeline build after -fae changes

atchertchian:


atchertchian:
See jira issue for all checks done: using " " instead of "" is the only change that worked ok to make the variable defined.
Also please let me know is there's any way to speed up the merge if needed, given the impact.

PR #4257, opened 2020-08-11
NXP-29533: improve build pipeline

troger:


ci/Jenkinsfiles/build.groovy
```diff
@@ -554,16 +559,14 @@ pipeline {
...
-          }
+        container('maven') {
+          echo """
+          ----------------------------------------
+          Run "dev" functional tests
+          ----------------------------------------"""
+          runFunctionalTests('ftests')
         }
-        findText regexp: ".*ERROR.*", fileSet: "ftests/**/log/server.log", unstableIfFound: true
+        findText regexp: ".*ERROR.*", fileSet: "ftests/**/log/server.log"
```
bdelbosc, line 569: Does this means that we now accept errors in server log?

troger: Nop, it means it fails the build, so a failure status, if there are errors in logs.

PR #4270, opened 2020-08-24
NXP-29571 runtime component listeners

atchertchian:


modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java
```diff
@@ -509,14 +510,26 @@ protected static void handleError(String message, Exception e) {
         log.info("Activate components");
         Watch watch = new Watch();
         watch.start();
+
+        // first instantiate resolved components: that allows some to register as listeners on ComponentManager before
+        // all components activation.
+        List<RegistrationInfo> iris = new ArrayList<>();
+        // first activate resolved components
+        for (RegistrationInfo ri : registry.getResolvedRegistrationInfo()) {
+            watch.start(ri.getName().getName());
```
kevinleturc, line 519: The watch calls on this step will be erased by the _activate_ step.
Here we should declare a new Watch, if we're interested, and write its metrics to another file or not do it.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java
```diff
@@ -410,7 +409,9 @@ void sendEvent(ComponentEvent event) {
     public synchronized void registerExtension(Extension extension) {
         ComponentName name = extension.getTargetComponent();
         RegistrationInfo ri = registry.getComponent(name);
-        if (ri != null && ri.getComponent() != null) {
+
+        if (ri != null && ri.getComponent() != null
+                && List.of(RegistrationInfo.ACTIVATED, RegistrationInfo.STARTED).contains(ri.getState())) {
```
kevinleturc, line 414:
```suggestion
                && Set.of(RegistrationInfo.ACTIVATED, RegistrationInfo.STARTED).contains(ri.getState())) {
```

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java
atchertchian: Indeed I thought they would addition (since component instantiation could be costly), so i'll add a new watch.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java
```diff
@@ -531,6 +544,19 @@ protected static void handleError(String message, Exception e) {
         return ris;
     }
 
+    /**
+     * Instantiates the given {@link RegistrationInfo}. This step will instantiate the component.
+     * <p>
+     * Allows registering listeners on ComponentManager at component instantiation, before all components activation.
+     * <p>
+     * Should be called before {@link #activateComponent(RegistrationInfo)}.
```
efge, line 552: You should add a `@return` explaining the boolean returned

efge (changes requested):

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/MockComponentManagerListener.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public List<ComponentEvent> getEvents() {
+        return Collections.unmodifiableList(events);
+    }
+
+    public boolean hasEvent(int event, String component) {
+        return events.stream()
+                     .filter(e -> event == e.id && component.equals(e.registrationInfo.getName().getName()))
+                     .findAny()
+                     .isPresent();
```
kevinleturc, line 54:
```suggestion
        return events.stream()
                     .anyMatch(e -> event == e.id && component.equals(e.registrationInfo.getName().getName()));
```

efge (approved):

kevinleturc (approved):

PR #4278, opened 2020-08-26
NXP-29566: Fix nuxeoctl register prompt

kevinleturc:


ataillefer (approved):

ataillefer:
Note: a comment would be nice to explain why we need to force the linking of stdin, stdout, stderr since we're running a background process.

ataillefer (approved):

bdelbosc (commented):
I find it a bit weird, normally stdout and stderr of a background process don't need to be redirected. 
About the stdin, what is happening if it is close just after the invocation, which should be the case if you run nuxeoctl within an ssh command (`ssh $HOST nuxeoctl CMD`)?

efge:
Agreed, I'm dubious about the fact that `<&0 >&1 2>&2` has any effect, it means redirect fd 0, 1 and 2 to themselves which seems a nop to me.

kevinleturc:
Indeed for `stdout` and `stderr` on background processes, there's only `stdin` which is detached for a background process, so we just need `<&0`. In other words, when the process is in the foreground, its `stdin` under `/proc/PID/fd/0` is linked to the `stdin` of the process launching it. This is not the case when the process is in the background, thus we need to link it with `<$0`.

I will test the ssh command scenario.

kevinleturc:
When running:
```
ssh $HOST nuxeoctl register
```
nuxeoctl will hang and never give back the prompt, I'm trying to understand what can we do.

Otherwise, adding the `-t` option is working.

kevinleturc:
I push a change to remove stdout and stderr redirection and an additional commit to make the ssh command work.

efge:
`<&0` means copy fd `0` to the fd for stdin, which is 0. So it's a nop. Where does this idiom come from and are you sure it works?

ataillefer (approved):

troger (approved):

PR #4280, opened 2020-08-28
Fix nxp 29596 fail pr build on dev unit tests failure

ataillefer:


kevinleturc (approved):

atchertchian (approved):
:+1: Looks good, the wording feels a bit weird to me though, i would have replaced "even if" by "when" (but this seems temporary anyway, so please dicard)

ci/Jenkinsfiles/build.groovy
```diff
@@ -295,7 +295,9 @@ def buildUnitTestStage(env) {
             }
           } catch(err) {
             // TODO NXP-29512: workaround to know which env is failing later on
-            throw new Exception("${env}: ${err.message}", err)
+            def errorMessage = "${env}: ${err.message}"
+            echo "Throwing error with message: ${errorMessage}"
+            error "${errorMessage}"
```
NourNuxeo, line 300: https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#error-error-signal
sais error aborts the build and avoids printing the stack trace. But since you are printing the error message anyway, why not sticking with the one-liner throw?

ci/Jenkinsfiles/build.groovy
```diff
@@ -536,10 +538,17 @@ pipeline {
...
+            // TODO NXP-29512: on a PR, make the build continue even if there is a test error
             // on other environments than the dev one
             // to remove when all test environments will be mandatory
-            if (!isPullRequest() || err.message.startsWith("dev:")) {
+            def errorMessage = err.message
+            if (isPullRequest() && !errorMessage.startsWith("dev:")) {
+              echo """
+                Unit tests error: ${errorMessage}
+                Waiting for NXP-29512, on a PR, the build continues even if there is a unit test error in other
+                environments than the dev one.
```
NourNuxeo, line 549:
```suggestion
                While waiting for NXP-29512 to be resolved, the build in non-dev environment continues even if there are unit test error.
```

ci/Jenkinsfiles/build.groovy
ataillefer: The reason of the change is all explained in the [commit message](https://github.com/nuxeo/nuxeo/pull/4280/commits/63db157b19a4f050fcbab3ccee62d6848e7eeaf5).

The `echo` here is for debugging purpose, to output the custom message that we're building.

The error will then be thrown and catched by the surrounding `try/catch` block: https://github.com/nuxeo/nuxeo/blob/fix-NXP-29596-fail-PR-build-on-dev-unit-tests-failure/ci/Jenkinsfiles/build.groovy#L540, that might decide to continue or to rethrow, in which case the error message will be finally printed at the end of the build logs.

ci/Jenkinsfiles/build.groovy
ataillefer: > on a PR

Cannot remove this, it's representative, look at the actual test in the `if`.

For the the beginning of the sentence, not sure about the added value: suggestion is longer, we try to keep things short.

For the end, it is not that clear: "the build in non-dev environment continues" suggests that only the build that would take place in the dev environment (so the "dev unit tests stage"?) would continue. In fact, we're talking about the whole build.

ci/Jenkinsfiles/build.groovy
NourNuxeo: Ok I get it. So this echo will be done as long as we catch and if in dev environment, the message will be printed at the end of the logs to facilitate seeing  it

ataillefer: If not on a PR and in dev yes.
The error is then simply propagated as for any kind of failure, so the pipeline will naturally make the build fail and display the error message at the end of the logs.

PR #4296, opened 2020-09-03
NXP-29612: don't relax when installing packages for Docker build

atchertchian:
… Dockerfile

### DIFF
--- a/docker/slim/docker-entrypoint.sh
+++ b/docker/slim/docker-entrypoint.sh
@@ -64,7 +64,7 @@ fi
 # Handle NUXEO_PACKAGES
 if [ -n "$NUXEO_PACKAGES" ]; then
   echo "ENTRYPOINT: Install Nuxeo packages: $NUXEO_PACKAGES"
-  nuxeoctl mp-install $NUXEO_PACKAGES --accept=true
+  nuxeoctl mp-install $NUXEO_PACKAGES --accept=true --relax no
 fi
 
 exec "$@"
--- a/docker/slim/install-local-packages.sh
+++ b/docker/slim/install-local-packages.sh
@@ -21,7 +21,7 @@ if [ -n "$packages" ]; then
 $packages
 EOF
   echo
-  NUXEO_CONF=$NUXEO_HOME/bin/nuxeo.conf $NUXEO_HOME/bin/nuxeoctl mp-install --accept yes --nodeps $packages
+  NUXEO_CONF=$NUXEO_HOME/bin/nuxeo.conf $NUXEO_HOME/bin/nuxeoctl mp-install --accept yes --relax no --nodeps $packages
 else
   echo 'Found no packages to install.'
 fi
--- a/docker/slim/install-packages.sh
+++ b/docker/slim/install-packages.sh
@@ -43,7 +43,7 @@ if [ -n "$connect_url" ]; then
 fi
 
 echo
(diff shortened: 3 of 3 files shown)


bdelbosc (dismissed):

bdelbosc:
Note that it is possible that this is going to break CI build, I am thinking of SNAPSHOT package, unreleased packages, or packages served by preprod connect

atchertchian:
> Note that it is possible that this is going to break CI build, I am thinking of SNAPSHOT package, unreleased packages, or packages served by preprod connect

SNAPSHOT packages are deployed to prod, so it should be ok (at least for packages part of the platform).
If this breaks CI build, i think it'll highlight cases where we're not handling target version mapping correctly on the packages for CI and releases (promoted or not), that's why i think it's important not to relax by default (and i can change the default value to limit the damage if we see a lot of issues, waiting for adjustments).

Note also that PR checks went through ok.

That being said i'm not sure where we use this command by default in nuxeo: i thought it was good practice to rely on install-local-packages (left relaxed by default...) for functional testing (apart from nuxeo-explorer, for which i'm making that change).

atchertchian:
Also note "relaxing" has been a problem on our cloud-related automation (similarly to strict mode enablement) so i linked more history to the jira issue.

NourNuxeo (approved):

bdelbosc:
I agree, I'm not saying don't do it, but we have to be prepared to deal with breakage -> wait for Monday?

ataillefer:
@bdelbosc this shouldn't break anything in the CI since it's never calling the impacted script [install-packages.sh](https://github.com/nuxeo/nuxeo/blob/master/docker/slim/install-packages.sh), that is only embedded in the Nuxeo Docker image for the purpose of building a new image on the top of it.

ataillefer:
@atchertchian :+1: for not relaxing by default, indeed, we didn't pay attention about the `--accept yes` => `--relax yes` relationship.

Yet, I think that we don't even want it configurable: we should never relax (as you're stating, it's error prone) and there doesn't seem to be any use case.
So, WDYT about simply adding `--relax no`?

Also, it would be consistent to also add `--relax no` in:
- [install-local-packages.sh](https://github.com/nuxeo/nuxeo/blob/master/docker/slim/install-local-packages.sh#L24): local package installation at build time
- [docker-entrypoint.sh](https://github.com/nuxeo/nuxeo/blob/master/docker/slim/docker-entrypoint.sh#L67): package installation at runtime

Thanks

atchertchian:
> @atchertchian +1 for not relaxing by default, indeed, we didn't pay attention about the `--accept yes` => `--relax yes` relationship.
> 
> Yet, I think that we don't even want it configurable: we should never relax (as you're stating, it's error prone) and there doesn't seem to be any use case.
> So, WDYT about simply adding `--relax no`?
> 
> Also, it would be consistent to also add `--relax no` in:
> 
>     * [install-local-packages.sh](https://github.com/nuxeo/nuxeo/blob/master/docker/slim/install-local-packages.sh#L24): local package installation at build time
> 
>     * [docker-entrypoint.sh](https://github.com/nuxeo/nuxeo/blob/master/docker/slim/docker-entrypoint.sh#L67): package installation at runtime
> 
> 
> Thanks

I'm all for it, but as of not making this configurable, I'll wait for @bdelbosc's go on the 10b benchmark package use case: I don't mind keeping it configurable if needed in the future.

bdelbosc:
indeed the benchmark 10b use case can be easily fixed, let's be strict

ataillefer (approved):

troger (approved):

PR #4300, opened 2020-09-04
NXP-29620: Add an external scroller mechanism to BAF

kevinleturc:


modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkService.java
```diff
@@ -80,4 +81,23 @@
...
+     * @param bucket the bucket to send to a bulk action
+     * @since 11.3
+     */
+    void appendExternalBucket(BulkBucket bucket);
+
+    /**
+     * Completes a running bulk computation having an external scroller.
+     *
+     * @param commandId the command id
+     * @param count the number of elements produced by the scroller
```
bdelbosc, line 98: the **total** number of elements produced by the **external** scroller

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
```diff
@@ -328,4 +332,36 @@ public boolean await(Duration duration) throws InterruptedException {
...
     }
 
+    @Override
+    public void appendExternalBucket(BulkBucket bucket) {
+        String commandId = bucket.getCommandId();
+
+        BulkCommand command = getCommand(commandId);
+        String stream = Framework.getService(BulkAdminService.class).getInputStream(command.getAction());
+
+        String key = commandId; // TODO do we keep a bucket number in the K/V to build this key?
```
bdelbosc, line 342: here we want to load balance processing so better commandId:counter counter can be a global atomic long

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
```diff
@@ -95,6 +98,9 @@
     // How long we keep the command and its status in the kv store once aborted
     public static final long ABORTED_TTL_SECONDS = 7_200;
 
+    // @since 11.3
+    public static final AtomicLong EXTERNAL_SCROLLER_COUNTER = new AtomicLong();
```
bdelbosc, line 102: protected is enough

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
```diff
@@ -162,9 +168,11 @@ protected void checkIfScrollerExists(BulkCommand command) {
                     GenericScrollRequest.builder(command.getScroller(), command.getQuery()).build())) {
                 throw new IllegalArgumentException("Unknown Generic Scroller for command: " + command);
             }
-        } else if (!scrollService.exists(
-                DocumentScrollRequest.builder(command.getQuery()).name(command.getScroller()).build())) {
-            throw new IllegalArgumentException("Unknown Document Scroller for command: " + command);
+        } else if (!command.useExternalScroller()) {
```
bdelbosc, line 171: maybe easier to understand with a dedicated check like ```if(command.useExternalScroller()) {return;}```

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
```diff
@@ -328,4 +329,30 @@ public boolean await(Duration duration) throws InterruptedException {
                  .collect(Collectors.toList());
     }
 
+    @Override
+    public void appendExternalBucket(BulkBucket bucket) {
+        String commandId = bucket.getCommandId();
+
+        BulkCommand command = getCommand(commandId);
```
bdelbosc, line 336: This means a KV store roundtrip for each bucket to get the action name which is a potential perf problem.
A possible solution is to have a very basic cache on getCommand (just keep the last entry), 
or to provide a different method signature appendExternalBucket(String action, BulkBucket bucket), because the scroller should have the knowledge of the action.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
```diff
@@ -328,4 +329,30 @@ public boolean await(Duration duration) throws InterruptedException {
...
+    @Override
+    public void completeExternalScroll(String commandId, long count) {
+        BulkStatus delta = BulkStatus.deltaOf(commandId);
+        delta.setState(RUNNING);
+        delta.setScrollEndTime(Instant.now());
+        delta.setTotal(count);
+
+        Record record = Record.of(commandId, BulkCodecs.getStatusCodec().encode(delta));
+
+        log.debug("Complete external scroll with key: {}, record: {}", commandId, record);
```
bdelbosc, line 355: the total count is interesting information to trace

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkCommand.java
```diff
@@ -188,7 +204,8 @@ public void setScroller(String scrollerName) {
          *
          * @param action the registered bulk action name
          * @param nxqlQuery by default an NXQL query that represents the document set to apply the action. When using a
-         *            generic scroller the query syntax is a convention with the scroller implementation.
+         *            generic scroller the query syntax is a convention with the scroller implementation. When using an
+         *            external scroller the field is the total number of elements if known.
```
bdelbosc, line 208: we need to rename nxqlQuery into query, but I don't see how the total number of elements for an external scroller is implemented

kevinleturc: It is not currently, and I will remove this part from the comment as the count is not known at the beginning of our scroller.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
kevinleturc: I chose to use a `PassiveExpiringMap` with a TTL of 1 minute because, in the context of the 10B benchmark, we may have two commands at the same time, one for each repository.

bdelbosc (approved):

ataillefer (approved):

bdelbosc (approved):

PR #4303, opened 2020-09-07
NXP-29626: fix FormData#getFormFields NPE when request contains binar…

atchertchian:
…y content


NourNuxeo (approved):

NourNuxeo (commented):
How about unit testing the change?

atchertchian:
> How about unit testing the change?

Of course, i would have gladly added a non-regression unit test.
And then i looked at the src/test/java content, and i felt that i already had a fair share of initiating tests on legacy code these days ;)

More seriously, this is a tiny part of the 4 PRs that need to be merged to achieve the actual goal: i'll see if i can do something easily, otherwise i'll probably defer this bugfix and use the non-buggy non-generic api to retrieve form field values.

Also, i'm pretty sure that this method has never really been used since it was created 10 years ago (very standard use case of multipart form fields retrieval, actually buggy when there is binary content in the request...)

atchertchian:
Done, note that unit test did not reproduce the issue with "real-life" org.apache.commons.fileupload.disk.DiskFileItem usage

NourNuxeo (approved):
💪 💪

bdelbosc (approved):

PR #4321, opened 2020-09-15
Fix NXP-29618 rate limit exceed gcp

kevinleturc:


troger (approved):

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-gcp/src/main/java/org/nuxeo/ecm/core/storage/gcp/GoogleStorageBinaryManager.java
```diff
@@ -180,11 +180,19 @@ protected FileStorage getFileStorage() {
...
-        public void storeFile(String key, File file) {
-            try {
-                bucket.create(bucketPrefix + key, new FileInputStream(file));
-            } catch (IOException e) {
-                throw new NuxeoException(e);
+        public void storeFile(String digest, File file) {
+            long t0 = System.currentTimeMillis();
+            log.debug("Storing blob with digest: {} to GCS", digest);
+            String key = bucketPrefix + digest;
+            if (bucket.get(key) == null) {
```
efge, line 187: I guess that `bucket.get(key)` only gets metadata and does not get the full blob stream, right?

efge (dismissed):

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-gcp/src/main/java/org/nuxeo/ecm/core/storage/gcp/GoogleStorageBinaryManager.java
kevinleturc: Indeed, the blob content is retrieved through the `Blob#downloadTo` method (or `reader`). I will add  a comment to state that.

efge (approved):

troger (approved):

PR #4348, opened 2020-09-29
NXP-29642: Use an non-interrupted Thread to close Kafka consumers

kevinleturc:


modules/runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/computation/log/ComputationRunner.java
```diff
@@ -215,7 +215,9 @@ public void run() {
                 log.debug(msg);
             }
         } catch (Exception e) {
-            if (Thread.currentThread().isInterrupted()) {
+            if (Thread.interrupted()) {
+                // clearing the interrupt flag is wanted as closeTailer method needs an non-interrupted thread
+                interrupted = true; // Thread.currentThread().interrupt() in finally
```
bdelbosc, line 220: This will close the consumer without throwing errors, but this is not the cause of NXP-29642 right?

kevinleturc: Indeed, the proposal didn't change this aspect, we were leading there before as `Thread.currentThread().isInterrupted()` should return the same thing as `Thread.interrupted()`, or I am missing something.

I think it might be related to [NXP-29642](https://jira.nuxeo.com/browse/NXP-29642) as the last log we've got is about a `SleepWork` timeout which is owned by another test. Furthermore, the assertion is about `2 != 1` where 2 is scheduled + running, I'm suspecting to have 1 for running due to a SleepWork still there.

I will also get more feedback by running this code, there may be a shallowed exception somewhere.

bdelbosc (approved):

efge (approved):

PR #4351, opened 2020-10-01
Fix NXP-29705 StreamWorkManager pool termination

bdelbosc:


modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/utils/ExceptionUtils.java
```diff
@@ -27,6 +27,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
+
```
kevinleturc, line 30: Extra line.

kevinleturc (dismissed):

kevinleturc (approved):

PR #4356, opened 2020-10-02
NXP-21725: make LongProperty wrap any NumberFormatException

troger:


modules/platform/nuxeo-binary-metadata/src/main/java/org/nuxeo/binary/metadata/internals/BinaryMetadataServiceImpl.java
```diff
@@ -34,6 +34,8 @@
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
```
kevinleturc, line 36: These ones should be cleared because unused.

kevinleturc (dismissed):

efge (changes requested):
I feel like this fix is incomplete if we don't do similar changes to `DoubleProperty`...

kevinleturc (approved):

kevinleturc (approved):

efge (approved):

PR #4357, opened 2020-10-05
Feature nxp 29659 pkg tp version range

yjulienne-nuxeo:


modules/runtime/nuxeo-connect-standalone/src/main/java/org/nuxeo/connect/update/xml/PackageDefinitionImpl.java
```diff
@@ -82,8 +82,15 @@
     /**
      * The target platforms where this package may be installed.
      */
+    @Deprecated
```
kevinleturc, line 85: Could you add a deprecated tag since 11.3 in the javadoc?

modules/runtime/nuxeo-connect-standalone/src/main/java/org/nuxeo/connect/update/xml/PackageDefinitionImpl.java
```diff
@@ -82,8 +82,15 @@
...
     protected String[] platforms;
 
+    /**
+     * The target platform range where this package may be installed.
```
kevinleturc, line 89: Could you add the `@since 11.3` tag to methods/fields you add?

kevinleturc (commented):
Thanks for the contribution 👍.

modules/runtime/nuxeo-connect-standalone/src/main/java/org/nuxeo/connect/update/xml/PackageDefinitionImpl.java
yjulienne-nuxeo: Forgot to remove it, it should not be deprecated as it will still be used by LTS packages, good catch

troger (commented):
Worth adding a test with
```
<target-platform>11.2</target-platform>
```
or in the JSON
```
"targetPlatformRange": "11.2",
```
?

We will probably update all packages inside `nuxeo/nuxeo` to use this format.

modules/runtime/nuxeo-connect-standalone/src/test/java/org/nuxeo/launcher/connect/TestConnectBroker.java
```diff
@@ -897,6 +991,85 @@ public void testInstallLocalPackageRequest() {
         assertThat(logOf(logCaptureResult)).isEqualTo(expectedLogs);
     }
 
+    @Test
+    @LogCaptureFeature.FilterWith(PkgRequestLogFilter.class)
+    public void testInstallLocalPackageRequestWithRange() throws Exception {
+        // GIVEN current target platform is server-11.2
```
troger, line 997:
```suggestion
        // GIVEN current target platform is server-11.2.3
```
?

troger (commented):
My above comment seems still valid (test with `<version>11.2</version>` if it's not required to have `[11.2]` :)

BTW, I'm a bit concerned by having different name between the `package.xml` and the packages JSON:
```
"targetPlatformRange": ...
```
VS
```
<version>...</version>
```

Don't we want to align on the same naming?

glefevre (approved):

yjulienne-nuxeo:
> My above comment seems still valid (test with `<version>11.2</version>` if it's not required to have `[11.2]` :)

The parsing of the version range is unit-tested in the nuxeo-connect-client where the PlatformVersionRange class is, but I can add one in the TestConnectBroker, sure
> 
> BTW, I'm a bit concerned by having different name between the `package.xml` and the packages JSON:
> 
> ```
> "targetPlatformRange": ...
> ```
> 
> VS
> 
> ```
> <version>...</version>
> ```
> 
> Don't we want to align on the same naming?

Might be doable yes ;)

modules/runtime/nuxeo-connect-standalone/src/main/java/org/nuxeo/launcher/connect/ConnectBroker.java
```diff
@@ -1181,12 +1182,12 @@ public boolean pkgRequest(List<String> pkgsToAdd, List<String> pkgsToInstall, Li
...
                                 "Package(s) %s not available on platform version %s.\n"
                                         + "Do you want to relax the constraint (yes/no)? [no] ",
-                                "no", StringUtils.join(nonCompliantPkg, ", "), actualTargetPlatform);
+                                "no", StringUtils.join(nonCompliantPkg, ", "), actualTargetPlatform.asString());
                     }
 
                     if (Boolean.parseBoolean(relax)) {
                         log.warn("Relax restriction to target platform {} because of package(s) {}",
-                                () -> actualTargetPlatform, () -> String.join(", ", nonCompliantPkg));
+                                () -> actualTargetPlatform.asString(), () -> String.join(", ", nonCompliantPkg));
```
kevinleturc, line 1190:
```suggestion
                                actualTargetPlatform::asString, () -> String.join(", ", nonCompliantPkg));
```

kevinleturc:
> > My above comment seems still valid (test with `<version>11.2</version>` if it's not required to have `[11.2]` :)
> 
> The parsing of the version range is unit-tested in the nuxeo-connect-client where the PlatformVersionRange class is, but I can add one in the TestConnectBroker, sure
> 
> > BTW, I'm a bit concerned by having different name between the `package.xml` and the packages JSON:
> > ```
> > "targetPlatformRange": ...
> > ```
> > 
> > 
> > VS
> > ```
> > <version>...</version>
> > ```
> > 
> > 
> > Don't we want to align on the same naming?
> 
> Might be doable yes ;)

If we need to choose between _version_ and _range_, let's use version (like maven).

yjulienne-nuxeo:
So, honestly, if it really matters to you, I can do the change from targetPlatformRange to targetPlatformVersion but it would require another quite big refactoring and another release of the connect-client (+redo all the testing from connect-client to NOS), plus I'm not sure it will really be clearer as the property IS actually a range... WDYT ? Can I leave it as it is ?

kevinleturc (approved):

kevinleturc (approved):

troger (approved):

PR #4366, opened 2020-10-09
NXP-29476: include MD5 when uploading to S3 bucket with ObjectLock

efge:


modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
```diff
@@ -265,6 +266,9 @@ protected String writeFile(String key, Path file, BlobContext blobContext, Strin
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             throw new NuxeoException(e);
+        } catch (SdkClientException e) {
```
kevinleturc, line 269: What do you think about catching `SdkBaseException`?

kevinleturc (dismissed):

kevinleturc (approved):

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
```diff
@@ -265,6 +266,9 @@ protected String writeFile(String key, Path file, BlobContext blobContext, Strin
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             throw new NuxeoException(e);
+        } catch (SdkBaseException e) {
+            // catch SdkBaseException and not just AmazonServiceException
```
ataillefer, line 270: What do you mean? That it's better to catch a wider exception than the one usually marked as thrown by the SDK API? Why?

PR #4378, opened 2020-10-12
improvement NXP 29686 shell scripts docker runtime

ataillefer:
Allow to run shell scripts when starting the Nuxeo Docker image (ENTRYPOINT).

Also, document utility scripts to install packages in a Docker image built from the Nuxeo image.

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -e
 
+# expand filename patterns which match no files to a null string, rather than themselves
+shopt -s nullglob
+
 # Allow supporting arbitrary user IDs
 if ! whoami &> /dev/null; then
   if [ -w /etc/passwd ]; then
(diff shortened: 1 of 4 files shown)


docker/docker-entrypoint.sh
```diff
@@ -53,6 +56,15 @@ if [[ ! -f $NUXEO_HOME/configured && ! -f $NUXEO_CONF ]]; then
   touch $NUXEO_HOME/configured
 fi
 
+# Handle shell scripts
+echo "ENTRYPOINT: Looking for shell scripts in /docker-entrypoint-initnuxeo.d"
+for f in /docker-entrypoint-initnuxeo.d/*; do
+  case "$f" in
+    *.sh)  echo "Running $f"; . "$f" ;;
```
efge, line 63: `.` does a "source" of the shell script in the same context as the current script. Meaning it can impact execution of the current script if it redefines environment variables for instance.
Another more common option outside of Docker is to just execute the script, but I don't know what's common in the Docker world.

ataillefer: Not aware of any best practices for executing a script from the ENTRYPOINT of a Docker image.
You're right, executing the scripts would probably be safer for the environment, but, it's not easy to achieve since we don't control the permissions on the scripts to execute, like `x`, required for a script execution.
Typically the `/docker-entrypoint-initnuxeo.d` directory can be mounted as a volume from a host directory containing the scripts to execute, in which case the resulting volume is owned by root, whereas the user running the entrypoint script is not root.

efge: Ownership is not a problem for execution. One way to execute the script is just `/bin/sh "$f"`.

efge: Or `/bin/bash` if that's Docker usage.

ataillefer: Interesting, thanks. Let's go for `/bin/bash`.

efge (approved):

kevinleturc (approved):

troger (approved):

PR #4384, opened 2020-10-14
NXP-29759: Make the task endpoint use the UserManagerResolver

charlesboidot:
Fix for NXP-29759: Make the task endpoint use the UserManagerResolver


kevinleturc (approved):

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/NuxeoPrincipal.java
```diff
@@ -163,6 +168,13 @@
      */
     boolean isAdministrator();
 
+    /**
+     * @since 11.4 Used in the case where we don't want to fetch the references, where we won't be able to return
+     *        complete informations on this principal (the missing attributes will be if the principal is an
+     *        Administrator and the groups he is in).
+     */
+    boolean isComplete();
```
efge, line 176: Adding this kind of method on public and important API like that is an architecture change that should be discussed first. Please open a discussion on Slack in `#team-platform` about this, because I'm really not sure it's the best way to move forward.

efge: Cc @nuxeo/platform

efge:
Apart from the architecture aspects I mentioned above, I don't understand from the patch who is the intended user of the new "isComplete" that is being returned in the JSON. Is there an associated ticket on the UI side that shows the goal?

efge:
*If* we want to add additional flags to the principal, then two things:
- I'd prefer if the flag was false by default on a complete instance, so named instead "isIncomplete" or "isPartial". But the doc and maybe the naming should be clear about in what way it's incomplete, what's the difference between a complete and incomplete instance?
- Setting the flag should be done at a lower level than in `UserManagerResolver`. It should probably be done in `UserManagerImpl.makePrincipal` when the caller (higher in the stack) has `fetchReferences=false`.

modules/platform/nuxeo-platform-document-routing/nuxeo-routing-rest-api/src/test/java/org/nuxeo/ecm/restapi/server/WorkflowEndpointTest.java
```diff
@@ -1240,11 +1240,18 @@ public void testTaskWorkflowInfo() throws IOException {
...
-            assertEquals("administrators", groups.get(0).textValue());
+
+            if(initiatorNode.get("isComplete").booleanValue()==true) {
+                assertTrue(initiatorNode.get("isAdministrator").booleanValue());
+                assertEquals(1, groups.size());
+                assertEquals("administrators", groups.get(0).textValue());
+            }
+            else {
+                assertFalse(initiatorNode.get("isAdministrator").booleanValue());
+                assertEquals(0, groups.size());
```
troger, line 1253:
```suggestion
                assertTrue(groups.isEmpty());
```

troger (commented):
I agree with @efge, nothing should be done in `UserManagerResolver`. What about removing the public API but just use a simple flag in the underlying `DocumentModel` through `#putContextData`?

Could be done in `UserManagerImpl#getUserModel` if `fetchReferences` is `false`, and then this flag will be read by the `NuxeoPrincipalWriter` to set accordingly the `isPartial` field?

charlesboidot:
> _If_ we want to add additional flags to the principal, then two things:
> 
>     * I'd prefer if the flag was false by default on a complete instance, so named instead "isIncomplete" or "isPartial". But the doc and maybe the naming should be clear about in what way it's incomplete, what's the difference between a complete and incomplete instance?
> 
>     * Setting the flag should be done at a lower level than in `UserManagerResolver`. It should probably be done in `UserManagerImpl.makePrincipal` when the caller (higher in the stack) has `fetchReferences=false`.

The difference between a complete and a partial principal are that in the case where we don't want to fetch the references we can't tell the groups a principal is in and without the groups we can't determine if he is an Administrator. In this case we will have two incorrect entries for the principal: the groups will be empty and the isAdministrator value will be false, because without fetching the groups we have no way of knowing either one.

For the second issue it could be done even inside the `getUserManager().getPrincipal()` method if the fetchReferences boolean is false.

PR #4385, opened 2020-10-15
NXP-29768: Add isEnabled on descriptor

kevinleturc:


bdelbosc (dismissed):
Great work

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/Descriptor.java
```diff
@@ -69,4 +76,18 @@ default boolean doesRemove() {
         return false;
     }
 
+    /**
+     * Returns a list of descriptor representing {@code d2} merged into {@code d1}.
```
ataillefer, line 80:
```suggestion
     * Returns a list of descriptors representing {@code d2} merged into {@code d1}.
```

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/DescriptorRegistry.java
```diff
@@ -50,10 +51,17 @@
...
-        return (T) merge(descriptors.getOrDefault(target, Collections.emptyMap())
-                                    .getOrDefault(xp, Collections.emptyMap())
-                                    .getOrDefault(id, Collections.emptyList()));
+        return this.<T> optDescriptor(target, xp, id).orElse(null);
+    }
 
+    /**
+     * @since 11.4
+     */
+    public <T extends Descriptor> Optional<T> optDescriptor(String target, String xp, String id) {
```
ataillefer, line 60: public or protected?

modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/FileImporterDescriptor.java
```diff
@@ -173,23 +176,14 @@ public Descriptor merge(Descriptor o) {
...
         return merged;
     }
-
-    protected <T> T defaultValue(T value, T defaultValue) {
-        return value == null ? defaultValue : value;
-    }
-
-    @Override
-    public boolean doesRemove() {
-        return !enabled;
```
ataillefer, line 195: Why can it be safely removed?

ataillefer (approved):

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/DescriptorRegistry.java
kevinleturc: I would keep the public modifier for later use if it's ok for the team?

modules/platform/nuxeo-platform-filemanager-core/src/main/java/org/nuxeo/ecm/platform/filemanager/service/extension/FileImporterDescriptor.java
kevinleturc: doesRemove makes the last descriptor null which is filtered out by the registry. Now that the registry takes care to filter out disabled contributions, this override is not needed anymore.

Furthermore, the doesRemove is better suited to re-init the contributions merge with such contributions:

```
<extension target="" point="">
  <something remove="true" />
  <something ...> <!-- completely new contribution -->
     ...
  </something>
</extension>
```

Whereas enabled with help of merge attribute allows us to do what we want in a single contribution.

troger (dismissed):

ataillefer (dismissed):

ataillefer (approved):

modules/platform/login/nuxeo-platform-login-openid/src/main/java/org/nuxeo/ecm/platform/oauth2/openid/OpenIDConnectProviderDescriptor.java
```diff
@@ -37,6 +42,8 @@
 public class OpenIDConnectProviderDescriptor implements Descriptor {
     protected static final long serialVersionUID = 1L;
 
+    protected static final Logger log = LogManager.getLogger(OpenIDConnectProviderDescriptor.class);
```
efge, line 45: `private`

modules/platform/login/nuxeo-platform-login-openid/src/main/java/org/nuxeo/ecm/platform/oauth2/openid/OpenIDConnectProviderDescriptor.java
```diff
@@ -161,6 +168,10 @@ public String getIcon() {
 
     @Override
     public boolean isEnabled() {
+        if (isBlank(clientId) || isBlank(clientSecret)) {
+            log.info("OpenId provider: {} is disabled because clientId and/or clientSecret are empty", name);
```
efge, line 172: Having a log in the getter means that this is going to potentially get logged numerous times, no?

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkActionDescriptor.java
```diff
@@ -38,7 +39,7 @@
 
     // @since 11.1
     @XNode("@enabled")
-    protected boolean isEnabled = true;
+    protected Boolean enabled;
```
efge, line 43: Aren't you changing the merge semantics? Previously if we had

    <extension target="" point="">
      <something enabled="false" />
      <something ... /> 
    </extension>

Then the new contribution would be enabled, but now it'll keep a merged enabled=false flag. Or is something else changing in the merge?

modules/platform/login/nuxeo-platform-login-openid/src/main/java/org/nuxeo/ecm/platform/oauth2/openid/OpenIDConnectProviderDescriptor.java
kevinleturc: It should be done once per Component#start calls. Otherwise, I can remove it, not sure that an info log is going to be useful at some point, wdyt?

efge: I don't think it's the descriptor's role to warn of such things. For me the descriptor is a pure carrier of data without any logic (and in general even the defaults should probably be elsewhere, in a service).

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkActionDescriptor.java
kevinleturc: You're right, this one will keep the {{enabled=false}}. 

This change was introduced in 11.x, I think we can break the compatibility there, wdyt?

efge: I'm ok, I just wanted to be sure you were aware of and ok with the change.

efge (dismissed):

PR #4386, opened 2020-10-15
NXP-29686: Run custom shell scripts at the end of ENTRYPOINT

ataillefer:
This allows for instance to invoke `nuxeoctl mp-install` in a registered way if the NUXEO_CLID environment varible is passed when running the container.


troger (approved):

kevinleturc (approved):

bdelbosc (approved):

bdelbosc (commented):
indeed I think it is better if the scripts are executed after the clid setting and before the mp-install packages, my use case is to use a cache for packages using a volume to prevent download on each start

PR #4398, opened 2020-10-19
NXP-29772: Fix list/grid view display when POST method is used by page providers

tmartins:
Manual PR after T&P passed and created https://github.com/nuxeo/nuxeo/pull/4397 but it contains the fix for 2 tickets, and the second one is not clean


loopingz (approved):
It seems weird to me to have both currentOffset and currentPageIndex but it is more on the underlying layer than this current PR

efge:
> It seems weird to me to have both currentOffset and currentPageIndex but it is more on the underlying layer than this current PR

Yes the PageProvider architecture is really old and messy.

PR #4411, opened 2020-10-22
NXP-29791: forbid deletion of the main blob of a document under retention / legal hold

RSalem07:
PR opened for now only on master to get Florent's feedbacks, about the implementation. 

Build on old infra: https://qa.nuxeo.org/jenkins/job/TestAndPush/job/ondemand-testandpush-saouana/251/

### DIFF
--- a/modules/core/nuxeo-core-storage/src/main/java/org/nuxeo/ecm/core/storage/BaseDocument.java
+++ b/modules/core/nuxeo-core-storage/src/main/java/org/nuxeo/ecm/core/storage/BaseDocument.java
@@ -674,13 +674,13 @@ public abstract class BaseDocument<T extends StateAccessor> implements Document 
 
     protected void setValueBlob(T state, Blob blob, String xpath) throws PropertyException {
         BlobInfo blobInfo = new BlobInfo();
+        DocumentBlobManager blobManager = Framework.getService(DocumentBlobManager.class);
+        try {
+            blobInfo.key = blobManager.writeBlob(blob, this, xpath);
+        } catch (IOException e) {
+            throw new PropertyException("Cannot get blob info for: " + blob, e);
+        }
         if (blob != null) {
-            DocumentBlobManager blobManager = Framework.getService(DocumentBlobManager.class);
-            try {
-                blobInfo.key = blobManager.writeBlob(blob, this, xpath);
-            } catch (IOException e) {
-                throw new PropertyException("Cannot get blob info for: " + blob, e);
-            }
             blobInfo.filename = blob.getFilename();
             blobInfo.mimeType = blob.getMimeType();
             blobInfo.encoding = blob.getEncoding();
@@ -691,17 +691,13 @@ public abstract class BaseDocument<T extends StateAccessor> implements Document 
     }
 
     protected void setPropertyBlobData(String xpath, String string) {
+        Blob blob = string == null ? null : Blobs.createBlob(string);
+        DocumentBlobManager blobManager = Framework.getService(DocumentBlobManager.class);
         String key;
-        if (string == null) {
-            key = null;
-        } else {
-            Blob blob = Blobs.createBlob(string);
-            DocumentBlobManager blobManager = Framework.getService(DocumentBlobManager.class);
-            try {
-                key = blobManager.writeBlob(blob, this, xpath);
-            } catch (IOException e) {
-                throw new PropertyException("Cannot write binary for doc: " + getUUID(), e);
-            }
+        try {
+            key = blobManager.writeBlob(blob, this, xpath);
+        } catch (IOException e) {
+            throw new PropertyException("Cannot write binary for doc: " + getUUID(), e);
         }
         setPropertyValue(xpath, key);
     }
--- a/modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManager.java
+++ b/modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManager.java
@@ -60,6 +60,9 @@ public interface DocumentBlobManager {
 
     /**
      * Writes a {@link Blob} to storage and returns its key.
+     * <p>
+     * The passed blob may be {@code null}, in which case a {@code null} key is returned after checking that deleting
+     * this blob is allowed.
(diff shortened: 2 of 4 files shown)


modules/core/nuxeo-core-storage/src/main/java/org/nuxeo/ecm/core/storage/BaseDocument.java
```diff
@@ -691,17 +691,13 @@ protected void setValueBlob(T state, Blob blob, String xpath) throws PropertyExc
...
-            Blob blob = Blobs.createBlob(string);
-            DocumentBlobManager blobManager = Framework.getService(DocumentBlobManager.class);
-            try {
-                key = blobManager.writeBlob(blob, this, xpath);
-            } catch (IOException e) {
-                throw new PropertyException("Cannot write binary for doc: " + getUUID(), e);
-            }
+        try {
+            key = blobManager.writeBlob(blob, this, xpath);
+        } catch (IOException e) {
```
nmpcunha, line 699: Not sure if I'm seeing this correctly but, would it make sense to just declare `key` and call `setPropertyValue` inside the _try_ block?

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -182,6 +182,14 @@ protected BlobProvider getBlobProvider(String key, String repositoryName) {
      */
     @Override
     public String writeBlob(Blob blob, Document doc, String xpath) throws IOException {
+        if (blob == null) {
+            if (MAIN_BLOB_XPATH.equals(xpath) && doc.isUnderRetentionOrLegalHold()) {
```
nmpcunha, line 186: IIRC it is only about `file:content`, right? Other blobs can be modified if a document is under retention?

nmpcunha (commented):
LGTM
Left just two small questions 🙂

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -182,6 +182,14 @@ protected BlobProvider getBlobProvider(String key, String repositoryName) {
...
+                throw new DocumentSecurityException(
+                        "Cannot delete blob from document " + doc.getUUID() + ", it is under retention / hold");
+            }
+            return null;
+        }
```
kevinleturc, line 191: What do you think about adding a new API such as `isBlobWriteable` which does this check, and maybe the one l. 211(?), in order to avoid to allow null values for _blob_ in this API?

nmpcunha: Not sure if I understood `and maybe the one l. 211(?)`, could you please elaborate a little bit?

Regarding the new API, would it also make sense for the 10.10 backport?
Thanks!

efge:
Apologies for being late to review, I'll definitely do it before the end of the weekend

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
kevinleturc: >Not sure if I understood and maybe the one l. 211(?), could you please elaborate a little bit?

Sure, I was talking about this part of the code https://github.com/nuxeo/nuxeo/blob/b65f2dd57707081d3f7451f90dc96ea01e8df7c5/modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java#L219-L222 which could leverage
the new API.

>Regarding the new API, would it also make sense for the 10.10 backport?

It would make sense for the 10.10 backport if the API has a default implementation.

modules/core/nuxeo-core-storage/src/main/java/org/nuxeo/ecm/core/storage/BaseDocument.java
```diff
@@ -691,17 +691,13 @@ protected void setValueBlob(T state, Blob blob, String xpath) throws PropertyExc
     }
 
     protected void setPropertyBlobData(String xpath, String string) {
+        Blob blob = StringUtils.isNotEmpty(string) ? Blobs.createBlob(string) : null;
```
efge: Testing for "empty" is wrong, it's legal too have a blob containing a 0-byte content. 
Even though for now this is used in cases where the difference doesn't really matter (binary fulltext), it's not the case in general.

modules/core/nuxeo-core-storage/src/main/java/org/nuxeo/ecm/core/storage/BaseDocument.java
efge: It's preferable to have try/catch blocks cover just what's needed and nothing more.

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
efge: It's not clear to me on what class you want to add this new API.
As it's a detail, let's postpone refactorings and just do a minimal change + backport.

efge (changes requested):

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManager.java
```diff
@@ -60,6 +60,9 @@
 
     /**
      * Writes a {@link Blob} to storage and returns its key.
+     * <p>
+     * The passed blob may be {@code null}, in which case a {@code null} key is returned after checking that deleting
+     * this blob is allowed.
```
efge, line 65: I added the Javadoc to.

nmpcunha (approved):

troger (approved):

PR #4060, opened 2020-05-23
NXP-29151: use HTTP code 400 instead of 500 for Automation request parsing errors

efge:

PR #4417, opened 2020-10-28
Fix NXP-29699 by returning 400-class error on bad queries

charlesboidot:
Fix for Jira Issue: https://jira.nuxeo.com/browse/NXP-29699

### DIFF
--- a/modules/core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/QueryParseException.java
+++ b/modules/core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/QueryParseException.java
@@ -21,6 +21,8 @@
 
 package org.nuxeo.ecm.core.query;
 
+import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
+
 import org.nuxeo.ecm.core.api.NuxeoException;
 
 /**
@@ -31,18 +33,19 @@ public class QueryParseException extends NuxeoException {
     private static final long serialVersionUID = 1L;
 
     public QueryParseException() {
+        super(SC_BAD_REQUEST);
     }
 
     public QueryParseException(String message) {
-        super(message);
+        super(message, SC_BAD_REQUEST);
     }
 
     public QueryParseException(String message, Throwable cause) {
-        super(message, cause);
+        super(message, cause, SC_BAD_REQUEST);
     }
 
     public QueryParseException(Throwable cause) {
-        super(cause);
+        super(cause, SC_BAD_REQUEST);
     }
 
 }
--- a/modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/jdbc/JDBCMapper.java
+++ b/modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/jdbc/JDBCMapper.java
@@ -19,6 +19,8 @@
  */
 package org.nuxeo.ecm.core.storage.sql.jdbc;
 
+import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
+
 import static org.nuxeo.ecm.core.api.ScrollResultImpl.emptyResult;
 
 import java.io.Serializable;
@@ -428,7 +430,7 @@ public class JDBCMapper extends JDBCRowMapper implements Mapper {
         try {
             return new ResultSetQueryResult(queryMaker, query, queryFilter, pathResolver, this, params);
         } catch (SQLException e) {
-            throw new NuxeoException("Invalid query: " + queryType + ": " + query, e);
+            throw new NuxeoException("Invalid query: " + queryType + ": " + query, e, SC_BAD_REQUEST);
         }
     }
 
@@ -557,7 +559,7 @@ public class JDBCMapper extends JDBCRowMapper implements Mapper {
                 return new PartialList<>(projections, totalSize);
             }
         } catch (SQLException e) {
-            throw new NuxeoException("Invalid query: " + query, e);
+            throw new NuxeoException("Invalid query: " + query, e, SC_BAD_REQUEST);
         }
     }
 


kevinleturc (commented):
Thanks for the contribution. What do you think about giving to all QueryParseException the same HTTP status code? IMHO, there's no other valid HTTP status code.

efge:
> Thanks for the contribution. What do you think about giving to all QueryParseException the same HTTP status code? IMHO, there's no other valid HTTP status code than the bad request.

Agreed, and it would avoid us changing `Scanner.java` and `parser.java` which are actually files generated by `nxql.flex` and `nxql.cup` for which the regeneration mechanism is manual (and poorly documented).

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/jdbc/NXQLQueryMaker.java
```diff
@@ -1101,7 +1101,7 @@ public void visitFromClause(FromClause node) {
                 }
                 Set<String> subTypes = model.getDocumentSubTypes(typeName);
                 if (subTypes == null) {
-                    throw new QueryParseException("Unknown type: " + typeName);
+                    throw new QueryParseException("Unknown type: " + typeName,SC_BAD_REQUEST);
```
efge, line 1104: Space after `,`

efge (changes requested):

modules/core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/QueryParseException.java
```diff
@@ -22,6 +22,7 @@
 package org.nuxeo.ecm.core.query;
 
 import org.nuxeo.ecm.core.api.NuxeoException;
+import static org.apache.http.HttpStatus.SC_BAD_REQUEST;
```
efge, line 25: Formatter should put static imports before regular ones.

modules/core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/QueryParseException.java
```diff
@@ -34,7 +35,7 @@ public QueryParseException() {
     }
 
     public QueryParseException(String message) {
-        super(message);
+        super(message, SC_BAD_REQUEST);
```
efge, line 38: Ok but all other `super` calls in other constructors should do the same

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/jdbc/JDBCMapper.java
```diff
@@ -428,7 +430,7 @@ public IterableQueryResult queryAndFetch(String query, String queryType, QueryFi
         try {
             return new ResultSetQueryResult(queryMaker, query, queryFilter, pathResolver, this, params);
         } catch (SQLException e) {
-            throw new NuxeoException("Invalid query: " + queryType + ": " + query, e);
+            throw new NuxeoException("Invalid query: " + queryType + ": " + query, e, SC_BAD_REQUEST);
```
efge, line 433: All changes in this class not needed after the constructor changes are done.

efge (changes requested):

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/jdbc/JDBCMapper.java
efge: Hm sorry forget about this, I was wrong.

charlesboidot: It's a NuxeoException not a QueryParseException, so by default it throws a 500 status code if not precised otherwise.

modules/core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/QueryParseException.java
```diff
@@ -34,15 +35,15 @@ public QueryParseException() {
     }
```
kevinleturc, line 35: This one should call `super(SC_BAD_REQUEST);`

modules/core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/QueryParseException.java
```diff
@@ -21,6 +21,8 @@
 
 package org.nuxeo.ecm.core.query;
 
+import static org.apache.http.HttpStatus.SC_BAD_REQUEST;
```
troger, line 24: What about using `javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST` like what we did in other Exception such as `DocumentNotFoundException` or `ConcurrentUpdateException`?

Same for other status code.

charlesboidot: I used the same SC_BAD_REQUEST as the PR [here](https://github.com/nuxeo/nuxeo/pull/4060/files)

troger: Oh well, ok ...
@efge shouldn't we want to align on `javax.servlet.http.HttpServletResponse` instead of relying on `org.apache.httpcomponents:httpcore` for those constants?

efge: @troger that would be a possible cleanup yes but I'd say the benefit is very low. We don't plan on removing HttpComponents from our dependencies anytime soon anyway.

efge: On the other hand for new code it's worth using standard JDK classes so `javax.servlet.http.HttpServletResponse` is slightly better

efge (dismissed):

modules/core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/QueryParseException.java
efge: @charlesboidot if you're not tired of doing iterations on this PR you could change the import yes, otherwise forget it

charlesboidot: It's ok I'll change it

efge (approved):

troger (approved):
👍 💪

PR #4420, opened 2020-10-28
NXP-29738: Fix S3 direct upload same file import

kevinleturc:


efge:
All the blob move stuff is needed only because we attempt to precompute the high-level digest, maybe together with https://jira.nuxeo.com/browse/NXP-29807 we don't need to keep all this logic. What do you think @nelsonsilva?

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
```diff
@@ -339,6 +334,19 @@ public boolean completeUpload(String batchId, String fileIndex, BatchFileInfo fi
         return true;
     }
 
+    protected String doMove(CopyObjectRequest request) {
```
efge, line 337: Could you add a Javadoc explaining that it returns the ETag of the resulting blob?

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
```diff
@@ -339,6 +334,19 @@ public boolean completeUpload(String batchId, String fileIndex, BatchFileInfo fi
...
+        Copy rename = getTransferManager().copy(request);
+        try {
+            var copyResult = rename.waitForCopyResult();
+            return copyResult.getETag();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new NuxeoException(e);
+        } finally {
+            amazonS3.deleteObject(bucket, request.getSourceKey());
```
efge, line 346: Should we also catch `AmazonS3Exception` for this delete and just log a debug, like in the case where `doesObjectExist` returns true?

efge (changes requested):

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
kevinleturc: I think so, object deletion shouldn't fail the upload if failing.

kevinleturc: I will also try/catch/addSuppressed at this location https://github.com/nuxeo/nuxeo/pull/4420/files#diff-6840572d6cd2147c22a5ea5800d2525059bb39c001a5909d15f0370f9ee506d8R330 to not shallow the original `NuxeoException`.

efge (approved):

bdelbosc (approved):

PR #4438, opened 2020-11-04
NXP-29825: keep only one install-packages.sh script

troger:


kevinleturc (dismissed):

docker/install-packages.sh
```diff
@@ -39,11 +43,19 @@ fi
 
 if [ -n "$connect_url" ]; then
   echo "Setting Connect URL: $connect_url"
-  printf "org.nuxeo.connect.url=%b\n" "$connect_url" > $NUXEO_HOME/bin/nuxeo.conf
+  printf "org.nuxeo.connect.url=%b\n" "$connect_url" >> $NUXEO_HOME/bin/nuxeo.conf
```
atchertchian, line 46: Would it make sense to separate this change (">" into ">>"), which is not directly related to the merge of both scripts, and fixes the bug that empties the nuxeo.conf file when using the connect URL parameter?

troger: Could be a separate commit yes.

ataillefer (approved):

kevinleturc (approved):

PR #4444, opened 2020-11-06
NXP-29587: Scroller produces records downstream depending on a thres…

bdelbosc:
…hold

If the bulk scroller materialize more than 1m ids, records are sent
downstream to prevent a possible OOM.

### DIFF
--- a/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkAdminServiceImpl.java
+++ b/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkAdminServiceImpl.java
@@ -47,6 +47,13 @@ public class BulkAdminServiceImpl implements BulkAdminService {
 
     public static final String BULK_SCROLL_PRODUCE_IMMEDIATE_PROPERTY = "nuxeo.core.bulk.scroller.produceImmediate";
 
+    // @since 11.4
+    public static final String BULK_SCROLL_PRODUCE_IMMEDIATE_THRESHOLD_PROPERTY = "nuxeo.core.bulk.scroller.produceImmediateThreshold";
+
+    // by default switch to produce immediate when there are more than 1m ids
+    // @since 11.4
+    public static final int DEFAULT_PRODUCE_IMMEDIATE_THRESHOLD_PROPERTY = 1_000_000;
+
     public static final int DEFAULT_SCROLL_SIZE = 100;
 
     public static final int DEFAULT_SCROLL_KEEP_ALIVE = 60;
--- a/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
+++ b/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
@@ -87,6 +87,9 @@ public class BulkScrollerComputation extends AbstractComputation {
 
     protected final boolean produceImmediate;
 
+    // @since 11.4
+    protected final long produceImmediateThreshold;
+
     protected final int transactionTimeoutSeconds;
 
     protected int scrollSize;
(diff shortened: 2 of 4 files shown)


modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkAdminServiceImpl.java
```diff
@@ -47,6 +47,13 @@
 
     public static final String BULK_SCROLL_PRODUCE_IMMEDIATE_PROPERTY = "nuxeo.core.bulk.scroller.produceImmediate";
 
+    // @since 11.4
+    public static final String BULK_SCROLL_PRODUCE_IMMEDIATE_THRESHOLD_PROPERTY = "nuxeo.core.bulk.scroller.produceImmediateThreshold";
+
+    // by default switch to produce immediate when there are more than 1m ids
+    // @since 11.4
+    public static final int DEFAULT_PRODUCE_IMMEDIATE_THRESHOLD_PROPERTY = 1_000_000;
```
efge, line 55: This constant isn't used...

efge (changes requested):

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkAdminServiceImpl.java
bdelbosc: yes thanks, the pr has been reworked a bit to fix this and to flush accumulated records downstream when reaching the threshold avoiding a reordering of ids.

kevinleturc (approved):

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
```diff
@@ -111,10 +114,18 @@ public BulkScrollerComputation(String name, int nbOutputStreams, int scrollBatch
     // @since 11.2
     public BulkScrollerComputation(String name, int nbOutputStreams, int scrollBatchSize, int scrollKeepAliveSeconds,
             Duration transactionTimeout, boolean produceImmediate) {
+        this(name, nbOutputStreams, scrollBatchSize, scrollKeepAliveSeconds, DEFAULT_SCROLL_TRANSACTION_TIMEOUT,
+                produceImmediate, 0);
+    }
+
+    // @since 11.4
+    public BulkScrollerComputation(String name, int nbOutputStreams, int scrollBatchSize, int scrollKeepAliveSeconds,
+            Duration transactionTimeout, boolean produceImmediate, int produceImmediateThreshold) {
```
efge, line 123: Isn't it worth considering a builder pattern, given the number of parameters that we now pass at construction?

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
```diff
@@ -272,17 +283,30 @@ protected void updateStatusAfterScroll(ComputationContext context, String comman
...
         }
         ids.clear(); // this clear the documentIds part that has been sent
     }
 
+    protected void flushRecords(ComputationContextImpl contextImpl, String commandId) {
+        log.warn("Scroller records threshold reached ({}) for action: {} on command: {}, flushing records downstream",
+                produceImmediateThreshold, actionStream, commandId);
+        contextImpl.getRecords(actionStream)
+                   .forEach(record -> contextImpl.produceRecordImmediate(actionStream, record));
+        contextImpl.getRecords(actionStream).clear();
```
efge, line 309: Can we put `contextImpl.getRecords(actionStream)` in a local variable or is the semantic wrong for that?

efge (approved):

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
bdelbosc: maybe, but this constructor is only used in one place and is not intended to be used elsewhere

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
bdelbosc: we could will it be more readable not sure

efge: It is more readable, you're just not used to it 😀 With a local variable it's immediately clear that we're acting twice on the same object instead of re-doing a call to getRecords that may return different things... Which was a question I asked myself when I read this!

efge: Here, for the reasons I mentioned, the code would have been better with a local variable. Please don't ignore reviewer's comments and requests for changes.

PR #4479, opened 2020-11-19
NXP-29585: Add Bulk command queryLimit

bdelbosc:
A default limit can be set on action and overriden when building command,
a flag in the status indicate if the limit has been reached for a command.

### DIFF
--- a/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
+++ b/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
@@ -20,7 +20,6 @@
 package org.nuxeo.ecm.core.bulk.computation;
 
 import static java.lang.Math.min;
-import static org.nuxeo.ecm.core.bulk.BulkAdminServiceImpl.DEFAULT_SCROLL_TRANSACTION_TIMEOUT;
 import static org.nuxeo.ecm.core.bulk.BulkServiceImpl.STATUS_STREAM;
 import static org.nuxeo.ecm.core.bulk.message.BulkStatus.State.ABORTED;
 import static org.nuxeo.ecm.core.bulk.message.BulkStatus.State.COMPLETED;
@@ -98,6 +97,20 @@ public class BulkScrollerComputation extends AbstractComputation {
 
     protected String actionStream;
 
+    public static Builder builder(String name, int nbOutputStreams) {
+        return new Builder(name, nbOutputStreams);
+    }
+
+    protected BulkScrollerComputation(Builder builder) {
+        super(builder.name, 1, builder.nbOutputStreams);
+        this.scrollBatchSize = builder.scrollBatchSize;
+        this.scrollKeepAliveSeconds = builder.scrollKeepAliveSeconds;
+        this.produceImmediate = builder.produceImmediate;
+        this.produceImmediateThreshold = builder.produceImmediateThreshold;
+        this.transactionTimeoutSeconds = Math.toIntExact(builder.transactionTimeout.toSeconds());
+        documentIds = new ArrayList<>(scrollBatchSize);
+    }
+
     /**
      * @param name the computation name
      * @param nbOutputStreams the number of registered bulk action streams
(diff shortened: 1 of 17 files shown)


kevinleturc (dismissed):
I think your commit _NXP-29585: Bump default partitions for async processing_ doesn't have the right NXP, shouldn't it be _NXP-29677_?

kevinleturc (approved):

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
```diff
@@ -111,6 +112,9 @@
     @Nullable
     protected Long processingDurationMillis;
 
+    @AvroDefault("false")
+    protected boolean limitedQuery;
```
efge, line 116: It's not clear to me what his means when it's true. Is it that there was a limit passed when doing the query? Or that the query is incomplete in its results returned.
A useful flag would be more about "limitReached", which I would call "resultsTruncated" or "resultsIncomplete", or "moreResults".

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
```diff
@@ -169,21 +176,33 @@ protected void processRecord(ComputationContext context, Record record) {
...
                     List<String> docIds = scroll.next();
-                    documentIds.addAll(docIds);
+                    int scrollCount = docIds.size();
+                    if (documentCount + scrollCount < queryLimit) {
+                        documentIds.addAll(docIds);
+                    } else {
+                        scrollCount = Math.toIntExact(queryLimit - documentCount);
+                        documentIds.addAll(docIds.subList(0, scrollCount));
+                        limitReached = true;
+                    }
```
efge, line 186: What about if we get exactly to the queryLimit value? Then we flag limitReached=true but there's no additional documents. The use of this for paging is not clear.

modules/core/nuxeo-core-bulk/src/test/java/org/nuxeo/ecm/core/bulk/action/TestWordCountAction.java
```diff
@@ -70,9 +71,57 @@ public void testWordCount() throws Exception {
...
+                .queryUnlimited()
+                .build();
+        commandId = bulkService.submit(command);
+        assertTrue(bulkService.await(commandId, Duration.ofSeconds(20)));
+        status = bulkService.getStatus(commandId);
+        assertEquals(COMPLETED, status.getState());
+        assertFalse(status.isLimitedQuery());
+        long resultUnlimited = (Integer) status.getResult().get("wordCount");
+        assertEquals(wordCount, resultUnlimited);
+    }
```
efge, line 123: Please add tests for the behavior where the `.queryLimit()` is wordCount-1, exactly wordCount, and wordCount+1 to test edge cases.

efge (changes requested):

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
bdelbosc: limitedQuery is true if the query has a limit and it has been reached which makes sense in the context of the status, the query itself and its attributes (like the limit) are part of the command. limitReached or queryLimitReached could be an alternative.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
bdelbosc: the limit should be taken as LIMIT in SQL, the bulk service is not done for paging and we don't have to go into an off-by-one problem here.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
bdelbosc: will use queryLimitReached

efge (approved):

kevinleturc (approved):

PR #4480, opened 2020-11-19
NXP-29833: check ACL name is not null on creation

charlesboidot:


modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/security/impl/ACLImpl.java
```diff
@@ -56,6 +57,9 @@
     private final boolean isReadOnly;
 
     public ACLImpl(String name, boolean isReadOnly) {
+        if(name==null) {
```
efge, line 60: There should be a space before `(` and around `==`.
But instead of these 3 lines please use `Objects.requireNonNull(name, "ACL name can't be null")`. It won't throw a `NuxeoException` but that's not a problem for this kind of check. Unit test will need to be updated of course.

modules/core/nuxeo-core-api/src/test/java/org/nuxeo/ecm/core/api/security/TestACL.java
```diff
@@ -46,6 +48,16 @@ public void testGetName() {
         assertEquals("test acl", acl.getName());
     }
 
+    @Test
+    public void testExceptionOnNullName() {
+        try {
+            acl = new ACLImpl(null, false);
+            fail("ACL is not supposed to be created.");
+        } catch (NuxeoException e) {
+            // Test ok
```
efge, line 57: Please assert on the message of the exception too:

    assertEquals("ACL name can't be null", e.getMessage());

efge (changes requested):

efge (approved):

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/security/impl/ACLImpl.java
```diff
@@ -56,6 +56,7 @@
     private final boolean isReadOnly;
 
     public ACLImpl(String name, boolean isReadOnly) {
+        Objects.requireNonNull(name, "ACL name can't be null");
         this.name = name;
```
kevinleturc, line 60:
```suggestion
        this.name = Objects.requireNonNull(name, "ACL name can't be null");
```

kevinleturc (approved):

efge (approved):

kevinleturc:
_license/cla_ didn't report its status -> as everything else is ok, merge with rights.

PR #4490, opened 2020-11-23
NXP-29580 runtime registries

atchertchian:
First Draft PR for first discussions around strategy and API for https://jira.nuxeo.com/browse/NXP-29580 (Rework descriptor mechanism)

Feedback welcome on class and method names (before i start migrating more xmap objects to the new system).


atchertchian:
Did a few more separated commits after discussions, to be reviewed too

atchertchian:
Closing as is, will create at least 2 new PRs instead: one with needed features, others with migrations to the new registry system.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ExtensionPointImpl.java
```diff
@@ -56,6 +66,15 @@
     @XParent
     public RegistrationInfo ri;
 
+    // potential registry class declaration
+    @XNode(value = "registry@class")
+    protected String registryKlass;
+
+    // final operating registry class
+    protected Registry registry;
```
kevinleturc, line 74: Could we have it outside of this class, like in `ComponentManager`? The reason is that this object should be a POJO imho.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ExtensionPointImpl.java
```diff
@@ -76,39 +95,114 @@ public String getSuperComponent() {
...
+    @Override
+    public Registry getRegistry() {
+        if (registry == null) {
+            registry = computeFinalRegistry();
+            if (registry == null) {
+                registry = NULL_REGISTRY;
+            }
+        }
+        if (registry.isNull()) {
+            return null;
```
kevinleturc, line 165: Why returning null instead of the `NULL_REGISTRY`?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/services/config/ConfigurationServiceImpl.java
```diff
@@ -59,32 +59,8 @@
...
-                if (d == null) {
-                    List<ConfigurationPropertyDescriptor> descs = getDescriptors(CONFIGURATION_EP);
-                    descriptors = d = descs.stream().collect(Collectors.toMap(desc -> desc.getId(), desc -> desc));
-                }
-            }
-        }
-        return d;
+    protected MapRegistry getMapRegistry() {
+        return getRegistry(CONFIGURATION_EP, MapRegistry.class);
     }
```
kevinleturc, line 64: I think we should have the same method on `MapRegistry` and `SingleRegistry`, depending on the implementation an exception may be thrown. Then this is the service's responsibility to know which methods it has to call, and this will avoid such methods.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ExtensionPointImpl.java
atchertchian: NULL_REGISTRY is only used for internal caching logics (to avoid detecting over and over that there is no registry defined), i can expose it and return it if better but i'm not convinced it's clearer (?)

kevinleturc: I think it is, the other side is getting NPE in border cases, it might help no?

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/AbstractRegistry.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    public boolean isFlagged(String id) {
+        return flags.contains(id);
+    }
+
+    @Override
+    public void register(Context ctx, XAnnotatedObject xObject, Element element, String flag) {
+        flags.add(flag);
+        registrations.add(new RegistryContribution(ctx, xObject, element, flag));
```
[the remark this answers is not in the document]
atchertchian: The idea here is to call the other register method (that will process DOM and apply merge logics and such) only when the registry getters are called (at initialization).
It should not hurt to process these right away because they should be retrieved at some point anyway, but in practice it is useful for hot reload logics (so that computations are not done on the fly, some of them being potentially not needed depending on the deploy/undeploy subsequent calls)

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ExtensionPointImpl.java
atchertchian: Ok i'll try to move it when reviewing all commits

atchertchian: Note that this means moving the registration logics out of this class (and similarly, the registry retrieval will trigger a lookup to the component manager).
Note that for compat logics, i also think i should stop loading contribution objects on the Extension objects (that should be a POJO too if we're consistent) when a registry is defined: nobody should expect the extension to hold contributions anymore when using registries, and this will spare unneeded xmap load of these objects, wdyt?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ExtensionPointImpl.java
atchertchian: Let's see when moving this API to the ComponentManager then

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/services/config/ConfigurationServiceImpl.java
atchertchian: I started by putting both Map and Single API on the interface indeed, but i'm not sure it's clearer, and since people can define their own registry class, maybe it's better to keep this out of the interface (?). We can rediscuss this in all cases, i'll try to re-arrange the code with other feedback first.

atchertchian:
see #4537 for first follow up PR

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ExtensionPointImpl.java
atchertchian: I re-arranged the logics while moving to ComponentManager, see #4537

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ExtensionPointImpl.java
atchertchian: done with #4537

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/services/config/ConfigurationServiceImpl.java
atchertchian: I think API should be simpler now, but we can discuss this again in the upcoming migration PR (will reference it here when done)

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/AbstractRegistry.java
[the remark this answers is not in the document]
atchertchian: I kept the logic as is in #4537, please tell if you still think there should be changes in this case

atchertchian:
see #4539 for descriptors migration draft PR

PR #4498, opened 2020-11-30
NXP-29908: allow enabling/disabling components conditionally

atchertchian:
First draft before starting to migrate nxftl templates to these logics

### DIFF
--- a/modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
+++ b/modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
@@ -72,8 +72,12 @@ public class RegistrationInfoImpl implements RegistrationInfo {
     @XNode("@name")
     ComponentName name;
 
-    @XNode("@disabled")
-    boolean disabled;
+    @XNode("@disable")
+    protected boolean disable;
+
+    // @since 11.5
+    @XNode("@enable")
+    protected boolean enable = true;
 
     @XNode("configuration")
     ConfigurationDescriptor config;
@@ -160,7 +164,12 @@ public class RegistrationInfoImpl implements RegistrationInfo {
 
     @Override
     public boolean isDisabled() {
-        return disabled;
+        return disable;
+    }
+
+    @Override
+    public boolean isEnabled() {
+        return enable;
     }
 
     @Override
--- a/modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/TestComponentEnablement.java
+++ b/modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/TestComponentEnablement.java
@@ -1,0 +1,138 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Anahide Tchertchian
+ */
+package org.nuxeo.runtime;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+import javax.inject.Inject;
+
+import org.junit.Test;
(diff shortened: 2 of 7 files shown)


modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
```diff
@@ -72,8 +72,12 @@
     @XNode("@name")
     ComponentName name;
 
-    @XNode("@disabled")
-    boolean disabled;
+    @XNode("@disable")
+    protected boolean disable = false;
```
efge, line 76: Remove ` = false`

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
```diff
@@ -72,8 +72,12 @@
     @XNode("@name")
     ComponentName name;
 
-    @XNode("@disabled")
-    boolean disabled;
+    @XNode("@disable")
```
efge, line 75: We're sure we don't need compat with the old name?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
```diff
@@ -72,8 +72,12 @@
...
+    protected boolean disable = false;
+
+    // @since 11.5
+    @XNode("@enable")
+    protected boolean enable = true;
```
efge, line 80: You could have only one field, for instance `enable` like elsewhere, and a setter for `@disable` that sets `enable = !disable`. The API could just have `isEnabled` and no redundant logic needed like `ri.isDisabled() || !ri.isEnabled()`

efge (changes requested):

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
atchertchian: Disablement API was only used in unused persisted registrations so I think we do not need compat indeed.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
atchertchian: Indeed but that would imply sorting properties (xmap supports it) to make sure disable is set after enable, and i was trying not to put logics inside the descriptor, as we're trying to stick to that "Data only object" rule.
We can still discuss it if it's better.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/RegistrationInfoImpl.java
kevinleturc: Why not removing it and leverage `enable`?

atchertchian: The idea is to be able to avoid scripting logics, so it's easier to have:
```
<component enable="${org.nuxeo.my.condition}">
<component>
```
for components to *enable* conditionally if org.nuxeo.my.condition runtime property is true, and also:
```
<component disable="${org.nuxeo.my.condition}">
<component>
```
for components to *disable* conditionally (on the same condition).

The alternative would be to introduce some scripting logics like

```
<component enable="!${org.nuxeo.my.condition}">
</component>
```

kevinleturc: Ok, it's clearer, thanks 👍.

I think the scripting logic is the way to go, otherwise, I'm pretty sure we're gonna end with:
```xml
<component enable="${org.nuxeo.my.condition.1}" disable="${org.nuxeo.my.condition.2}">
</component>
```
I'm afraid about maintenance on this.

Furthermore, improving the scripting logic could be interesting as we may need to have OR and AND boolean operators in the future. 
This may be needed by drive change finder which has 3 implementations and only one might be enabled.

atchertchian: Well if we need scripting logics, i'm afraid using nxftl is a better option: the goal of this feature was to avoid having scripting logics introduced for simple conditional enablements and disablements, and maybe simplify some configuration templates using (or over-using) freemarker templating for components configuration.
This does not prevent from keeping nxftl, and i do not see the point of this PR anymore if we're not happy with this improvement: i've discussed this several times with you both and you seemed in favor of it so i'm confused (but it's ok if you changed your minds, this was a quick try at simplifying some use cases).

atchertchian:
Status: need to re-discuss the goal and how to achieve it

PR #4510, opened 2020-12-03
NXP-29924: avoid storing intermediate chain results

nelsonsilva:
This should make async chain execution a bit more efficient as we just need the final chain output, no need to keep putting op outputs in the transient store.

Original bug was due to us detaching the document just so we could put it in the transient store but we never attached it again until the end of the chain.

### DIFF
--- a/modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
+++ b/modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
@@ -148,6 +148,8 @@ public class AsyncOperationAdapter extends DefaultAdapter {
 
         opCtx.setCallback(new OperationCallback() {
 
+            private Object output;
+
             @Override
             public void onChainEnter(OperationType chain) {
                 //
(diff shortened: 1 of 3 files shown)


troger (approved):

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
```diff
@@ -148,13 +148,16 @@ public Object doPost(ExecutionRequest xreq) {
...
+            private Object output;
+
             @Override
             public void onChainEnter(OperationType chain) {
                 //
             }
 
             @Override
             public void onChainExit() {
+                setOutput(executionId, (Serializable) output);
```
efge, line 160: Is it worth doing `output = null` after this to avoid keeping a reference to the output in a field?

efge (approved):

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
nelsonsilva: The operation callback is referenced by the context which is referenced by the thread that handles the chain execution so I'd expect it to be ready for GC once execution is done, no?

efge: Yes if it doesn't last longer than that then no need for manual "cleanup".

PR #4520, opened 2020-12-08
NXP-29837: handle a password on easyshare

s-guillaume:
Features :

Secure Easyshare shared folder with a password from Web-UI available on creation or edit layouts

On a secured Easyshare shared folder, when a « public » end-user is accessing a secured Easyshare shared folder, it will be prompt the accessCode

If the « public » end-user entered the defined accessCode, he will have access to the shared folder content from the current browser.
=> Some extra informations will be added the your browser URL
=> These extra informations are mandatory once connected and cannot be removed or shared to other « public » end-users (these informations are dynamically changed on each call with a matching based on browser JSESSIONID so using another browser will imply to re-enter accessCode or share it, notice it’s not working with Incognito session on Browsers)

AccessCode is controled on each call that means if you change your accessCode in the EasyShare folder, previously connected users will be disconnected and prompted a new AccessCode

Core/WebUI update
Two new metadatas as accessCodeRequired (Boolean) and accessCode (String) value added to existing easyshare schema (updated XSD)

WebUI Layouts (creation /edit) and JSF layout (creation/edit)

Checkbox for accessCodeRequired metadata to activate password requirement on public EasyShare access
Input text for accessCode to provide a password
Translation available in English/US, French, German and Spanish to check

Easyshare "public" access is not changing with no password
If EasyShare folder has no password set, access to public Easyshare remains available (no access code required so public access will remain as current version of EasyShare).

Capacity of checking checkbox "Access code required ?" and setting password to secure access to an Easyshare collection
If checkbox is checked, password is mandatory

Easyshare "public" access will ask for a password (access code) if the collection has password protected access (checkbox checked).
In that case, the user is prompted to enter the password.

If the right password is entered, the user will be able to navigate in the collection. Otherwise, he must re-enter the password.
Notice all links are available for end-users include 3 new dynamic url parameters (a, s and j). They are calculated on the fly, encrypted with default and generated salt and only available on the current browser.


modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShare.java
```diff
@@ -91,59 +98,159 @@ public Object run(CoreSession session, IdRef docRef) throws NuxeoException {
...
+                            return Response.serverError().status(Response.Status.NOT_FOUND).build();
+                        }
+
+                        if (!checkIfAccessCodeIsValid(docShare,accessCode)) {
+                            return Response.serverError().status(Response.Status.FORBIDDEN).build();
+                        }
+                        
+                        if (StringUtils.isBlank(jSessionId) || !jSessionId.equals(getRequestJSESSIONID())) {
+                            return Response.serverError().status(Response.Status.FORBIDDEN).build();
+                        }
```
kevinleturc, line 220: Doesn't it a backward compatibility change? What happens if the query parameter for jSessionId is not present?

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShare.java
```diff
@@ -185,18 +292,118 @@ private boolean checkIfShareIsValid(DocumentModel docShare) {
...
+        return buildUnrestrictedRunner(decryptedJSessionId, decryptedAccessCode, folderId, pageIndex).runUnrestricted(shareId);
+    }
+
+    @Path("{shareId}/{folderId}")
+    @POST
+    public Object postFolderListing(@PathParam("shareId") String shareId,
+            @PathParam("folderId") final String folderId,
+            @DefaultValue(DEFAULT_PAGE_INDEX) @FormParam("p") Long pageIndex,
+            @FormParam("accessCode") final String accessCode) {
+        return buildUnrestrictedRunner(getRequestJSESSIONID(), accessCode, folderId, pageIndex).runUnrestricted(shareId);
```
kevinleturc, line 346: Why not getting the JSessionId from the parameters like it is done on the GET endpoint?

Or, why getting the JSessionId from the parameters in the GET endpoint and not from the cookie?

kevinleturc: In fact, I don't understand why we're having POST endpoint for get folder listing, get blob stream, etc...

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShare.java
```diff
@@ -321,4 +487,4 @@ protected String getIpAddr() {
         }
         return ip;
     }
-}
+}
```
kevinleturc, line 490: Could you revert this change?

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
+package org.nuxeo.easyshare;
```
kevinleturc, line 1: Missing Copyright header.

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.SecretKeyFactory;
+import javax.crypto.spec.PBEKeySpec;
+import javax.crypto.spec.PBEParameterSpec;
+
+import org.apache.commons.codec.binary.Base64;
+
+public class EasyShareEncryptionUtil {
```
kevinleturc, line 22: Could you add:
```suggestion
/**
 * @since 11.5
 */
public class EasyShareEncryptionUtil {
```

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+
+	private EasyShareEncryptionUtil() {
+	}
+
+	/**
+	 * Return a predefined SALT
+	 * 
+	 * @return Predefined SAL (@see SALT)
+	 */
+	public static String getDefaultSaltAsString() {
```
kevinleturc, line 38:
```suggestion
	public static String getDefaultSalt() {
```

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+
+		return Base64.encodeBase64String(SALT);
+	}
+
+	/**
+	 * Generate a custom salt as base64 encoded string
+	 * 
+	 * @return random base64 string
+	 */
+	public static String getNextSaltAsString() {
```
kevinleturc, line 48:
```suggestion
	public static String getNextSalt() {
```

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+	 * Encode an input string with passed salt string (default value will be
+	 * used for salt if param is empty/null) (wrapper for byte[] salt)
+	 * 
+	 * @param input
+	 *            String to encode
+	 * @param salt
+	 *            SALT to use
+	 * @return An encoded string
+	 */
+	public static String encodeWithSaltString(String input, String salt) {
```
kevinleturc, line 63:
```suggestion
	public static String encode(String input, String salt) {
```

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	/**
+	 * Encode an input string with passed salt string (default value will be
+	 * used for salt if param is empty/null) (wrapper for byte[] salt)
+	 * 
+	 * @param input
+	 *            String to encode
+	 * @param salt
+	 *            SALT to use
```
kevinleturc, line 60: Could you apply our formatter?
Our lines are 120 characters long and there's no carriage return for the javadoc parameters.

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+	 * Decode an encoded token with passed salt string (default value will be
+	 * used for salt if param is empty/null)
+	 * 
+	 * @param token
+	 *            Encoded String to decode using passed salt
+	 * @param salt
+	 *            SALT to use when decoding
+	 * @return Decoded string from token
+	 */
+	public static String decodeWithSaltString(String token, String salt) {
```
kevinleturc, line 81:
```suggestion
	public static String decode(String token, String salt) {
```

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+		try {
+			return decodeWithSaltByte(token, Base64.decodeBase64(salt));
+		} catch (IllegalArgumentException e) {
+			return null;
+		}
```
kevinleturc, line 86: When does it happen? Shouldn't we re-throw a NuxeoException with a Not found/Denied code?

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+	 * Encode an input string with passed salt byte[] (default value will be
+	 * used for salt if param is empty/null)
+	 * 
+	 * @param input
+	 *            String to encode
+	 * @param salt
+	 *            SALT to use
+	 * @return An encoded string
+	 */
+	private static String encodeWithSaltByte(String input, byte[] salt) {
```
kevinleturc, line 110:
```suggestion
	private static String encode(String input, byte[] salt) {
```

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
kevinleturc: It seems that the whole file contains tabs, could you replace them by spaces?

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+			SecretKey key = SecretKeyFactory.getInstance("PBEWithMD5AndDES").generateSecret(keySpec);
+
+			Cipher ecipher = Cipher.getInstance(key.getAlgorithm());
+			ecipher.init(Cipher.ENCRYPT_MODE, key, paramSpec);
+
+			byte[] enc = ecipher.doFinal(input.getBytes());
+
+			String res = new String(Base64.encodeBase64(enc));
+			// escapes for url
+			res = res.replace('+', '-').replace('/', '_').replace("%", "%25").replace("\n", "%0A");
```
kevinleturc, line 129: Can't we use `URLEncoder` or something from a library? Apache libs we depend on may have such things.

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+
+			return res;
+
+		} catch (InvalidKeySpecException | NoSuchAlgorithmException | NoSuchPaddingException | InvalidKeyException
+				| InvalidAlgorithmParameterException | IllegalBlockSizeException | BadPaddingException e) {
```
kevinleturc, line 134: All these exceptions extends `GeneralSecurityException`, could you catch it instead?
```suggestion
		} catch (GeneralSecurityException e) {
```

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+			return null;
```
kevinleturc, line 135: Same thing for re-throwing a NuxeoException with the right error code.

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+	 * Decode an encoded token with passed salt byte[] (default value will be
+	 * used for salt if param is empty/null)
+	 * 
+	 * @param token
+	 *            Encoded String to decode using passed salt
+	 * @param salt
+	 *            SALT to use when decoding
+	 * @return Decoded string from token
+	 */
+	private static String decodeWithSaltByte(String token, byte[] salt) {
```
kevinleturc, line 149:
```suggestion
	private static String decode(String token, byte[] salt) {
```

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
```diff
@@ -0,0 +1,180 @@
...
+			String result = new String(decoded);
+			return result;
+
+		} catch (InvalidKeySpecException | NoSuchAlgorithmException | NoSuchPaddingException | InvalidKeyException
+				| InvalidAlgorithmParameterException | IllegalBlockSizeException | BadPaddingException e) {
+			// use logger in production code
+			return null;
+		}
+	}
+}
```
kevinleturc, line 180: Could you a new line at the end of the file?

modules/platform/nuxeo-easyshare-core/src/main/resources/data/schemas/easysharefolder.xsd
```diff
@@ -5,4 +5,6 @@
   <xs:element name="contactEmail" type="xs:string" />
   <xs:element name="shareComment" type="xs:string" />
   <xs:element name="hasNotification" type="xs:boolean" default="true" />
-</xs:schema>
+  <xs:element name="accessCodeRequired" type="xs:boolean" default="false"/>
+  <xs:element name="accessCode" type="xs:string" />
+</xs:schema>
```
kevinleturc, line 10: Could you add a new line at the end of the file?

modules/platform/nuxeo-easyshare-core/src/main/resources/i18n/messages.properties
```diff
@@ -1,10 +1,18 @@
 easyshare.label.title=Nuxeo - Easyshare
 easyshare.label.description=Nuxeo Easyshare Folder
-easyshare.label.denied=Oh no! This folder doesn't exist.
-easyshare.label.expired=Oh no! This Easy Share folder has expired.
+easyshare.label.denied=Oh no! This access is secured.
```
kevinleturc, line 3: I think we want to keep the denied error as the folder doesn't exist, otherwise, it gives extra information to the user making the request (ie: the document with this id/path exists).

kevinleturc (changes requested):
Thanks for your contribution 👍.

I made several comments we should address, could you have a look?

Thanks

s-guillaume:
> Thanks for your contribution .
> 
> I made several comments we should address, could you have a look?
> 
> Thanks

Hi Kevin, my mistake on this part as I already worked on source code from version 10.10.  I updated the master with the work previously done but I did not merged as I should. I reviewed all classes reporting my changes from the master branch

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShare.java
```diff
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *         http://www.apache.org/licenses/LICENSE-2.0
```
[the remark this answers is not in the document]
s-guillaume: OK, done on the code

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShare.java
```diff
@@ -14,23 +14,28 @@
  * limitations under the License.
  *
  * Contributors:
- *     Michal Obrebski - Nuxeo
+ *         Michal Obrebski - Nuxeo
```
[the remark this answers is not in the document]
s-guillaume: OK, done on the code

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShare.java
```diff
@@ -63,19 +68,21 @@
...
 
     @GET
     public Object doGet() {
         return getView("index");
     }
 
-    public EasyShareUnrestrictedRunner buildUnrestrictedRunner(final String docId, final Long pageIndex) {
+    public EasyShareUnrestrictedRunner buildUnrestrictedRunner(final String jSessionId,
+            final String accessCode,
+            final String docId, final Long pageIndex) {
```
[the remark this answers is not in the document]
s-guillaume: I just worked on features as the previous code was formatted as it in 10.10. I reapply the settings for this project and format classes.

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShare.java
```diff
@@ -91,59 +98,159 @@ public Object run(CoreSession session, IdRef docRef) throws NuxeoException {
                         return getView("expired").arg("docShare", docShare);
                     }
 
-                    DocumentModel document = session.getDocument(new IdRef(docId));
+                    // Check if ref exists preventing user as document not found
+                    // instead of denied
+                    if (session.exists(new IdRef(docId))) {
```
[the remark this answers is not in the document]
s-guillaume: You're right, done

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShare.java
```diff
@@ -63,19 +68,21 @@
 public class EasyShare extends ModuleRoot {
 
     private static final String DEFAULT_PAGE_INDEX = "0";
-
+
```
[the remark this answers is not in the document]
s-guillaume: done

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
s-guillaume: OK, done

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
s-guillaume: Done for "since 11.5" and re-format using nuxeo formater

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
s-guillaume: Done

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
s-guillaume: Done using java.net.URLEncoder and java.net.URLDecoder (for backwards compatibility, used StandardCharsets.UTF-8.toString() instead of StandardCharsets in encode/decode). As Java11 is mandatory now, should remove the toString() no ?

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
s-guillaume: Done

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
s-guillaume: Following code makes sense for you ?
```
        try {

            ...
            // escapes for url
            res = URLEncoder.encode(res, StandardCharsets.UTF_8.toString());

            return res;

        } catch (GeneralSecurityException e) {
            // Error on security part, send 403
            throw new NuxeoException("Error encoding/crypting input data", e, SC_FORBIDDEN);
        } catch (UnsupportedEncodingException e) {
            // Error on conversion part, send 500
            throw new NuxeoException("Error escaping encoded base64 data for URL", e, SC_INTERNAL_SERVER_ERROR);
        }
```

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
s-guillaume: Done

modules/platform/nuxeo-easyshare-core/src/main/resources/i18n/messages.properties
s-guillaume: OK, I can keep the logic in the code but just change the label ? Could you validate this part internally ?

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShareEncryptionUtil.java
s-guillaume: Changed it to remove the try/catch (managed in decodeWithSaltByte throwing NuxeoException

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShare.java
s-guillaume: No backward. the case is :

- Retrieve the JSESSIONID when a client "log in" providing the right accessCode and put it as a URL parameter (j), this information will be propagated for all generated links (browse in folders or download content)
- When connected, the user will click on a link to fetch folder content or download a document (the link is dynamically generated on each refresh)
- On click, retrieve the URL parameter j (previously encoded with the original JSESSIONID) and compare it with the current JSESSIONID
- If encoded URL JSESSIONID and current JSESSIONID do not match, we can consider the client (browser) is not the same (maybe the URL has been copied and transferred to another desktop, ...) and so refused the access

modules/platform/nuxeo-easyshare-core/src/main/java/org/nuxeo/easyshare/EasyShare.java
s-guillaume: Can check my answer : [](https://github.com/nuxeo/nuxeo/pull/4520#discussion_r569420929)

PR #4539, opened 2020-12-11
NXP-29580: Runtime Descriptors Migrations

atchertchian:


modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ExtensionPointImpl.java
```diff
@@ -76,39 +81,24 @@ public String getSuperComponent() {
...
-                }
-            }
-            try {
-                contribs = xmap.loadAll(new XMapContext(extension.getContext()), extension.getElement());
-            } catch (XMapException e) {
+    @Override
+    public XMap getXMap() {
+        XMap xmap = new XMap();
+        for (int i = 0; i < contributions.length; i++) {
+            Class<?> contrib = contributions[i];
```
efge, line 93: You don't need `i`, so use `for (Class<?> contrib : contributions) {`

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/Registry.java
```diff
@@ -32,6 +32,21 @@
      */
     boolean isNull();
 
+    /**
+     * Marks the registry with given identifier.
+     */
+    void mark(String id);
```
efge, line 38: Naming suggestion: `tag` or `label` instead of `mark`.

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/Registry.java
```diff
@@ -32,6 +32,21 @@
      */
     boolean isNull();
 
+    /**
+     * Marks the registry with given identifier.
```
efge, line 36: You should mention that a registry can have not just zero or one but also more than one.

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/DummyRegistry.java
```diff
@@ -0,0 +1,60 @@
...
+public class DummyRegistry extends AbstractRegistry {
+
+    public Optional<Registry> getTargetRegistry() {
+        return Framework.getRuntime()
+                        .getComponentManager()
+                        .getExtensionPointRegistry(ComponentWithXPoint.NAME.getName(), ComponentWithXPoint.XP);
+    }
+
+    @Override
+    public void register(Context ctx, XAnnotatedObject xObject, Element element, String flag) {
```
efge, line 42: `flag` -> `mark` (or whatever new name is chosen)

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/XAnnotatedMembers.java
```diff
@@ -0,0 +1,104 @@
...
+    public XAnnotatedMembers(XMap xmap, XAccessor setter, XNodes anno) {
+        this(xmap, setter, anno.values(), anno.separator(), anno.defaultAssignment());
+    }
+
+    public XAnnotatedMembers(XMap xmap, XAccessor setter, String[] values, String separator, String defaultValue) {
+        super(xmap, setter);
+        XAnnotatedMember[] members = new XAnnotatedMember[values.length];
+        for (int i = 0; i < values.length; i++) {
+            members[i] = new XAnnotatedReference(xmap, String.class, values[i], null, null);
+        }
```
efge, line 54: XAnnotatedMember[] members = Arrays.stream(values)
                                       .map(value -> new XAnnotatedReference(xmap, String.class, value, null, null))
                                       .toArray(XAnnotatedMember[]::new);

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/XAnnotatedMembers.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    protected boolean hasValue(Context ctx, Element element) {
+        for (XAnnotatedMember member : members) {
+            if (member.hasValue(ctx, element)) {
+                return true;
+            }
+        }
+        return false;
```
efge, line 79: return Arrrays.stream(members).anyMatch(member -> member.hasValue(ctx, element));

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/XAnnotatedMembers.java
```diff
@@ -0,0 +1,104 @@
...
+        for (XAnnotatedMember member : members) {
+            if (member.hasValue(ctx, base)) {
+                Object mvalue = member.getValue(ctx, base);
+                values.add(mvalue != null ? String.valueOf(mvalue) : "");
+            }
+        }
+        if (values.isEmpty()) {
+            return getDefaultValue(ctx);
+        }
+        return StringUtils.join(values, separator);
```
efge, line 94: `return String.join(separator, values);`

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/XAnnotatedMembers.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public void toXML(Object instance, Element parent) {
+        for (XAnnotatedMember member : members) {
+            member.toXML(instance, parent);
+        }
```
efge, line 101: `Arrays.asList(members).forEach(member -> member.toXML(instance, parent))`

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/XMap.java
```diff
@@ -270,8 +271,13 @@ private void scanObjectRegistryAnnotations(XAnnotatedObject xob, Class<?> klass)
...
         if (xregistryId != null) {
-            xob.setRegistryId(
-                    new XAnnotatedReference(this, String.class, xregistryId.value(), xregistryId.fallback(), null));
+            if (xregistryId.value().length == 1) {
+                xob.setRegistryId(new XAnnotatedReference(this, String.class, xregistryId.value()[0],
+                        xregistryId.fallback(), xregistryId.defaultAssignment()));
+            } else {
+                xob.setRegistryId(new XAnnotatedMembers(this, null, xregistryId.value(), xregistryId.separator(),
+                        xregistryId.defaultAssignment()));
+            }
```
efge, line 280: `xregistryId.value()` can be put in a local variable

modules/runtime/nuxeo-runtime/src/test/java/org/nuxeo/runtime/TestDescriptorRegistry.java
```diff
@@ -24,9 +24,14 @@
 import java.util.List;
 
 import org.junit.Test;
+import org.nuxeo.common.xmap.registry.Registry;
 import org.nuxeo.runtime.model.Descriptor;
 import org.nuxeo.runtime.model.DescriptorRegistry;
 
+/**
+ * @deprecated since 11.5: use {@link Registry} associated annotations instead.
+ **/
```
efge, line 33: `**/` -> `*/`

modules/runtime/nuxeo-runtime-cluster/src/main/java/org/nuxeo/runtime/cluster/ClusterServiceImpl.java
```diff
@@ -67,7 +66,7 @@ public int getApplicationStartedOrder() {
 
     @Override
     public void start(ComponentContext context) {
-        ClusterNodeDescriptor descr = getDescriptor(XP_CONFIG, UNIQUE_DESCRIPTOR_ID);
+        ClusterNodeDescriptor descr = (ClusterNodeDescriptor) getRegistryContribution(XP_CONFIG).orElse(null);
```
efge, line 69: Is the cast needed?

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
```diff
@@ -194,7 +193,11 @@ protected static KeyValueStore getKeyValueStore() {
     }
 
     public Collection<MigrationDescriptor> getMigrationDescriptors() {
-        return getDescriptors(XP_CONFIG);
+        return getRegistryContributions(XP_CONFIG);
+    }
+
+    protected MigrationDescriptor getMigrationDescriptor(String id) {
+        return (MigrationDescriptor) getRegistryContribution(XP_CONFIG, id).orElse(null);
```
efge, line 200: Cast needed?

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/services/config/TestConfigurationService.java
```diff
@@ -120,6 +120,7 @@ public void testCompatWarn() throws Exception {
         assertEquals(0, Framework.getRuntime().getMessageHandler().getMessages(Level.WARNING).size());
 
         List<String> caughtEvents = logCaptureResult.getCaughtEventMessages();
+        // FIXME
```
efge, line 123: Do you want to keep this comment?

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/services/config/ConfigurationServiceImpl.java
```diff
@@ -191,9 +148,8 @@ protected static boolean startsWithNamespace(String string, String namespace) {
      */
     @Override
     public Optional<String> getString(String key) {
-        return Optional.ofNullable(getDescriptors().get(key))
-                       .map(ConfigurationPropertyDescriptor::getValue)
-                       .filter(StringUtils::isNotBlank);
+        Optional<ConfigurationPropertyDescriptor> desc = getRegistryContribution(CONFIGURATION_EP, key);
+        return Optional.ofNullable(desc.map((value) -> value.getValue()).filter(StringUtils::isNotBlank).orElse(null));
```
efge, line 152: This seems weird. Isn't this simpler?

    desc.map(value -> value.getValue()).filter(StringUtils::isNotBlank);

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/services/config/ConfigurationServiceRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+import org.nuxeo.common.xmap.Context;
+import org.nuxeo.common.xmap.XAnnotatedMember;
+import org.nuxeo.common.xmap.XAnnotatedObject;
+import org.nuxeo.common.xmap.registry.MapRegistry;
+import org.w3c.dom.Element;
+
+/**
+ * Custom registry to accomodate for {@link ConfigurationPropertyDescriptor} merge logics.
+ *
+ * @since TODO
```
efge, line 30: `@since 11.5`

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/registry/SampleComponent.java
```diff
@@ -0,0 +1,86 @@
...
+import org.nuxeo.common.xmap.registry.SingleRegistry;
+import org.nuxeo.runtime.model.ComponentInstance;
+import org.nuxeo.runtime.model.DefaultComponent;
+
+/**
+ * @since 11.5
+ */
+public class SampleComponent extends DefaultComponent {
+
+    public static final String SINGLE_POINT = "single";
```
efge, line 35: To follow other usage I would have named these `SINGLE_XP` etc.

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/annotation/XNode.java
```diff
@@ -41,6 +41,39 @@
...
 
+    /**
+     * Marker for fallback to be ignored.
+     *
+     * @since 11.5
+     */
+    public static final String NO_FALLBACK_MARKER = "__NO_FALLBACK_MARKER__";
+
+    /**
+     * An xpathy expression specifying the fallback XML node to bind to.
```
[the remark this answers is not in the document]
atchertchian: I'm actually not sure as the expression is not "real" xpath, hence "xpath-like" or "xpathy", i can rephrase in all cases.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/ExtensionPoint.java
```diff
@@ -65,4 +68,18 @@
      */
     String getSuperComponent();
 
+    /**
+     * Returns the potential custom registry class for this extension point.
+     *
+     * @since 11.5
+     */
+    String getRegistryClass();
```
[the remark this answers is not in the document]
atchertchian: This depends if we'd like to keep ExtensionPointImpl as a simple POJO.
Also using Optional might interfere with existing serialization logic.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java
```diff
@@ -453,26 +474,127 @@ public synchronized void unregisterExtension(Extension extension) {
...
+                                     .map(xmap::getObject)
+                                     .filter(Objects::nonNull)
+                                     .map(xmap::getRegistry)
+                                     .filter(Objects::nonNull)
+                                     .findFirst()
+                                     .orElse(null);
+                }
+            }
+            if (registry == null) {
+                registry = NULL_REGISTRY;
```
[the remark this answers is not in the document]
atchertchian: I changed it, please tell if still not that-readable

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/Registry.java
atchertchian: Used "tag" instead

modules/runtime/nuxeo-runtime-cluster/src/main/java/org/nuxeo/runtime/cluster/ClusterServiceImpl.java
atchertchian: Yes: "Type mismatch: cannot convert from Object to ClusterNodeDescriptor"

modules/runtime/nuxeo-runtime-migration/src/main/java/org/nuxeo/runtime/migration/MigrationServiceImpl.java
atchertchian: Same: orElse brings ambiguity hence the cast

modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/services/config/TestConfigurationService.java
atchertchian: no :)

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/AbstractRegistry.java
```diff
@@ -0,0 +1,91 @@
...
+
+    protected void setInitialized(boolean initialized) {
+        this.initialized = initialized;
+    }
+
+    protected void checkInitialized() {
+        if (isInitialized()) {
+            return;
+        }
+        initialize();
```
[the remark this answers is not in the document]
atchertchian: Not sure indeed, to be checked

atchertchian:
Thanks for the feedback, here are some TODO notes for this PR:
- discuss + check unresolved conversations
- introduce compat behavior for merge (?) unexpected merges could be turned down for compatibilty using a runtime property (?)
- migrate remaining Descriptor implementations before migrating all other descriptors

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/services/config/ConfigurationServiceRegistry.java
```diff
@@ -0,0 +1,59 @@
...
+        XAnnotatedMember remove = xObject.getRemove();
+        if (remove != null && Boolean.TRUE.equals(remove.getValue(ctx, element))) {
+            contributions.remove(id);
+            return;
+        }
+        ConfigurationPropertyDescriptor contrib = (ConfigurationPropertyDescriptor) xObject.newInstance(ctx, element);
+        if (Framework.getProperties().containsKey(id)) {
+            String message = "Property '" + id + "' should now be contributed to extension "
+                    + "point 'org.nuxeo.runtime.ConfigurationService', using target 'configuration'";
+            DeprecationLogger.log(message, "7.4");
```
atchertchian, line 48: Note i made the "initialize" method public on registry, and explicitly called on component start, so that these deprecation logs can be shown at framework startup

atchertchian:
There are some remaining Descritptor implementations but they're a bit more complicated to handle or have interactions with other contributions, not yet using registries, so they will be dealt with in later PRs.

Update on the TODO list:
* discuss + check unresolved conversations (about multi-threaded usage mostly for now)
* check tests status

Compatibility behavior for merge has been discarded for now (not easy to turn off merge conditionally in annotations), this will encourage migration, even if a list of impacted extension points could be useful for migration.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/services/config/ConfigurationServiceRegistry.java
atchertchian: (Refined later when working on automation registries, to be able to extract the extension id from tag)

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/annotation/XNode.java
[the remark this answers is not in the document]
efge: Not very important, let's keep as is.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/ExtensionPoint.java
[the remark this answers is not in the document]
efge: Ok let's keeps is simple and old-style.

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/AbstractRegistry.java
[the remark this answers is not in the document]
efge: The registration part seems safe as it's called by the startup thread. However other `getContribution*` calls are potentially done after startup.
So let's play it safe: make the `initialized` flag volatile and use double-checked locking to change it (and also setting it in `setInitialized` should also have a synchronization).
Also the datastructures in the registries (`tags`, `registrations` in the abstract class and other ones in the concrete classes) should all be thread-safe (—> `new ConcurentHashMap`, `ConcurrentHashMap.newKeySet()`, `new CopyOnWriteArrayList`, and read/write the `enabled` flag in a synchronized block too)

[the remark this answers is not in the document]
atchertchian: Pushed in a separate commit for now, please let me know if that matches the need.

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/XAnnotatedObject.java
```diff
@@ -78,21 +93,38 @@ public Path getPath() {
...
+     */
+    public Object newInstance(Context ctx, Element element, Object existing) {
+        if (existing == null) {
+            Object ob;
+            try {
+                ob = constructor.newInstance();
+            } catch (InstantiationException e) {
+                throw new IllegalArgumentException(e);
+            } catch (IllegalAccessException e) {
+                throw new IllegalArgumentException(e);
```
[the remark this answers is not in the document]
atchertchian: I did not see any need of extracting the cause in runtime unit tests so i changed this in a separate commit, to see how it goes with all other tests.

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/SingleRegistry.java
```diff
@@ -34,7 +34,7 @@
 
     protected Object contribution;
```
efge, line 35: `contribution` must be volatile too, as it's read by `getContribution()` which is potentially called from different threads.

efge (changes requested):

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/AbstractRegistry.java
```diff
@@ -49,11 +49,15 @@ public boolean isNull() {
...
     }
 
     protected void setInitialized(boolean initialized) {
-        this.initialized = initialized;
+        synchronized (this) {
+            this.initialized = initialized;
+        }
     }
 
     protected void checkInitialized() {
```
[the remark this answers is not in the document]
atchertchian: Yes but i see no reason of preventing contributions retrieval (so registry initialization) before the framework triggers it when starting.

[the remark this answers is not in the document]
efge: Lazy init (on-demand) is always more complex, as we see. But that should be ok with the changes you have.

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/MapRegistry.java
```diff
@@ -81,6 +82,10 @@ public void initialize() {
     @Override
     protected void register(Context ctx, XAnnotatedObject xObject, Element element) {
         String id = (String) xObject.getRegistryId().getValue(ctx, element);
+        if (id == null) {
+            // prevent NPE on map key
+            id = "null";
```
efge, line 87: Ah good point, the concurrent maps disallow null keys...

efge (dismissed):

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/MapRegistry.java
atchertchian: This was visible in tests (core bulk config declares a logConfig without id, maybe this should be changed)

kevinleturc (approved):
Very nice work 👌.

Just details/suggestions that can be addressed in the future.

Indeed, the automatic type inference is a pain, mainly for Optional because we would like to chain the result directly.
See below my thoughts for future improvements:
- replace cast by manual type inference
- add back APIs with the class parameter, in addition to current APIs
- add a getNullableRegistryContribution for former usages

Let's merge it 💪

atchertchian:
Squashed a few commits (and took latest feedback into account, that should not impact tests), waiting for a last complete validation before merge.

efge (approved):

kevinleturc (approved):

modules/platform/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/serializer/service/TemplateSerializerServiceImpl.java
```diff
@@ -41,16 +43,14 @@
...
+        Optional<SerializerDescriptor> optContrib = getRegistryContribution(EXTENSION_POINT_NAME, id);
+        if (optContrib.isEmpty()) {
+            optContrib = getRegistryContribution(EXTENSION_POINT_NAME, DEFAULT_SERIALIZER_NAME);
+            if (optContrib.isEmpty()) {
                 throw new NuxeoException("UnknownSerializer named " + id);
             }
         }
-
-        return contrib.newInstance();
+        return optContrib.get().newInstance();
```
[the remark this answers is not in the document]
atchertchian: Done with #4575

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/AutomationServerComponent.java
```diff
@@ -55,7 +55,8 @@
     @Override
     public void start(ComponentContext context) {
         super.start(context);
-        List<MarshallerDescriptor> marshallers = getDescriptors(XP_MARSHALLER);
+        List<MarshallerDescriptor> marshallers = ((MarshallerRegistry) getExtensionPointRegistry(
+                XP_MARSHALLER)).getContributionValues();
         marshallers.forEach(m -> {
             writers.addAll(m.writers);
             readers.addAll(m.readers);
```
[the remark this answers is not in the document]
atchertchian: Done with #4575

modules/core/nuxeo-core-cache/src/main/java/org/nuxeo/ecm/core/transientstore/TransientStorageComponent.java
```diff
@@ -81,41 +85,30 @@ public synchronized TransientStore getStore(String name) {
...
-    }
-
     @Override
     public void start(ComponentContext context) {
         // make sure we have a default store
         getStore(DEFAULT_STORE_NAME);
         // instantiate all registered stores
-        getDescriptors(EP_STORE).forEach(desc -> getStore(desc.getId()));
+        List<TransientStoreConfig> stores = getRegistryContributions(EP_STORE);
+        stores.forEach(desc -> getStore(desc.getId()));
```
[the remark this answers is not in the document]
atchertchian: Done with #4575

PR #4541, opened 2020-12-11
component manager listener declaration

atchertchian:
suggested change to ease up registration of component manager listeners (created the PR as draft to see how tests go)

### DIFF
--- a/modules/core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkManagerImpl.java
+++ b/modules/core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkManagerImpl.java
@@ -92,7 +92,7 @@ import io.opencensus.trace.Tracing;
  *
  * @since 5.6
  */
-public class WorkManagerImpl extends DefaultComponent implements WorkManager {
+public class WorkManagerImpl extends DefaultComponent implements WorkManager, ComponentManager.Listener {
 
     private static final Logger log = LogManager.getLogger(WorkManagerImpl.class);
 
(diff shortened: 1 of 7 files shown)


modules/core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkManagerImpl.java
```diff
@@ -426,41 +427,34 @@ public void init() {
...
-                        log.warn("WorkManager processing has been disabled on this node");
-                        return;
-                    }
-                    List<WorkQueueDescriptor> descriptors = getDescriptors(QUEUES_EP);
-                    for (WorkQueueDescriptor descriptor : descriptors) {
-                        activateQueue(descriptor);
-                    }
-                }
+    @Override
+    public void afterStart(ComponentManager mgr, boolean isResume) {
```
efge, line 434: The existing naming for `ComponentManager.Listener` methods gets a bit confusing when we have this method inside a `Component`, because now we have a `start` method that is called when the component is started, but `afterStart` is after the _whole runtime_ has started. Could you look at renaming/aliasing things so that here we can have a method name `afterRuntimeStart` (for instance).

kevinleturc (commented):
Looks good 👌.

modules/core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/WorkManagerImpl.java
atchertchian: This was just a suggestion to improve listeners declaration, we can review the API indeed but i'll create a jira first if we're happy with this way of registering listeners, to address the methods confusion.

atchertchian:
Work-related unit tests are now failing --> to be investigated

atchertchian:
see #4544 (WIP) after creation of related jira issue: https://jira.nuxeo.com/browse/NXP-29969

atchertchian:
see #4544 (WIP) after creation of related jira issue: https://jira.nuxeo.com/browse/NXP-29969

PR #4544, opened 2020-12-14
NXP-29969: component manager listener

atchertchian:


modules/runtime/nuxeo-runtime-test/src/test/java/org/nuxeo/runtime/MockEventsInfo.java
```diff
@@ -0,0 +1,34 @@
...
+ *     Anahide Tchertchian
+ */
+package org.nuxeo.runtime;
+
+/**
+ * @since 11.5
+ */
+public class MockEventsInfo {
+
+    public int beforeStop = 0;
```
efge, line 26: Old code but `= 0` is redundant. You should probably keep Bogdan's name in the contributors as well.

modules/runtime/nuxeo-runtime-stream/src/main/java/org/nuxeo/runtime/stream/StreamServiceImpl.java
```diff
@@ -247,21 +246,18 @@ public void stopProcessors() {
...
-        @Override
-        public void afterStart(ComponentManager mgr, boolean isResume) {
-            // this is called once all components are started and ready
-            log.debug("afterStart");
-            startProcessors();
-        }
+    @Override
+    public void afterRuntimeStart(ComponentManager mgr, boolean isResume) {
+        // this is called once all components are started and ready
+        log.debug("afterStart");
```
efge, line 252: Maybe change the debug messages?

efge (dismissed):

kevinleturc (approved):

atchertchian:
TODO: check regression visible on StreamWorkManager tests

atchertchian:
Reverting migration of WorkManagerImpl and StreamWorkManager listeners: there are intricate initializations impacted, listeners can stay as is.

atchertchian:
Need to refine the behavior: some listeners were previously registered at component start, so listener start methods were following the start order on components (which is not the case anymore, visible in unit tests were UserProfile service expects the RepositoryService to have initialized the repo)

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java
```diff
@@ -1314,16 +1315,30 @@ public void afterDeactivation() {
...
         }
 
         public void afterStart(boolean isResume) {
-            for (Object listener : listeners.getListeners()) {
-                ((ComponentManager.Listener) listener).afterRuntimeStart(ComponentManagerImpl.this, isResume);
-            }
+            sortedForStart().forEach(listener -> listener.afterRuntimeStart(ComponentManagerImpl.this, isResume));
         }
 
         public void beforeStop(boolean isStandby) {
```
efge, line 1344: To be consistent I think that stop should use the reverse order.

modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java
```diff
@@ -1283,49 +1293,63 @@ public void remove(ComponentManager.Listener listener) {
...
-                ((ComponentManager.Listener) listener).beforeStart(ComponentManagerImpl.this, isResume);
-            }
+            sortedForStart().forEach(listener -> listener.beforeRuntimeStart(ComponentManagerImpl.this, isResume));
         }
 
         public void afterStart(boolean isResume) {
-            for (Object listener : listeners.getListeners()) {
-                ((ComponentManager.Listener) listener).afterStart(ComponentManagerImpl.this, isResume);
-            }
+            sortedForStart().forEach(listener -> listener.afterRuntimeStart(ComponentManagerImpl.this, isResume));
```
efge, line 1341: To be consistent I think that stop should use the reverse order.

efge (changes requested):

efge (approved):

efge (dismissed):

atchertchian:
Will rebase on master (trying to understand what can cause validation issues)

atchertchian:
Ftests are actually failing because of ERROR log

`[org.nuxeo.ecm.user.center.profile.UserProfileImporter] No importer configuration could be found`

atchertchian:
Fixed change to UserProfileServiceImpl (should not schedule import if config is null)

efge (approved):

PR #4545, opened 2020-12-15
NXP-29978: add support for idempotency key in request controller filter

atchertchian:
Early draft, to be discussed:
- should filter init params be initialized differently? (through ConfigurationService props for instance)
- does the logic match what you originally had in mind?
- should the response status be stored too?
- is HttpServletResponse.SC_CONFLICT the response status we'd like to have while the request is in progress? What about the content (some RFCs do have specifications for the content in this case)

Thanks

### DIFF
--- a/ftests/nuxeo-server-tests/src/test/java/org/nuxeo/ftest/server/ITNuxeoIdempotentRequestTest.java
+++ b/ftests/nuxeo-server-tests/src/test/java/org/nuxeo/ftest/server/ITNuxeoIdempotentRequestTest.java
@@ -1,0 +1,128 @@
+/*
+ * (C) Copyright 2021 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Anahide Tchertchian
+ */
+package org.nuxeo.ftest.server;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.nuxeo.ecm.platform.web.common.idempotency.NuxeoIdempotentFilter.HEADER_KEY;
+import static org.nuxeo.functionaltests.AbstractTest.NUXEO_URL;
+import static org.nuxeo.functionaltests.Constants.ADMINISTRATOR;
+
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.Test;
+import org.nuxeo.client.NuxeoClient;
+import org.nuxeo.client.objects.Document;
+import org.nuxeo.ecm.core.query.sql.NXQL;
+import org.nuxeo.ecm.platform.dublincore.constants.DublinCoreConstants;
+import org.nuxeo.functionaltests.RestHelper.NuxeoClientForNuxeo;
+
+/**
+ * Tests for idempotent request mechanism.
+ *
+ * @since 11.5
+ */
+public class ITNuxeoIdempotentRequestTest {
+
+    private static final String TEST_KEY = "idempotenttestkey" + new Date().getTime();
+
+    private static final NuxeoClient.Builder CLIENT_BUILDER = new NuxeoClientForNuxeo.BuilderForNuxeo().url(
+            NUXEO_URL).authentication(ADMINISTRATOR, ADMINISTRATOR).schemas("*");
+
+    private static final NuxeoClient CLIENT = CLIENT_BUILDER.connect();
+
+    private static final NuxeoClient IDEMPOTENT_CLIENT = CLIENT_BUILDER.header(HEADER_KEY, TEST_KEY).connect();
+
+    private static final String PARENT_PATH = "/default-domain/workspaces/";
+
+    private static final String QUERY_CHILDREN = String.format("SELECT * FROM Document WHERE %s STARTSWITH '%s'",
+            NXQL.ECM_PATH, PARENT_PATH);
+
+    private static final String TEST_TITLE = "testdoc";
+
+    private static final String TEST_TYPE = "File";
+
+    protected String createDocument(NuxeoClient client) {
+        Document document = Document.createWithName(TEST_TITLE, TEST_TYPE);
+        document.setProperties(Map.of(DublinCoreConstants.DUBLINCORE_TITLE_PROPERTY, TEST_TITLE));
+        Document created = client.repository().createDocumentByPath(PARENT_PATH, document);
+        waitForAsyncWork();
+        return created.getId();
+    }
+
+    /**
+     * Prevents from random failures when counting children.
+     */
+    protected void waitForAsyncWork() {
+        Map<String, Object> parameters = new HashMap<>();
+        parameters.put("timeoutSecond", Integer.valueOf(110));
+        parameters.put("refresh", Boolean.TRUE);
+        parameters.put("waitForAudit", Boolean.TRUE);
+        CLIENT.operation("Elasticsearch.WaitForIndexing").parameters(parameters).execute();
+    }
+
+    protected int getNumberOfChildren() {
+        return CLIENT.repository().query(QUERY_CHILDREN).size() - 1;
+    }
+
+    protected Document fetch(String id) {
+        return CLIENT.repository().fetchDocumentById(id);
+    }
+
+    protected void delete(String id) {
+        CLIENT.repository().deleteDocument(id);
+    }
+
+    @Test
+    public void testIdempotentRequest() {
+        String id = null;
+        String dupeId = null;
+        try {
+            assertEquals(0, getNumberOfChildren());
+
+            // create child document
+            id = createDocument(IDEMPOTENT_CLIENT);
+            assertEquals(TEST_TITLE, fetch(id).getTitle());
+            // doc created
+            assertEquals(1, getNumberOfChildren());
+
+            // try creating it again
+            dupeId = createDocument(IDEMPOTENT_CLIENT);
+            assertEquals(id, dupeId);
+            // dupe not created
+            assertEquals(1, getNumberOfChildren());
+
+            // create again without idempotency key
+            dupeId = createDocument(CLIENT);
+            assertNotEquals(id, dupeId);
+            // dupe created
+            assertEquals(2, getNumberOfChildren());
+        } finally {
+            if (id != null) {
+                delete(id);
+            }
+            if (dupeId != null && !dupeId.equals(id)) {
+                delete(dupeId);
+            }
+        }
+    }
+
+}
--- a/modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/CopyingResponseWrapper.java
+++ b/modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/CopyingResponseWrapper.java
@@ -1,0 +1,119 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Anahide Tchertchian
+ */
+package org.nuxeo.ecm.platform.web.common.idempotency;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
+import java.nio.file.Files;
+
+import javax.servlet.ServletOutputStream;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpServletResponseWrapper;
+
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.output.DeferredFileOutputStream;
+
+/**
+ * Response wrapper that can capture response result, using a {@link DeferredFileOutputStream}.
+ *
+ * @since 11.5
+ */
+public class CopyingResponseWrapper extends HttpServletResponseWrapper implements AutoCloseable {
+
+    // invariant: only one of output or writer may be non-null, never both at the same time
+    protected ServletOutputStream output;
+
+    protected PrintWriter writer;
+
+    protected final DeferredFileOutputStream copy;
+
+    public CopyingResponseWrapper(int threshold, HttpServletResponse response) {
+        super(response);
+        copy = new DeferredFileOutputStream(threshold, response.getBufferSize(), "nxidem", null, null);
+    }
+
+    protected CopyingServletOutputStream getCopyingOutputStream() throws IOException {
+        return new CopyingServletOutputStream(getResponse().getOutputStream(), copy);
+    }
+
+    @Override
+    public ServletOutputStream getOutputStream() throws IOException {
+        if (writer != null) {
+            throw new IllegalStateException("getWriter() has already been called on this response.");
+        }
+        if (output == null) {
+            output = getCopyingOutputStream();
+        }
+        return output;
+    }
+
+    @Override
+    public PrintWriter getWriter() throws IOException {
+        if (output != null) {
+            throw new IllegalStateException("getOutputStream() has already been called on this response.");
+        }
+        if (writer == null) {
+            writer = new PrintWriter(new OutputStreamWriter(getCopyingOutputStream(), getCharacterEncoding()));
+        }
+        return writer;
+    }
+
+    @Override
+    public void flushBuffer() throws IOException {
+        super.flushBuffer();
+
+        if (writer != null) {
+            writer.flush();
+        } else if (output != null) {
+            output.flush();
+        }
+    }
+
+    @Override
+    public void close() throws IOException {
+        if (writer != null) {
+            writer.close();
+        } else if (output != null) {
+            output.close();
+        }
+        if (!copy.isInMemory()) {
+            // tmp file cleanup
+            File file = copy.getFile();
+            if (file != null) {
+                Files.delete(file.toPath());
+            }
+        }
+    }
+
+    public long getCopySize() {
+        return copy.getByteCount();
+    }
+
+    public byte[] getCopyAsBytes() throws IOException {
+        if (copy.isInMemory()) {
+            return copy.getData();
+        } else {
+            copy.flush();
+            return FileUtils.readFileToByteArray(copy.getFile());
+        }
+    }
+
+}
--- a/modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
+++ b/modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
@@ -1,0 +1,199 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Anahide Tchertchian
+ */
+package org.nuxeo.ecm.platform.web.common.idempotency;
+
+import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
+import static javax.servlet.http.HttpServletResponse.SC_CONFLICT;
+import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
+
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Set;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpFilter;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.http.client.methods.HttpDelete;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpHead;
+import org.apache.http.client.methods.HttpOptions;
+import org.apache.http.client.methods.HttpPut;
+import org.apache.http.client.methods.HttpTrace;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.nuxeo.runtime.api.Framework;
+import org.nuxeo.runtime.kv.KeyValueService;
+import org.nuxeo.runtime.kv.KeyValueStore;
+import org.nuxeo.runtime.services.config.ConfigurationService;
+
+/**
+ * Filter handling an idempotency key in POST requests.
+ * <p>
+ * If {@link #HEADER_KEY} is found in the request header, will intercept request handling to:
+ * <ul>
+ * <li>mark the request as being processed
+ * <li>capture the response when request was processed without any error and store it
+ * <li>return the stored response if a subsequent request with the same key is processed again
+ * <li>return a conflict response if a request with the same key is processed while the first request is still in
+ * progress.
+ * </ul>
+ *
+ * @since 11.5
+ */
+public class NuxeoIdempotentFilter extends HttpFilter {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final Logger log = LogManager.getLogger(NuxeoIdempotentFilter.class);
+
+    public static final String HEADER_KEY = "Idempotency-Key";
+
+    public static final String STORE_PROPERTY = "org.nuxeo.request.idempotency.keyvaluestore.name";
+
+    public static final String DEFAULT_STORE = "idempotentrequest";
+
+    protected static final Duration DEFAULT_TTL = Duration.ofDays(1);
+
+    public static final String TTL_DURATION_PROPERTY = "org.nuxeo.request.idempotency.ttl.duration";
+
+    public static final String INPROGRESS_MARKER = "{\"inprogress\":true}";
+
+    public static final String INFO_SUFFIX = "_info";
+
+    protected static final int DEFERRED_OUTPUT_STREAM_THRESHOLD = 1024 * 1024; // 1 MB
+
+    protected static final int MAX_CONTENT_SIZE = 1024 * 1024 * 5; // 5 MB
+
+    protected static final Set<String> IDEMPOTENT_METHODS = Set.of(
+            // safe methods according to RFC 7231 4.2.1
+            HttpGet.METHOD_NAME, HttpHead.METHOD_NAME, HttpOptions.METHOD_NAME, HttpTrace.METHOD_NAME,
+            // idempotent methods according to RFC 7231 4.2.2
+            HttpPut.METHOD_NAME, HttpDelete.METHOD_NAME);
+
+    protected Duration getTTL() {
+        ConfigurationService cs = Framework.getService(ConfigurationService.class);
+        if (cs != null) {
+            return cs.getDuration(TTL_DURATION_PROPERTY, DEFAULT_TTL);
+        }
+        return DEFAULT_TTL;
+    }
+
+    protected String getStoreName() {
+        ConfigurationService cs = Framework.getService(ConfigurationService.class);
+        if (cs != null) {
+            return cs.getString(STORE_PROPERTY, DEFAULT_STORE);
+        }
+        return DEFAULT_STORE;
+    }
+
+    @Override
+    protected void doFilter(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        if (!doFilterIdempotent(request, response, chain)) {
+            chain.doFilter(request, response);
+        }
+    }
+
+    protected boolean doFilterIdempotent(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        String method = request.getMethod();
+        if (IDEMPOTENT_METHODS.contains(method)) {
+            log.debug("No idempotent processing done: method is already idempotent: {}", method);
+            return false;
+        }
+        String key = request.getHeader(HEADER_KEY);
+        if (key == null) {
+            log.debug("No idempotent processing done: no {} header present", HEADER_KEY);
+            return false;
+        }
+        log.debug("Idempotent request key: {}", key);
+        KeyValueService kvs = Framework.getService(KeyValueService.class);
+        if (kvs == null) {
+            log.debug("KeyValueService not present");
+            return false;
+        }
+        KeyValueStore store = kvs.getKeyValueStore(getStoreName());
+        String storeStatus = store.getString(key + INFO_SUFFIX);
+        byte[] storeContent = store.get(key);
+        if (storeStatus == null) {
+            log.debug("Handle new request for key: {}", key);
+            long ttl = getTTL().toSeconds();
+            store.put(key + INFO_SUFFIX, INPROGRESS_MARKER, ttl);
+            try {
+                try (CopyingResponseWrapper wrapper = new CopyingResponseWrapper(DEFERRED_OUTPUT_STREAM_THRESHOLD,
+                        response)) {
+                    wrapper.setHeader(HEADER_KEY, key);
+                    chain.doFilter(request, wrapper);
+                    wrapper.flushBuffer();
+                    long size = wrapper.getCopySize();
+                    if (size > MAX_CONTENT_SIZE) {
+                        log.debug(
+                                "Not storing response for key: {} (status: {}, size in bytes: {}), max content size exceeded: {}",
+                                key, wrapper.getStatus(), size, MAX_CONTENT_SIZE);
+                        return true;
+                    }
+                    byte[] content = wrapper.getCopyAsBytes();
+                    store.put(key, content, ttl);
+                    store.put(key + INFO_SUFFIX, NuxeoIdempotentResponse.save(wrapper), ttl);
+                    log.debug("Stored response for key: {} (status: {}, size in bytes: {})", key, wrapper.getStatus(),
+                            content.length);
+                    return true;
+                }
+            } catch (IOException | ServletException e) {
+                if (!response.isCommitted()) {
+                    response.setStatus(SC_INTERNAL_SERVER_ERROR);
+                    response.setHeader(HEADER_KEY, key);
+                }
+                throw e;
+            } finally {
+                if (response.getStatus() >= SC_BAD_REQUEST) {
+                    // error request: cleanup store
+                    store.put(key, (String) null);
+                    store.put(key + INFO_SUFFIX, (String) null);
+                    log.debug("Cleanup store: error for key: {}", key);
+                }
+            }
+        } else if (INPROGRESS_MARKER.equals(storeStatus)) {
+            // request already in progress -> conflict
+            // Don't call response.sendError, because it commits the response
+            // which prevents NuxeoExceptionFilter from returning a custom error page.
+            response.setStatus(SC_CONFLICT);
+            response.setHeader(HEADER_KEY, key);
+            log.debug("Conflict response for key: {}", key);
+            return true;
+        } else {
+            try {
+                // request already done: return stored result
+                NuxeoIdempotentResponse.restore(response, storeStatus.getBytes());
+                response.setHeader(HEADER_KEY, key);
+                response.getOutputStream().write(storeContent);
+                log.debug("Returning stored response for key: {} (status: {}, size in bytes: {})", key,
+                        response.getStatus(), storeContent.length);
+                return true;
+            } catch (IOException e) {
+                log.error("Error processing stored result for key: {}", key, e);
+                return false;
+            }
+        }
+    }
+
+}
--- a/modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentResponse.java
+++ b/modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentResponse.java
@@ -1,0 +1,109 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Anahide Tchertchian
+ */
+package org.nuxeo.ecm.platform.web.common.idempotency;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+
+import javax.servlet.http.HttpServletResponse;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.MapperFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.ObjectWriter;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
+/**
+ * POJO representing response to be serialized and served by {@link NuxeoIdempotentFilter}.
+ *
+ * @since 11.5
+ */
+public class NuxeoIdempotentResponse {
+
+    public static final Set<String> SKIPPED_HEADERS = Set.of("Transfer-Encoding");
+
+    protected int status;
+
+    protected Map<String, Collection<String>> headers = new LinkedHashMap<>();
+
+    protected static final ObjectMapper MAPPER = //
+            new ObjectMapper().configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true)
+                              .configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true);
+
+    protected static final ObjectReader READ_MAPPER = MAPPER.readerFor(NuxeoIdempotentResponse.class)
+                                                            .withoutRootName()
+                                                            .without(JsonParser.Feature.AUTO_CLOSE_SOURCE);
+
+    protected static final ObjectWriter WRITE_MAPPER = MAPPER.writerFor(NuxeoIdempotentResponse.class)
+                                                             .withoutRootName()
+                                                             .with(JsonGenerator.Feature.FLUSH_PASSED_TO_STREAM)
+                                                             .without(JsonGenerator.Feature.AUTO_CLOSE_TARGET);
+
+    public int getStatus() {
+        return status;
+    }
+
+    public void setStatus(int status) {
+        this.status = status;
+    }
+
+    public Map<String, Collection<String>> getHeaders() {
+        return headers;
+    }
+
+    public void setHeader(String name, Collection<String> value) {
+        headers.put(name, value);
+    }
+
+    public static final void restore(HttpServletResponse response, byte[] bytes) throws IOException {
+        NuxeoIdempotentResponse stored = READ_MAPPER.readValue(bytes);
+        response.setStatus(stored.getStatus());
+        stored.getHeaders().forEach((name, values) -> {
+            boolean isFirst = true;
+            for (String value : values) {
+                if (SKIPPED_HEADERS.contains(name)) {
+                    continue;
+                }
+                if (isFirst) {
+                    response.setHeader(name, value);
+                    isFirst = false;
+                } else {
+                    response.addHeader(name, value);
+                }
+            }
+        });
+    }
+
+    public static final byte[] save(HttpServletResponse response) throws JsonProcessingException {
+        NuxeoIdempotentResponse stored = new NuxeoIdempotentResponse();
+        stored.setStatus(response.getStatus());
+        response.getHeaderNames()
+                .stream()
+                .filter(h -> !SKIPPED_HEADERS.contains(h))
+                .forEach(name -> stored.setHeader(name, response.getHeaders(name)));
+        return WRITE_MAPPER.writeValueAsBytes(stored);
+    }
+
+}
--- a/modules/platform/nuxeo-platform-web-common/src/main/resources/OSGI-INF/idempotency-configuration.xml
+++ b/modules/platform/nuxeo-platform-web-common/src/main/resources/OSGI-INF/idempotency-configuration.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0"?>
+<component name="org.nuxeo.ecm.platform.web.idempotency">
+
+  <extension target="org.nuxeo.runtime.ConfigurationService" point="configuration">
+    <documentation>
+      Properties controlling idempotent requests TTL and KeyValue storename.
+
+      Default TTL in seconds matches 1 day.
+
+      @since 11.5
+    </documentation>
+    <property name="org.nuxeo.request.idempotency.ttl.duration">1d</property>
+    <property name="org.nuxeo.request.idempotency.keyvaluestore.name">idempotentrequest</property>
+  </extension>
+
+
+</component>
--- a/modules/platform/nuxeo-platform-web-common/src/test/java/org/nuxeo/ecm/platform/web/idempotency/TestNuxeoIdempotentFilter.java
+++ b/modules/platform/nuxeo-platform-web-common/src/test/java/org/nuxeo/ecm/platform/web/idempotency/TestNuxeoIdempotentFilter.java
@@ -1,0 +1,406 @@
+/*
+ * (C) Copyright 2020 Nuxeo (http://nuxeo.com/) and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributors:
+ *     Anahide Tchertchian
+ */
+package org.nuxeo.ecm.platform.web.idempotency;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static javax.servlet.http.HttpServletResponse.SC_CONFLICT;
+import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+import static javax.servlet.http.HttpServletResponse.SC_OK;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.fail;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyInt;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.nuxeo.ecm.platform.web.common.idempotency.NuxeoIdempotentResponse.SKIPPED_HEADERS;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.inject.Inject;
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.ServletOutputStream;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpPost;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+import org.nuxeo.ecm.platform.web.common.idempotency.NuxeoIdempotentFilter;
+import org.nuxeo.runtime.kv.KeyValueService;
+import org.nuxeo.runtime.kv.KeyValueStoreProvider;
+import org.nuxeo.runtime.test.runner.Deploy;
+import org.nuxeo.runtime.test.runner.Features;
+import org.nuxeo.runtime.test.runner.FeaturesRunner;
+import org.nuxeo.runtime.test.runner.RuntimeFeature;
+
+/**
+ * Checks idempotent requests management.
+ *
+ * @since 11.5
+ */
+@RunWith(FeaturesRunner.class)
+@Features(RuntimeFeature.class)
+@Deploy("org.nuxeo.runtime.kv")
+@Deploy("org.nuxeo.ecm.platform.web.common:OSGI-INF/idempotency-configuration.xml")
+public class TestNuxeoIdempotentFilter {
+
+    protected static final String KEY = "mykey";
+
+    protected static final String CONTENT = "test content";
+
+    protected static final Map<String, Collection<String>> RESPONSE_HEADERS = new LinkedHashMap<>();
+
+    static {
+        RESPONSE_HEADERS.put("Accept", List.of("text/html", "application/xhtml+xml", "*/*;q=0.8"));
+        RESPONSE_HEADERS.put("Connection", List.of("Keep-Alive"));
+        RESPONSE_HEADERS.put("Content-Encoding", List.of("gzip"));
+        RESPONSE_HEADERS.put("Content-Type", List.of("text/html; charset=utf-8"));
+        RESPONSE_HEADERS.put("Set-Cookie", List.of("sessionId=38afes7a8", "id=a3fWa; Max-Age=2592000"));
+        RESPONSE_HEADERS.put("Transfer-Encoding", List.of("chunked")); // should be filtered
+    }
+
+    protected static final Map<String, Collection<String>> KEY_RESPONSE_HEADERS = new LinkedHashMap<>();
+
+    static {
+        KEY_RESPONSE_HEADERS.put(NuxeoIdempotentFilter.HEADER_KEY, List.of(KEY));
+    }
+
+    protected static final Map<String, Collection<String>> FINAL_RESPONSE_HEADERS = new LinkedHashMap<>();
+
+    static {
+        FINAL_RESPONSE_HEADERS.putAll(RESPONSE_HEADERS);
+        FINAL_RESPONSE_HEADERS.putAll(KEY_RESPONSE_HEADERS);
+    }
+
+    protected static final Map<String, Collection<String>> FINAL_COPY_RESPONSE_HEADERS = new LinkedHashMap<>();
+
+    static {
+        FINAL_COPY_RESPONSE_HEADERS.putAll(FINAL_RESPONSE_HEADERS);
+        FINAL_COPY_RESPONSE_HEADERS.keySet().removeAll(SKIPPED_HEADERS);
+    }
+
+    protected NuxeoIdempotentFilter filter;
+
+    protected FilterChain chain;
+
+    protected HttpServletRequest request;
+
+    protected MockResponse mockResponse;
+
+    @Inject
+    protected KeyValueService kvs;
+
+    protected KeyValueStoreProvider store;
+
+    protected static class MockResponse {
+
+        protected HttpServletResponse response;
+
+        protected int status;
+
+        protected Map<String, Collection<String>> headers = new LinkedHashMap<>();
+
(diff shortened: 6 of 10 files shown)


atchertchian:
Additional concerns:
- for now GET and POST requests are handled, but I'm concerned about flooding KVS in some cases (blob download for instance)
- any suggestions about writing adequate functional tests on this?

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
```diff
@@ -70,13 +72,36 @@
 
     protected static final int LOCK_TIMEOUT_S = 120;
 
+    protected static final String IDEMPOTENCY_KEY = "NuxeoIdempotencyKey";
```
BoboTiG, line 75: Could we use something more generic, like `Idempotency-Key` or `X-Idempotency-Key` maybe?

BoboTiG:
I am not sure we should do it for `GET` requests though 🤔

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
```diff
@@ -295,4 +319,34 @@ protected void addCacheHeaders(HttpServletResponse response, RequestFilterConfig
...
+                    store.put(key, IDEMPOTENCY_INPROGRESS_MARKER, idempotencyTTLSeconds);
+                    CopyingHttpServletResponse wrapper = new CopyingHttpServletResponse(response);
+                    chain.doFilter(request, wrapper);
+                    store.put(key, wrapper.getCopy(), idempotencyTTLSeconds);
+                } else if (IDEMPOTENCY_INPROGRESS_MARKER.equals(value)) {
+                    // request already in progress -> conflict (?) any content?
+                    response.sendError(HttpServletResponse.SC_CONFLICT);
+                } else {
+                    // request already done: return stored result
+                    // XXX get back status code?
```
BoboTiG, line 341: https://github.com/paypal/api-standards/blob/master/patterns.md#idempotency explains that should return HTTP 200 + the result.

When it is a new request, it should return HTTP 201  + the result.

I am not familiar with the code here, does that return such HTTP codes?

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
atchertchian: Of course, I was not sure if we wanted this header to be Nuxeo-specific so i just went with this value for now

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
atchertchian: I went with conflict because that's what's documented here https://tools.ietf.org/html/draft-idempotency-header-01.
As for 201, this filter cannot know this is for creation so no sure 201 applies here, so the code is left unchanged compared to a request that would go through without idempotency.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
```diff
@@ -70,13 +72,36 @@
...
 
+    protected static final String IDEMPOTENCY_KEY = "NuxeoIdempotencyKey";
+
+    protected static final String IDEMPOTENCY_STORE_NAME_PROPERTY = "idempotency_keyvaluestore";
+
+    protected static final String IDEMPOTENCY_TTL_SECONDS_PARAM = "idempotency_ttl_seconds";
+
+    protected static final String IDEMPOTENCY_INPROGRESS_MARKER = "IDEMPOTENCY_INPROGRESS_MARKER";
+
+    protected int idempotencyTTLSeconds = 300;
```
BoboTiG, line 83: 5 minute seems quite small, no?

atchertchian: I put a (quite) low TTL for testing, we can set a better default value indeed.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
BoboTiG: Conflict seems the good error to throw.

200 or 201, I would say this is fine as-is. As soon as this is not an error code :)

BoboTiG:
> is HttpServletResponse.SC_CONFLICT the response status we'd like to have while the request is in progress? What about the content (some RFCs do have specifications for the content in this case)

I would not bother with contents. Just returning the HTTP 409 is enough IMO.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
atchertchian: Well it will be an error code if there's indeed an error while processing the request.
That's also why we may want the code to be stored too the first time that the request is processed.

BoboTiG: Hmm how would you make the difference between an HTTP 500 error and a "valid" error like a 404? The 500 error should not be cached as another call may unlock the real result. And a 404 now may not be revelant 2 minutes later. Same for 401 and 403 errors for example.

atchertchian: Ok so maybe the code should be checked before storing the result, and key removed from the store in this case?

atchertchian:
Reviewed the code and added unit tests, will refine according to feedback and leave open conversations about subjects still to be discussed.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
atchertchian: I moved the key in the code but we should still discuss naming, so keeping this conversation open

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
atchertchian: I changed the processing so that in case of error, store content is nullified for this key

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
atchertchian: I moved these values but we should still discuss the default value for TTL

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    public IdempotentRequestHandler(String storename, int ttl) {
+        this.store = storename;
+        this.ttl = ttl;
+    }
+
+    public void handle(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        if (!"POST".equals(request.getMethod())) {
```
atchertchian, line 101: Only handling POST requests --> to be discussed, maybe

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,142 @@
...
+
+    protected final String store;
+
+    protected final int ttl;
+
+    public IdempotentRequestHandler(FilterConfig filterConfig) {
+        // default values
+        int ttl = 330;
+        String store = "requestcontroller";
+        // retrieve potential configuration from filter
```
atchertchian, line 77: Maybe should retrieve configuration from the ConfigurationService instead?

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,142 @@
...
+            } finally {
+                if (cleanup || response.getStatus() >= 400) {
+                    // error request: cleanup store
+                    store.put(IdempotentRequestHandler.CONTENT_PREFIX + key, (String) null, ttl);
+                    store.put(IdempotentRequestHandler.STATUS_PREFIX + key, (String) null, ttl);
+                }
+            }
+        } else if (IdempotentRequestHandler.INPROGRESS_MARKER.equals(storedContent)) {
+            // request already in progress -> conflict
+            response.sendError(HttpServletResponse.SC_CONFLICT,
```
atchertchian, line 133: Using CONFLICT status + mentioning key in the response body -> to be discussed, maybe?

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,144 @@
...
+        String storedContent = store.getString(IdempotentRequestHandler.CONTENT_PREFIX + key);
+        if (storedContent == null) {
+            // new request
+            boolean cleanup = false;
+            store.put(IdempotentRequestHandler.CONTENT_PREFIX + key, IdempotentRequestHandler.INPROGRESS_MARKER, ttl);
+            try {
+                CopyingResponseWrapper wrapper = new CopyingResponseWrapper(response);
+                chain.doFilter(request, wrapper);
+                store.put(IdempotentRequestHandler.CONTENT_PREFIX + key, wrapper.getCaptureAsString(), ttl);
+                store.put(IdempotentRequestHandler.STATUS_PREFIX + key, String.valueOf(wrapper.getStatus()), ttl);
```
efge, line 122: One of them, probably the content, could have no prefix at all. And for the other I would have used a suffix, to make listings/dumps easier to read.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,144 @@
...
+            } catch (IOException | ServletException e) {
+                cleanup = true;
+                throw e;
+            } finally {
+                if (cleanup || response.getStatus() >= 400) {
+                    // error request: cleanup store
+                    store.put(IdempotentRequestHandler.CONTENT_PREFIX + key, (String) null, ttl);
+                    store.put(IdempotentRequestHandler.STATUS_PREFIX + key, (String) null, ttl);
```
efge, line 130: Don't use a TTL when putting `null`

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,144 @@
...
+                }
+            }
+        } else if (IdempotentRequestHandler.INPROGRESS_MARKER.equals(storedContent)) {
```
efge, line 133: For in-progress it's better to use a special status value (negative?) instead of the content. This way the content is pure content and not mixed with other logic.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,144 @@
...
+            // request already in progress -> conflict
+            response.sendError(HttpServletResponse.SC_CONFLICT,
+                    String.format("Idempotent request already in progress for key '%s'", key));
+        } else {
+            // request already done: return stored result
+            response.getWriter().write(storedContent);
```
efge, line 139: As mentioned you should use bytes, not string. So use `getOutputStream` here instead of `getWriter`.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,144 @@
...
+            response.setStatus(Integer.valueOf(store.getString(IdempotentRequestHandler.STATUS_PREFIX + key)));
```
efge, line 140: In addition to the status we must also save/restore all headers (which will include cookies and values like content-type and content-length).
So the whole status key in the k/v store could contain some JSON with this info (+ the http status)

modules/platform/nuxeo-platform-web-common/src/test/java/org/nuxeo/ecm/platform/web/requestcontroller/filter/TestIdempotentRequestFilter.java
```diff
@@ -0,0 +1,345 @@
...
+        checkStore(null, null);
+    }
+
+    protected void setResult(HttpServletResponse response, int status, String content) throws IOException {
+        response.setStatus(status);
+        response.getWriter().write(content);
+    }
+
+    @Test
+    public void testPostRequest() throws Exception {
```
efge, line 231: In addition to this base test we should test for content that is not representable as a string, and headers (content type, custom headers, cookies).

efge (changes requested):

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
atchertchian: I'm not sure if you mean is should filter out all of these instead of accepting just POST?
Also javax.ws.rs.HttpMethod declares these as constants, except TRACE, so not sure the full list is already available somewhere else?

efge: Yes I mean filter out all these.
See what `NuxeoCorsCsrfFilter` does for the constants.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
efge: `Idempotency-Key` is good it's what's in the draft RFC and used by lots of organizations already.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
efge: The base default should be at least one day I think. Some organizations use 30 days as TTL!

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
efge: Yes it'd be easier to change through contributions rather than changing a filter config.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
efge: I don't think it'll help clients, they won't read the body in case of 409. Just debug logs is enough.

efge:
I get the testing as a separate servlet filter, but in the end this should move to `NuxeoRequestControllerFilter` which already has some buffering logic (although it should be rethought together with the CopyingServletOutputStream if we want to have never do streaming for non-GET requests if there's an idempotency key).

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/CopyingServletOutputStream.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Captures content written to the target stream.
+ *
+ * @since 11.5
+ */
+public class CopyingServletOutputStream extends ServletOutputStream {
+
+    protected final ServletOutputStream output;
+
+    protected final ByteArrayOutputStream capture;
```
efge, line 36: Good for an initial prototype, but in general we can't buffer in memory.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,144 @@
...
+            try {
+                CopyingResponseWrapper wrapper = new CopyingResponseWrapper(response);
+                chain.doFilter(request, wrapper);
+                store.put(IdempotentRequestHandler.CONTENT_PREFIX + key, wrapper.getCaptureAsString(), ttl);
+                store.put(IdempotentRequestHandler.STATUS_PREFIX + key, String.valueOf(wrapper.getStatus()), ttl);
+            } catch (IOException | ServletException e) {
+                cleanup = true;
+                throw e;
+            } finally {
+                if (cleanup || response.getStatus() >= 400) {
```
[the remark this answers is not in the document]
atchertchian: ? this is still an error response that KV should not store, right?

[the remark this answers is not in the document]
efge: Yes I just mean use the constant instead of a number.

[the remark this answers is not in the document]
atchertchian: ah ok :+1:

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,144 @@
...
+
+    public IdempotentRequestHandler(String storename, int ttl) {
+        this.store = storename;
+        this.ttl = ttl;
+    }
+
+    public void handle(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        if (!"POST".equals(request.getMethod())) {
+            chain.doFilter(request, response);
```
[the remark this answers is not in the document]
atchertchian: I added a number of debug logs, let me know if they should be changed.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,168 @@
...
+    /** @since 11.5 */
+    public static final String HEADER_KEY = "Idempotency-Key";
+
+    /** @since 11.5 */
+    public static final String STORE_PROPERTY = "org.nuxeo.request.controller.keyvaluestore.name";
+
+    protected static final String DEFAULT_STORE = "idempotentrequest";
+
+    /** 1 day */
+    protected static final int DEFAULT_TTL_SECONDS = 86400;
```
efge, line 71: FYI for new code I'm trying to use the idiom `long DEFAULT_TTL_SECONDS = Duration.ofDays(1).toSeconds()` to be explicit. And make the ttl a long everywhere

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,168 @@
...
+
+    /** @since 11.5 */
+    public static final String TTL_SECONDS_PROPERTY = "org.nuxeo.request.controller.ttl.seconds";
```
efge, line 74: This should mention request idempotency instead of `request.controller` (which is an implementation detail)

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,168 @@
...
+
+    /** @since 11.5 */
+    public static final String INPROGRESS_MARKER = "IDEMPOTENCY_INPROGRESS_MARKER";
+
+    /** @since 11.5 */
+    public static final String STATUS_SUFFIX = "_status";
+
+    // safe methods according to RFC 7231 4.2.1
+    protected static final Set<String> SAFE_METHODS = new HashSet<>(
+            Arrays.asList(HttpGet.METHOD_NAME, HttpHead.METHOD_NAME, HttpOptions.METHOD_NAME, HttpTrace.METHOD_NAME));
```
efge, line 84: Do we plan on backporting? Otherwise `Set.of()` is simpler

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,168 @@
...
+
+    protected String getStoreName() {
+        ConfigurationService cs = Framework.getService(ConfigurationService.class);
+        if (cs != null) {
+            return cs.getString(STORE_PROPERTY, DEFAULT_STORE);
+        }
+        return DEFAULT_STORE;
+    }
+
+    protected boolean shouldIgnore(String method) {
```
efge, line 106: Suggest `shouldIgnore` -> `isIdempotentMethod`

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,168 @@
...
+        return SAFE_METHODS.contains(method) || IDEMPOTENT_METHODS.contains(method);
+    }
+
+    public void handle(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        String method = request.getMethod();
+        log.debug("Method: " + method);
+        if (shouldIgnore(method)) {
+            log.debug("Request ignored");
```
efge, line 115: Add detail: `No idempotent processing done: method is already idempotent` (the request isn't really ignored)
I would actually log the method here and not in a separate line.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,168 @@
...
+            chain.doFilter(request, response);
+            return;
+        }
+        String key = request.getHeader(IdempotentRequestHandler.HEADER_KEY);
+        if (key == null) {
+            log.debug("No idempotent request key");
```
efge, line 121: `No idempotent processing done: no Idempotent-Key present`

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,168 @@
...
+            chain.doFilter(request, response);
+            return;
+        }
+        log.debug("Idempotent request key: " + key);
+        KeyValueService kvs = Framework.getService(KeyValueService.class);
+        if (kvs == null) {
+            log.debug("KeyValueStore not present");
```
efge, line 128: `KeyValueService not present`

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,168 @@
...
+            chain.doFilter(request, response);
+            return;
+        }
+        KeyValueStore store = kvs.getKeyValueStore(getStoreName());
```
efge, line 132: Check `store` is not `null` otherwise log and skip idempotent processing.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,168 @@
...
+        if (key == null) {
+            log.debug("No idempotent request key");
+            chain.doFilter(request, response);
+            return;
+        }
+        log.debug("Idempotent request key: " + key);
+        KeyValueService kvs = Framework.getService(KeyValueService.class);
+        if (kvs == null) {
+            log.debug("KeyValueStore not present");
+            chain.doFilter(request, response);
```
efge, line 129: You have many `if (...) { chain.doFilter(request, response); return; }` so maybe let the caller do the chain call if this method returns a boolean `false`.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
```diff
@@ -70,13 +70,15 @@
 
     protected static final int LOCK_TIMEOUT_S = 120;
 
+    protected IdempotentRequestHandler handler;
```
efge, line 73: `handler` -> `idempotentRequestHandler`

efge (changes requested):

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
atchertchian: Alright but that's actually something i found quite strange in the KeyValueService implementation: even when the store with given name is not present, a store is instantiated from the default definition (and an exception is thrown if there is no default definition).

efge: Ah I see. Yes in that case no need for a check you're right.
This behavior is done to avoid explicit definitions of lots of stores in XML.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
```diff
@@ -0,0 +1,168 @@
...
+ * progress.
+ * </ul>
+ *
+ * @since 11.5
+ */
+public class IdempotentRequestHandler {
+
+    private static final Logger log = LogManager.getLogger(IdempotentRequestHandler.class);
+
+    /** @since 11.5 */
```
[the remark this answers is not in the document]
atchertchian: Will update this if/when merging code with filter code.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
atchertchian: I kept this key as is in place of json serialization of response headers + status, let me know if you still think I should change this marker (sorry, renamed and moved all classes...)

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
atchertchian: Added a serialization of response status + header as the stored json.

modules/platform/nuxeo-platform-web-common/src/test/java/org/nuxeo/ecm/platform/web/requestcontroller/filter/TestIdempotentRequestFilter.java
atchertchian: I added quick mock headers, as for the content, not sure if this would belong in this test as everything here is actually mocked... to be discussed again, keeping this conversation open for this.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/CopyingServletOutputStream.java
atchertchian: Reworked to rely on apache DeferredFileOutputStream

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/IdempotentRequestHandler.java
atchertchian: I did not change this, but since in the end there is a dedicated filter maybe it's alright as is?

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/filter/NuxeoRequestControllerFilter.java
atchertchian: As discussed, moved this handler to a dedicated filter class.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+    protected static final Set<String> SAFE_METHODS = Set.of(HttpGet.METHOD_NAME, HttpHead.METHOD_NAME,
+            HttpOptions.METHOD_NAME, HttpTrace.METHOD_NAME);
+
+    // idempotent methods according to RFC 7231 4.2.2
+    protected static final Set<String> IDEMPOTENT_METHODS = Set.of(HttpPut.METHOD_NAME, HttpDelete.METHOD_NAME);
+
+    protected long getTTLSeconds() {
+        ConfigurationService cs = Framework.getService(ConfigurationService.class);
+        if (cs != null) {
+            return cs.getLong(TTL_SECONDS_PROPERTY, DEFAULT_TTL_SECONDS);
```
efge, line 90: Probably better to use `getDuration` here, it's more flexible for users. And then convert to seconds. And rename the property to not have "seconds" in the name or value. Then `DEFAULT_TTL_SECONDS` can be just a duration too, and you only convert to seconds at the last moment: `long ttl = getTTLSeconds()` -> `long ttl = getTTL().toSeconds()`.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+    protected String getStoreName() {
+        ConfigurationService cs = Framework.getService(ConfigurationService.class);
+        if (cs != null) {
+            return cs.getString(STORE_PROPERTY, DEFAULT_STORE);
+        }
+        return DEFAULT_STORE;
+    }
+
+    protected boolean isIdempotentMethod(String method) {
+        return SAFE_METHODS.contains(method) || IDEMPOTENT_METHODS.contains(method);
```
efge, line 104: You could have a single set with everything in it.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain chain)
+            throws IOException, ServletException {
+        HttpServletRequest request = (HttpServletRequest) servletRequest;
+        HttpServletResponse response = (HttpServletResponse) servletResponse;
+        String method = request.getMethod();
+        if (isIdempotentMethod(method)) {
+            log.debug("No idempotent processing done: method is already idempotent: " + method);
```
efge, line 114: With log4j2 you should use `{}` in the log message, not string concatenation.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+            chain.doFilter(request, response);
+            return;
+        }
+        String key = request.getHeader(NuxeoIdempotentFilter.HEADER_KEY);
```
efge, line 118: No need for `NuxeoIdempotentFilter.` in front of all your local constants.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+            long ttl = getTTLSeconds();
+            store.put(key + NuxeoIdempotentFilter.INFO_SUFFIX, NuxeoIdempotentFilter.INPROGRESS_MARKER, ttl);
+            try {
+                CopyingResponseWrapper wrapper = new CopyingResponseWrapper(DEFERRED_OUTPUT_STREAM_THRESHOLD, response);
+                chain.doFilter(request, wrapper);
+                store.put(key, wrapper.getCaptureAsBytes(), ttl);
+                store.put(key + NuxeoIdempotentFilter.INFO_SUFFIX, NuxeoIdempotentResponse.write(wrapper), ttl);
+                log.debug("Stored response for key: " + key);
+            } catch (IOException | ServletException e) {
+                response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
```
efge, line 144: Please do static imports of `HttpServletResponse.SC_*` constants

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentResponse.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public Map<String, Collection<String>> getHeaders() {
+        return headers;
+    }
+
+    public void setHeader(String name, Collection<String> value) {
+        headers.put(name, value);
+    }
+
+    protected static final ObjectMapper getMapper() {
```
efge, line 62: You should put the mapper in a static final field instead of constructing it every time.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentResponse.java
```diff
@@ -0,0 +1,99 @@
...
+            boolean isFirst = true;
+            for (String value : values) {
+                if (isFirst) {
+                    response.setHeader(name, value);
+                } else {
+                    response.addHeader(name, value);
+                    isFirst = false;
+                }
+            }
+        });
```
efge, line 85: I think `stored.getHeaders().forEach(response::addHeader)` will work as well.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+        } else if (NuxeoIdempotentFilter.INPROGRESS_MARKER.equals(status)) {
+            // request already in progress -> conflict
+            // Don't call response.sendError, because it commits the response
+            // which prevents NuxeoExceptionFilter from returning a custom error page.
+            response.setStatus(HttpServletResponse.SC_CONFLICT);
+            log.debug("Conflict response for key: " + key);
+        } else {
+            // request already done: return stored result
+            response.getOutputStream().write(store.get(key));
+            NuxeoIdempotentResponse.restore(response, store.get(key + NuxeoIdempotentFilter.INFO_SUFFIX));
```
efge, line 163: You should do this first, and if there's an `IOException` (due to errors parsing the JSON) log it but abandon everything and process normally the chain.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+        String method = request.getMethod();
+        if (isIdempotentMethod(method)) {
+            log.debug("No idempotent processing done: method is already idempotent: " + method);
+            chain.doFilter(request, response);
+            return;
+        }
+        String key = request.getHeader(NuxeoIdempotentFilter.HEADER_KEY);
+        if (key == null) {
+            log.debug("No idempotent processing done: no Idempotent-Key present");
+            chain.doFilter(request, response);
```
efge, line 121: I'd prefer all the repeated `chain.doFilter(request, response)` in a caller method. So rename the current method to `doFilterIdempotent` and then have:

    @Override 
    public void doFilter(...) throws ... {
        if (!doFilterIdempotent()) {
            chain.doFilter(request, response);
        }
    }

And return `false` from `doFilterIdempotent` whenever you haven't done any processing.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+ * <li>mark the request as being processed
+ * <li>capture the response when request was processed without any error and store it
+ * <li>return the stored response if a subsequent request with the same key is processed again
+ * <li>return a conflict response if a request with the same key is processed while the first request is still in
+ * progress.
+ * </ul>
+ *
+ * @since 11.5
+ */
+public class NuxeoIdempotentFilter implements Filter {
```
efge, line 60: Extend `HttpFilter` instead, which gives you `doFilter` methods overrides with parameters already converted to `HttpServletRequest`/`HttpServletResponse`.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+        KeyValueStore store = kvs.getKeyValueStore(getStoreName());
+        String status = store.getString(key + NuxeoIdempotentFilter.INFO_SUFFIX);
+        if (status == null) {
+            log.debug("Handle new request for key: " + key);
+            long ttl = getTTLSeconds();
+            store.put(key + NuxeoIdempotentFilter.INFO_SUFFIX, NuxeoIdempotentFilter.INPROGRESS_MARKER, ttl);
+            try {
+                CopyingResponseWrapper wrapper = new CopyingResponseWrapper(DEFERRED_OUTPUT_STREAM_THRESHOLD, response);
+                chain.doFilter(request, wrapper);
+                store.put(key, wrapper.getCaptureAsBytes(), ttl);
```
efge, line 140: We should have a protection here against captures that are too big, as the k/v store is not designed to store arbitrarily big payloads.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+                store.put(key + NuxeoIdempotentFilter.INFO_SUFFIX, NuxeoIdempotentResponse.write(wrapper), ttl);
+                log.debug("Stored response for key: " + key);
```
efge, line 142: I'd like to see the response size in bytes logged as well, and the status (here and in the "return stored result" case)

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,168 @@
...
+
+    public static final String STORE_PROPERTY = "org.nuxeo.request.idempotency.keyvaluestore.name";
+
+    public static final String DEFAULT_STORE = "idempotentrequest";
+
+    protected static final long DEFAULT_TTL_SECONDS = Duration.ofDays(1).toSeconds();
+
+    public static final String TTL_SECONDS_PROPERTY = "org.nuxeo.request.idempotency.ttl.seconds";
+
+    public static final String INPROGRESS_MARKER = "IDEMPOTENCY_INPROGRESS_MARKER";
```
efge, line 74: Let's make this `{"inprogress":true}` so that it's at least valid JSON.

modules/platform/nuxeo-platform-web-common/src/test/java/org/nuxeo/ecm/platform/web/requestcontroller/filter/TestNuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,386 @@
...
+        if (status == null) {
+            assertNull(store.getString(ikey));
+        } else {
+            if (NuxeoIdempotentFilter.INPROGRESS_MARKER.equals(status)) {
+                assertEquals(status, store.getString(ikey));
+            } else {
+                String info = "{\"headers\":" //
+                        + "{\"Connection\":[\"Keep-Alive\"]," //
+                        + "\"Content-Encoding\":[\"gzip\"]," //
+                        + "\"Content-Type\":[\"text/html; charset=utf-8\"]" //
```
efge, line 228: Could you add case with a multi-value header? For example `Set-Cookie` and `Accept` are common headers seen multiple times.

efge (changes requested):

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentResponse.java
atchertchian: response does not have a addHeader(String, Collection<String>) api, but i can remove the "isFirst" and setHeader call (making sure header value is "reset") if unnecessary

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/CopyingResponseWrapper.java
```diff
@@ -0,0 +1,102 @@
...
+        if (writer != null) {
+            writer.close();
+        } else if (output != null) {
+            output.close();
+        }
+
+        if (capture.isInMemory()) {
+            return capture.getData();
+        } else {
+            return FileUtils.readFileToByteArray(capture.getFile());
```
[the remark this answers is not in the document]
atchertchian: added a call to `wrapper#flushBuffer` before calling `wrapper#getCaptureAsBytes`, still kept a call to `capture#flush` before calling `capture#getFile`, please tell if that's what it expected here.

modules/platform/nuxeo-platform-web-common/src/test/java/org/nuxeo/ecm/platform/web/requestcontroller/filter/TestNuxeoIdempotentFilter.java
atchertchian: Fixed the "isFirst" buggy logic with these additional checks :+1:

BoboTiG:
Would it be possible to return the idempotency key in the JSON as well?

modules/platform/nuxeo-platform-web-common/src/test/java/org/nuxeo/ecm/platform/web/idempotency/TestNuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,397 @@
...
+/**
+ * Checks idempotent requests management.
+ *
+ * @since 11.5
+ */
+@RunWith(FeaturesRunner.class)
+@Features(RuntimeFeature.class)
+@Deploy("org.nuxeo.runtime.kv")
+@Deploy("org.nuxeo.ecm.platform.web.common:OSGI-INF/idempotency-configuration.xml")
+public class TestNuxeoIdempotentFilter {
```
ataillefer, line 81: There are quite a few Sonar warnings about:
- Remove useless curly braces around statement.
- Remove useless curly braces around statement and then remove useless return keyword.
- Define and throw a dedicated exception instead of using a generic one.

Easy to fix if you can take the time!

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/CopyingResponseWrapper.java
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @since 11.5
+ */
+public class CopyingResponseWrapper extends HttpServletResponseWrapper implements AutoCloseable {
+
+    private final DeferredFileOutputStream capture;
+
+    private ServletOutputStream output;
+
+    private PrintWriter writer;
```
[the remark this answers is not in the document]
atchertchian: I can make that change

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,192 @@
...
+
+    protected static final Set<String> IDEMPOTENT_METHODS = Set.of(
+            // safe methods according to RFC 7231 4.2.1
+            HttpGet.METHOD_NAME, HttpHead.METHOD_NAME, HttpOptions.METHOD_NAME, HttpTrace.METHOD_NAME,
+            // idempotent methods according to RFC 7231 4.2.2
+            HttpPut.METHOD_NAME, HttpDelete.METHOD_NAME);
+
+    protected Duration getTTL() {
+        ConfigurationService cs = Framework.getService(ConfigurationService.class);
+        if (cs != null) {
```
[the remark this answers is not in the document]
atchertchian: `getService` will throw an error only if the runtime is not initialized, otherwise it could return null until the target component is activated (which can in theory happen as the servlet can be called regardless of the runtime status, especially during hot reload, but at startup too)

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,192 @@
...
+            return false;
+        }
+        String key = request.getHeader(HEADER_KEY);
+        if (key == null) {
+            log.debug("No idempotent processing done: no {} header present", HEADER_KEY);
+            return false;
+        }
+        log.debug("Idempotent request key: {}", key);
+        KeyValueService kvs = Framework.getService(KeyValueService.class);
+        if (kvs == null) {
```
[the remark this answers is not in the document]
atchertchian: In theory yes (and Florent stated earlier "You should protect against NPEs here, for the service and the store. Filters may be called very early/late and independently of Nuxeo startup itself.")

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,192 @@
...
+        } else if (INPROGRESS_MARKER.equals(status)) {
+            // request already in progress -> conflict
+            // Don't call response.sendError, because it commits the response
+            // which prevents NuxeoExceptionFilter from returning a custom error page.
+            response.setStatus(SC_CONFLICT);
+            log.debug("Conflict response for key: {}", key);
+            return true;
+        } else {
+            try {
+                // request already done: return stored result
```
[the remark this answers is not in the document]
atchertchian: Here because it was already processed: if we called the filter chain, the request would be processed again and potentially break idempotency (for Drive use case, the file would be imported again, and a new document would be created again)

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,192 @@
...
+            } finally {
+                if (response.getStatus() >= SC_BAD_REQUEST) {
+                    // error request: cleanup store
+                    store.put(key, (String) null);
+                    store.put(key + INFO_SUFFIX, (String) null);
+                    log.debug("Cleanup store: error for key: {}", key);
+                }
+            }
+        } else if (INPROGRESS_MARKER.equals(status)) {
+            // request already in progress -> conflict
```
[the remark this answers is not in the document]
atchertchian: That's the goal of idempotent processing: we want to protect against the request to be processed (here because it is already being processed)

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,192 @@
...
+            // Don't call response.sendError, because it commits the response
+            // which prevents NuxeoExceptionFilter from returning a custom error page.
+            response.setStatus(SC_CONFLICT);
+            log.debug("Conflict response for key: {}", key);
+            return true;
+        } else {
+            try {
+                // request already done: return stored result
+                NuxeoIdempotentResponse.restore(response, store.get(key + INFO_SUFFIX));
+                byte[] content = store.get(key);
```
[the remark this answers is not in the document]
atchertchian: In practice it would be very bad luck but maybe this can be made safer indeed

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -149,13 +149,15 @@ protected boolean doFilterIdempotent(HttpServletRequest request, HttpServletResp
                                 key, wrapper.getStatus(), content.length, MAX_CONTENT_SIZE);
                         return true;
                     }
+                    response.setHeader(HEADER_KEY, key);
```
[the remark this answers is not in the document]
atchertchian: I did not find any better way to handle it given the fact that response is wrapped when being processed (for response copy) and because of the pattern to return a boolean, as asked by Florent, to avoid repeating calls to `chain.doFilter` whenever the request is not handled.
Note I did set this header either when the response is too big to be stored in the KV store (maybe it should still be added ?)

modules/platform/nuxeo-platform-web-common/src/test/java/org/nuxeo/ecm/platform/web/idempotency/TestNuxeoIdempotentFilter.java
atchertchian: I fixed a few, please tell if you see remaining ones.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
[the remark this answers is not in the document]
ataillefer: OK.
> maybe it should still be added ?

I'd say yes

ataillefer (dismissed):
Good job!

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,196 @@
...
+            response.setStatus(SC_CONFLICT);
+            response.setHeader(HEADER_KEY, key);
+            log.debug("Conflict response for key: {}", key);
+            return true;
+        } else {
+            try {
+                // request already done: return stored result
+                NuxeoIdempotentResponse.restore(response, storeStatus.getBytes());
+                response.getOutputStream().write(storeContent);
+                response.setHeader(HEADER_KEY, key);
```
efge, line 185: You have to set the header before sending the body, because in most cases (unless there's buffering, or in mocked classes) sending the body "commits" the response (see `ServletResponse.isCommitted`) and you can't set any header or status after that.

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,196 @@
...
+                    }
+                    store.put(key, content, ttl);
+                    store.put(key + INFO_SUFFIX, NuxeoIdempotentResponse.write(wrapper), ttl);
+                    log.debug("Stored response for key: {} (status: {}, size in bytes: {})", key, wrapper.getStatus(),
+                            content.length);
+                    return true;
+                }
+            } catch (IOException | ServletException e) {
+                response.setHeader(HEADER_KEY, key);
+                response.setStatus(SC_INTERNAL_SERVER_ERROR);
```
efge, line 162: This should be wrapped with a `if (!response.isCommitted())`

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,196 @@
...
+            log.debug("Handle new request for key: {}", key);
+            long ttl = getTTL().toSeconds();
+            store.put(key + INFO_SUFFIX, INPROGRESS_MARKER, ttl);
+            try {
+                try (CopyingResponseWrapper wrapper = new CopyingResponseWrapper(DEFERRED_OUTPUT_STREAM_THRESHOLD,
+                        response)) {
+                    chain.doFilter(request, wrapper);
+                    wrapper.flushBuffer();
+                    byte[] content = wrapper.getCaptureAsBytes();
+                    wrapper.setHeader(HEADER_KEY, key);
```
efge, line 147: You won't be able to set the header if the response is committed, which will usually be the case (unless buffering) after the flush. So maybe set before calling `doFilter`?

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentResponse.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @since 11.5
+ */
+public class NuxeoIdempotentResponse {
+
+    protected int status;
+
+    protected Map<String, Collection<String>> headers = new LinkedHashMap<>();
+
+    protected static final ObjectMapper mapper = //
```
efge, line 48: Constants -> all-uppercase

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/idempotency/NuxeoIdempotentResponse.java
```diff
@@ -0,0 +1,100 @@
...
+                    response.setHeader(name, value);
+                    isFirst = false;
+                } else {
+                    response.addHeader(name, value);
+                }
+            }
+        });
+    }
+
+    public static final byte[] write(HttpServletResponse response) throws JsonProcessingException {
```
efge, line 93: I don't like `write` much. Maybe `save` instead?

modules/platform/nuxeo-platform-web-common/src/main/resources/OSGI-INF/idempotency-configuration.xml
```diff
@@ -0,0 +1,17 @@
...
+
+  <extension target="org.nuxeo.runtime.ConfigurationService" point="configuration">
+    <documentation>
+      Properties controlling idempotent requests TTL and KeyValue storename.
+
+      Default TTL in seconds matches 1 day.
+
+      @since 11.5
+    </documentation>
+    <property name="org.nuxeo.request.idempotency.ttl.duration">P1D</property>
```
efge, line 12: `1d` would probably be more readable.

efge (changes requested):

atchertchian:
Addressed latest feedback, rebased and squashed commits

ataillefer (dismissed):

efge (dismissed):

atchertchian:
Rebased + squashed latest changes (visible at https://github.com/nuxeo/nuxeo/pull/4545/commits/7a102ae2e348915fce3260f1dff0d6a78725428d)

modules/platform/nuxeo-platform-web-common/src/test/java/org/nuxeo/ecm/platform/web/idempotency/TestNuxeoIdempotentFilter.java
```diff
@@ -0,0 +1,406 @@
...
+    static {
+        FINAL_RESPONSE_HEADERS.putAll(RESPONSE_HEADERS);
+        FINAL_RESPONSE_HEADERS.putAll(KEY_RESPONSE_HEADERS);
+    }
+
+    protected static final Map<String, Collection<String>> FINAL_COPY_RESPONSE_HEADERS = new LinkedHashMap<>();
+
+    static {
+        FINAL_COPY_RESPONSE_HEADERS.putAll(FINAL_RESPONSE_HEADERS);
+        SKIPPED_HEADERS.forEach(FINAL_COPY_RESPONSE_HEADERS::remove);
```
efge, line 115: `FINAL_COPY_RESPONSE_HEADERS.keySet().removeAll(SKIPPED_HEADERS)` instead for clarity?

efge (changes requested):

efge (dismissed):

atchertchian:
added ftest at ITNuxeoIdempotentRequestTest

efge (approved):

ftests/nuxeo-server-tests/src/test/java/org/nuxeo/ftest/server/ITNuxeoIdempotentRequestTest.java
```diff
@@ -0,0 +1,128 @@
...
+    protected String createDocument(NuxeoClient client) {
+        Document document = Document.createWithName(TEST_TITLE, TEST_TYPE);
+        document.setProperties(Map.of(DublinCoreConstants.DUBLINCORE_TITLE_PROPERTY, TEST_TITLE));
+        Document created = client.repository().createDocumentByPath(PARENT_PATH, document);
+        waitForAsyncWork();
+        return created.getId();
+    }
+
+    /**
+     * Prevents from random failures when counting children.
```
ataillefer, line 72: How come? There should be no async jobs involved here, and I guess the transaction should always have been committed when getting  the answer from `createDocument` @kevinleturc?

atchertchian: The rest query API (used to count children) relies on the page provider REST_API_SEARCH_ADAPTER, using ElasticSearch and not the CoreSession, hence the need for this to avoid random failures.

Note i needed this trick in Explorer ftests too, so i think this whole method would be useful on the RestHelper API.

atchertchian: And just to answer the fact that this would be needed to get the document id after creation: it's not.
I'm just following a test pattern where the async jobs should be waited for after the action triggering them, as shutting down the test server while async jobs are still running is also exposing tests to random failures, with errors in logs, typically).

ataillefer (approved):

PR #4552, opened 2020-12-22
Feature NXP-29991 better zip preview

kevinleturc:


modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Previewer for Zip blob.
+ * <p>
+ * It sanitizes each zip entry if needed.
+ *
+ * @since 11.5
+ */
+public class ZipPreviewer implements MimeTypePreviewer {
+
+    protected static final Set<String> HTML_MIME_TYPES = Set.of("text/html", "text/xml", "text/plain");
```
efge, line 55: You could use `javax.ws.rs.core.MediaType.TEXT_HTML` etc. for the constants (if it's not a dependency to add)

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private static final Logger log = LogManager.getLogger(ZipPreviewer.class);
```
efge, line 57: Please put the `log` as the first constant.

modules/platform/nuxeo-preview-core/src/test/java/org/nuxeo/ecm/platform/preview/tests/adapter/TestPreviewAdapter.java
```diff
@@ -103,16 +110,53 @@ public void testXMLDocument() throws Exception {
...
+    @Test
+    @Deploy("org.nuxeo.ecm.platform.preview:test-zip-preview-contrib.xml")
+    public void testZIPDocument() throws Exception {
+        File file = Framework.createTempFile("testZIPDocument", ".zip");
+        try (ZipOutputStream out = new ZipOutputStream(new FileOutputStream(file))) {
+            writeZipEntry(out, "payload.html");
+            writeZipEntry(out, "payload.xml");
+            writeZipEntry(out, "payload.txt");
+        }
+        Blob blob = Blobs.createBlob(file, "application/zip", "UTF-8", "testZIPDocument.zip");
```
efge, line 122: `"UTF-8"` -> `null`, it's the default and doesn't need to be explicit

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,131 @@
...
+        String filename = blob.getFilename();
+        if (!isSanitizable(blob)) {
+            log.debug("ZipEntryBlob: {} is not sanitizable", filename);
+            return blob;
+        }
+        try {
+            String content = blob.getString();
+            content = getHtmlSanitizerService().sanitizeString(content, null);
+            content = PreviewHelper.makeHtmlPage(content);
+            return Blobs.createBlob(content, "text/html", "UTF-8", filename);
```
efge, line 87: `"UTF-8"` -> `null`

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,131 @@
...
+        } catch (IOException e) {
+            throw new PreviewException("Cannot fetch ZipEntryBlob content with filename: " + filename, e);
```
efge, line 89: Is that the right error message?

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    protected boolean isSanitizable(Blob blob) {
+        String mimeType = getMimetypeRegistry().getMimetypeFromFilenameWithBlobMimetypeFallback(blob.getFilename(), blob, null);
+        boolean sanitized = false;
+        if (mimeType == null) {
+            try (InputStream stream = blob.getStream()) {
+                // last chance introspect the content
+                LineIterator lineIt = IOUtils.lineIterator(stream, defaultIfNull(blob.getEncoding(), "UTF-8"));
+                while (lineIt.hasNext()) {
```
efge, line 100: Isn't there a risk that we'll read all the "lines" of a binary file for example?

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,131 @@
...
+                    String line = lineIt.nextLine();
+                    sanitized = line.contains("<script");
+                    if (sanitized) {
+                        break;
+                    }
```
efge, line 105: You could return directly here, no need for a local `sanitized` variable (which I'd have called `sanitizable`)

efge (changes requested):

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
kevinleturc: Yes it is possible if we don't succeed to detect the mime type. This piece of code is there to handle xml files, because the mime type registry doesn't recognize them due to its ambiguous definition.
Do you see another way to do this?

efge: I don't see another way if we can't detect the type, we have to read it manually.

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
kevinleturc: Yes, IOException is only thrown by `blob.getString()`.

efge (approved):

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,133 @@
...
+import org.nuxeo.ecm.core.convert.api.ConversionException;
+import org.nuxeo.ecm.core.convert.api.ConversionService;
+import org.nuxeo.ecm.platform.htmlsanitizer.HtmlSanitizerService;
+import org.nuxeo.ecm.platform.mimetype.interfaces.MimetypeRegistry;
+import org.nuxeo.ecm.platform.preview.api.PreviewException;
+import org.nuxeo.ecm.platform.preview.helper.PreviewHelper;
+import org.nuxeo.runtime.api.Framework;
+
+/**
+ * Previewer for Zip blob.
```
ataillefer, line 50: Zip blobs

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final Logger log = LogManager.getLogger(ZipPreviewer.class);
+
+    protected static final Set<String> HTML_MIME_TYPES = Set.of(TEXT_HTML, TEXT_XML, TEXT_PLAIN);
+
+    @Override
+    public List<Blob> getPreview(Blob blob, DocumentModel dm) throws PreviewException {
+        ConversionService conversionService = Framework.getService(ConversionService.class);
+        String converterName = conversionService.getConverterName("application/zip", "text/html");
+        if (converterName == null) {
+            throw new PreviewException("Unable to find zip2html converter");
```
ataillefer, line 67: maybe a more generic message like "Unable to find converter from application/zip to text/html"

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        try {
+            BlobHolder result = conversionService.convert(converterName, new SimpleBlobHolder(blob), null);
+            List<Blob> blobs = new ArrayList<>(result.getBlobs());
+            blobs.replaceAll(this::sanitize);
+            return blobs;
```
ataillefer, line 74: Any point using streams here?

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,133 @@
...
+                throw new PreviewException("Unable to introspect content");
+            }
+        } else {
+            return HTML_MIME_TYPES.contains(mimeType);
+        }
+    }
+
+    protected MimetypeRegistry getMimetypeRegistry() {
+        MimetypeRegistry registry = Framework.getService(MimetypeRegistry.class);
+        if (registry == null) {
```
ataillefer, line 120: can this happen?

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
```diff
@@ -0,0 +1,133 @@
...
+            throw new PreviewException("Cannot find MimetypeRegistry");
+        }
+        return registry;
+    }
+
+    protected HtmlSanitizerService getHtmlSanitizerService() {
+        HtmlSanitizerService sanitizer = Framework.getService(HtmlSanitizerService.class);
```
ataillefer, line 127: same question

ataillefer (approved):

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
kevinleturc: Yes, we save some lines, I'll change it to Stream.

modules/platform/nuxeo-preview-core/src/main/java/org/nuxeo/ecm/platform/preview/adapter/ZipPreviewer.java
kevinleturc: Probably not. I will remove them.

ataillefer (approved):

efge (approved):

PR #4553, opened 2020-12-23
NXP-29996: update the blob when updating the blobHolder

troger:


kevinleturc (approved):

efge:
The commit message is too technical about what is changed. Could you update it instead to explain the goal/effect of the change? Like "fix conversion service to ..."?

efge (approved):

ataillefer (approved):

kevinleturc (approved):

PR #4555, opened 2020-12-23
Task nxp 27712 rest api test status check

ataillefer:


ci/Jenkinsfiles/build.groovy
```diff
@@ -760,6 +760,30 @@ pipeline {
...
+        ----------------------------------------
+        """
+        script {
+          def parameters = [
+            string(name: 'NUXEO_REPOSITORY', value: "${repositoryUrl}"),
+            string(name: 'NUXEO_SHA', value: "${GIT_COMMIT}"),
+          ]
+          if (isPullRequest()) {
+            parameters.add(string(name: 'NUXEO_VERSION', value: "${VERSION}"))
+          }
```
kevinleturc, line 777: Shouldn't we put every time the NUXEO_VERSION? This will avoid potential issues regarding the alias and imagePullPolicy.

kevinleturc (approved):

ci/Jenkinsfiles/build.groovy
ataillefer: Actually, I'm relying on the fact of passing this parameter or not to detect if the upstream (nuxeo/nuxeo) is a PR or master. Could parse the upstream job full name but I thought the current approach was simpler.

In any case, you're right about the `imagePullPolicy` needing to be `Always` in the nuxeo chart used by the REST API tests pipeline, since the default image is `11.x` which is a rolling tag.

troger (approved):

PR #4559, opened 2020-12-24
NXP-30010: improve concurrent updates removing elements from lists on MongoDB

efge:


modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/TestSQLRepositoryProperties.java
```diff
@@ -378,6 +378,61 @@ protected void addValueToDoc(String value, CyclicBarrier barrier, List<Exception
...
+                // wait for both threads to be at the same point
+                try {
+                    barrier.await(5, TimeUnit.SECONDS);
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                    throw new NuxeoException(e);
+                } catch (BrokenBarrierException | TimeoutException e) {
+                    throw new NuxeoException(e);
+                }
+                // add the value to the existing list
```
troger, line 421:
```suggestion
                // remove the value from the existing list
```

PR #4580, opened 2021-01-08
NXP-29983: Add property name on constraint violation error message

charlesboidot:


modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/DocumentValidationException.java
```diff
@@ -61,11 +61,17 @@ public DocumentValidationReport getReport() {
     public String getMessage() {
         if (report.hasError()) {
             int num = report.numberOfErrors();
+            ValidationViolation violation = report.asList().get(0);
+            String property = "";
+            if (violation instanceof ConstraintViolation) {
+                ConstraintViolation cv = (ConstraintViolation) violation;
+                property = cv.getPath().get(0).toString();
```
efge, line 68: `getPath().get(0)` just returns the first part of the property xpath, but for a complex value it would be useful to have the full path. So all parts would be joined by `/`. Note that care must be taken for list elements (see logic in `PathNode.toString()`). All this logic could probably be put in a new static method `PathNode.toString(List<PathNode>)` that returns a `String`.
While doing this please rename `property` to `xpath` as that's what it is now.

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/api/validation/TestDocumentValidationService.java
```diff
@@ -621,8 +629,16 @@ public void testDocumentValidationReportMessage() {
...
+        if (violation instanceof ConstraintViolation) {
+            cv = (ConstraintViolation) violation;
+        }
+
         e = new DocumentValidationException(report);
-        expectedMessage = String.format(DocumentValidationException.MESSAGE_SINGLE, MESSAGE_FOR_USERS_FIRSTNAME);
+
+        //changed in 11.5 since the message includes the property name (NXP-29983)
+        expectedMessage = String.format(DocumentValidationException.MESSAGE_SINGLE, cv.getPath().get(0),
+                MESSAGE_FOR_USERS_FIRSTNAME);
```
efge, line 641: In expected message it's better to have the explicit value as a string constant instead of `cv.getPath().get(0)` that hides what we really are expecting.

efge (changes requested):

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/api/validation/TestDocumentValidationService.java
efge: So here really we want a full string constant for the entire expected message, no reuse of `DocumentValidationException.MESSAGE_SINGLE` (which means the previous code was bad).

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/DocumentValidationException.java
efge: Can you check if there's a unit test that will show this full xpath in the error message as well?

charlesboidot: There is not any unit test that show the XPath which is understandable as the XPath wasn't a part of the error message before.

efge: Ok so there's probably a need for one to be sure the feature works as we expect.

charlesboidot: The following test was updated to test the XPath.

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/api/validation/TestDocumentValidationService.java
```diff
@@ -610,9 +616,16 @@ public void testDocumentValidationReportMessage() {
         String expectedMessage = String.join("\n", MESSAGE_FOR_GROUP_CODE, MESSAGE_FOR_USERS_FIRSTNAME);
         assertEquals(expectedMessage, report.toString());
 
+        ValidationViolation violation = report.asList().get(0);
+        ConstraintViolation cv = null;
+        if (violation instanceof ConstraintViolation) {
+            cv = (ConstraintViolation) violation;
+        }
+
```
efge, line 624: This isn't needed anymore (same further down)

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/api/validation/TestDocumentValidationService.java
```diff
@@ -610,9 +616,16 @@ public void testDocumentValidationReportMessage() {
...
         DocumentValidationException e = new DocumentValidationException(report);
-        expectedMessage = String.format(DocumentValidationException.MESSAGE, report.numberOfErrors(),
-                MESSAGE_FOR_GROUP_CODE);
+
+        //changed in 11.5 since the message includes the property name (NXP-29983)
```
efge, line 627: No need for this kind of comment, git history is enough

efge (changes requested):

efge (dismissed):

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/ConstraintViolation.java
```diff
@@ -96,6 +96,17 @@ public Schema getSchema() {
...
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < path.size(); i++) {
+            sb.append(path.get(i).toString());
+            if (i != path.size() - 1) {
+                sb.append("/");
+            }
+        }
+        return sb.toString();
+    }
+
```
troger, line 109: Couldn't it be simplified to something like

```suggestion
    public String getPathAsString() {
        path.stream().map(PathNode::toString).collect(Collectors.joining("/"));
    }
```
?

troger (approved):

efge (approved):

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/validation/DocumentValidationException.java
```diff
@@ -61,11 +61,17 @@ public DocumentValidationReport getReport() {
     public String getMessage() {
         if (report.hasError()) {
             int num = report.numberOfErrors();
-            String violationMessage = report.asList().get(0).getMessage(null);
+            ValidationViolation violation = report.asList().get(0);
+            String xpath = "";
+            if (violation instanceof ConstraintViolation) {
+                ConstraintViolation cv = (ConstraintViolation) violation;
+                xpath = cv.getPathAsString();
```
atchertchian, line 68: Hi, after this change I get an error message on a custom validation (in explorer code) which is not ideal (see https://github.com/nuxeo/nuxeo-explorer/blob/master/modules/nuxeo-apidoc-core/src/main/java/org/nuxeo/apidoc/repository/RepositoryDistributionSnapshot.java#L602).
The new error message is: `Constraint violation thrown on property : 'Distribution key or alias is reserved: 'current''`, see https://jenkins.platform.dev.nuxeo.com/job/nuxeo/job/nuxeo-explorer/job/master/237/testReport/junit/org.nuxeo.apidoc.test/TestSnapshotPersist/testUpdatePersisted/

This is a detail but can we use the old message when xpath is empty, for that kind of use cases? Note I could work around this by defining a constraint violation on my end to give a xpath here, but maybe we should also handle cases where the ValidationViolation is an instance of GlobalViolation, and not of ConstraintViolation, as expected here.

charlesboidot: Taken into account, created PR https://github.com/nuxeo/nuxeo/pull/4586 to fix this

atchertchian: Thanks :+1:

PR #4588, opened 2021-01-13
NXP-29984 Runtime Registries Migrations #2

atchertchian:


modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobManagerComponent.java
```diff
@@ -62,89 +55,36 @@
...
-        @Override
-        public boolean isSupportingMerge() {
-            return true;
-        }
-
-        public void clear() {
-            currentContribs.clear();
-        }
+    // volatile for double-checked locking
+    protected volatile boolean blobProvidersInitialized;
```
efge, line 59: I'm not seeing any use of this field for double-checked locking.

modules/core/nuxeo-core-el/src/test/java/org/nuxeo/platform/el/TestELService.java
```diff
@@ -26,28 +27,42 @@
...
 import org.nuxeo.ecm.platform.el.ELService;
+import org.nuxeo.ecm.platform.el.ExpressionContext;
 import org.nuxeo.platform.el.DummyELContextFactory.DummyELContext;
 import org.nuxeo.runtime.api.Framework;
 import org.nuxeo.runtime.test.runner.Deploy;
 import org.nuxeo.runtime.test.runner.Features;
 import org.nuxeo.runtime.test.runner.FeaturesRunner;
 import org.nuxeo.runtime.test.runner.RuntimeFeature;
 
+import com.google.inject.Inject;
```
efge, line 38: Use `javax` not `com.google`

modules/core/nuxeo-core-convert/src/main/java/org/nuxeo/ecm/core/convert/extension/GlobalConfigDescriptor.java
```diff
@@ -21,38 +21,41 @@
...
     public static final String DEFAULT_CACHING_DIRECTORY = "convertcache";
 
-    @XNode("enableCache")
-    protected Boolean enableCache;
-
-    public boolean isCacheEnabled() {
-        return enableCache == null ? DEFAULT_CACHE_ENABLED : enableCache.booleanValue();
-    }
+    @XNode(value = "enableCache")
+    protected boolean enableCache = true;
```
efge, line 44: Shouldn't this `true` be some kind of "defaultValue" (I don't remember the exact name)?

modules/core/nuxeo-core-cache/src/main/java/org/nuxeo/ecm/core/cache/CacheServiceImpl.java
```diff
@@ -189,7 +183,7 @@ public int getApplicationStartedOrder() {
     public void start(ComponentContext context) {
         super.start(context);
         ClusterService clusterService = Framework.getService(ClusterService.class);
-        if (clusterService.isEnabled()) {
+        if (clusterService != null && clusterService.isEnabled()) {
```
efge, line 186: If `clusterService` is `null` then we have a problem because the point of these caches is to have cluster-wide invalidations... Except maybe in unit tests but I'd rather have the normal code changed just for unit tests that don't initialize all they should

modules/core/nuxeo-core-cache/src/main/java/org/nuxeo/ecm/core/cache/CacheServiceImpl.java
```diff
@@ -264,8 +262,11 @@ public Cache getCache(String name) {
     /**
      * @since 9.3
      */
-    public CacheDescriptor getCacheDescriptor(String descriptor) {
-        return getDescriptor(XP_CACHES, descriptor);
+    public CacheDescriptor getCacheDescriptor(String name) {
+        if (programmaticCaches.containsKey(name)) {
+            return programmaticCaches.get(name);
+        }
```
efge, line 268: CacheDescriptor descriptor = programmaticCaches.get(name);
        if (descriptor != null) {
              return descriptor;
          }

modules/core/nuxeo-core-cache/src/test/java/org/nuxeo/ecm/core/cache/TestCacheDescriptor.java
```diff
@@ -21,83 +21,90 @@
...
 import org.junit.runner.RunWith;
-import org.nuxeo.runtime.model.DescriptorRegistry;
+import org.nuxeo.runtime.api.Framework;
+import org.nuxeo.runtime.test.runner.Deploy;
 import org.nuxeo.runtime.test.runner.Features;
 import org.nuxeo.runtime.test.runner.FeaturesRunner;
+import org.nuxeo.runtime.test.runner.HotDeployer;
 import org.nuxeo.runtime.test.runner.RuntimeFeature;
 
+import com.google.inject.Inject;
```
efge, line 39: `javax`

modules/core/nuxeo-core-cache/src/test/java/org/nuxeo/ecm/core/cache/TestCacheDescriptor.java
```diff
@@ -21,83 +21,90 @@
...
-    private static final String COMPONENT = "component";
+    protected static final String NAME = "myid";
 
-    private static final String XP = "extension-point";
+    @Inject
+    protected HotDeployer hotDeployer;
 
-    protected static final String NAME = "myid";
+    protected CacheServiceImpl getService() {
+        return (CacheServiceImpl) Framework.getService(CacheService.class);
```
efge, line 52: You could still inject it though.

modules/runtime/nuxeo-runtime-datasource/src/main/java/org/nuxeo/runtime/datasource/DataSourceComponent.java
```diff
@@ -18,146 +18,55 @@
...
-
-    protected void removeDataSource(DataSourceDescriptor contrib) {
-        unbindDataSource(contrib);
-        datasources.remove(contrib.getName());
+        poolRegistry = null;
     }
 
     protected void bindDataSource(DataSourceDescriptor descr) {
-        log.info("Registering datasource: " + descr.getName());
+        log.info("Registering datasource: {}", descr.getName());
```
efge, line 64: `descr::getName`

efge (changes requested):

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobManagerComponent.java
atchertchian: Actually it's used in a synchronized method, so i can just remove "volatile" ?

modules/core/nuxeo-core-convert/src/main/java/org/nuxeo/ecm/core/convert/extension/GlobalConfigDescriptor.java
atchertchian: defaultAssignment. I've kept this here as this descriptor is instantiated by the service when there is no configuration provided through XML.

modules/core/nuxeo-core-cache/src/main/java/org/nuxeo/ecm/core/cache/CacheServiceImpl.java
atchertchian: It was indeed in unit tests -- can't remember which right now but will lookup the stack trace in next validation tests output.

atchertchian: Actually TestCacheDescriptor is the one running into this issue: will adapt the test features.

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/BlobManagerComponent.java
efge: You can remove `volatile` but _all_ uses must be synchronized, even when you just set back to `false`:

    synchronized (this) {
        blobProvidersInitialized = false;
    }

Or keep `volatile` and use actual double-checked locking in `getBlobProviders`.

atchertchian: :+1: (removed volatile)

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/CoreService.java
```diff
@@ -75,75 +75,57 @@ public void deactivate(ComponentContext context) {
...
+    public void start(ComponentContext context) {
+        versionRemovalPolicy = this.<CoreServicePolicyDescriptor> getRegistryContribution(VERSION_REMOVAL_POLICY_XP)
+                                   .map(this::createVersionRemovalPolicy)
+                                   .orElse(null);
+        orphanVersionRemovalFilters.addAll(
+                this.<CoreServiceOrphanVersionRemovalFilterDescriptor> getRegistryContributions(
+                        ORPHAN_VERSION_REMOVAL_FILTER_XP)
+                    .stream()
+                    .map(this::createOrphanVersionRemovalFilter)
+                    .collect(Collectors.toList()));
```
kevinleturc, line 87: What do you think about assigning the result to `orphanVersionRemovalFilters` and nullifying it in the `stop` method?

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/CoreService.java
```diff
@@ -75,75 +75,57 @@ public void deactivate(ComponentContext context) {
...
-            registerVersionRemovalPolicy((CoreServicePolicyDescriptor) contrib);
-        } else if (ORPHAN_VERSION_REMOVAL_FILTER_XP.equals(point)) {
-            registerOrphanVersionRemovalFilter((CoreServiceOrphanVersionRemovalFilterDescriptor) contrib);
-        } else {
-            throw new RuntimeException("Unknown extension point: " + point);
-        }
+    public void start(ComponentContext context) {
+        versionRemovalPolicy = this.<CoreServicePolicyDescriptor> getRegistryContribution(VERSION_REMOVAL_POLICY_XP)
+                                   .map(this::createVersionRemovalPolicy)
+                                   .orElse(null);
```
kevinleturc, line 81: What do you think about giving `DEFAULT_VERSION_REMOVAL_POLICY` to `orElse` and removing the ternary in `getVersionRemovalPolicy`?

modules/platform/nuxeo-platform-importer/nuxeo-importer-scan-xml-parser/src/test/java/org/nuxeo/ecm/platform/importer/xml/parser/TestDeployment.java
```diff
@@ -24,30 +24,30 @@
...
 import org.junit.runner.RunWith;
 import org.nuxeo.ecm.core.test.CoreFeature;
-import org.nuxeo.ecm.platform.importer.service.DefaultImporterService;
-import org.nuxeo.runtime.api.Framework;
+import org.nuxeo.ecm.platform.scanimporter.service.ScannedFileMapperService;
 import org.nuxeo.runtime.test.runner.Deploy;
 import org.nuxeo.runtime.test.runner.Features;
 import org.nuxeo.runtime.test.runner.FeaturesRunner;
 
+import com.google.inject.Inject;
```
kevinleturc, line 32: Could you replace it by `javax.inject.Inject`?

modules/platform/nuxeo-platform-usermanager/src/main/java/org/nuxeo/ecm/platform/computedgroups/ComputedGroupsServiceImpl.java
```diff
@@ -50,65 +50,39 @@
...
-        super.activate(context);
-        computers.clear();
-        computerNames.clear();
+    public void start(ComponentContext context) {
+        computers.putAll(this.<GroupComputerDescriptor> getRegistryContributions(COMPUTER_EP)
+                             .stream()
+                             .collect(Collectors.toConcurrentMap(GroupComputerDescriptor::getName,
+                                     GroupComputerDescriptor::getComputer)));
+        this.<GroupComputerChainDescriptor> getRegistryContribution(CHAIN_EP)
+            .ifPresent(gcc -> computerNames.addAll(gcc.getComputerNames()));
```
kevinleturc, line 66: What about assignation/nullifying (same question as above)?

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/CoreService.java
atchertchian: I've been using that pattern to avoid having API calls that would be done before component start to end with a NPE, but maybe i should assume no call should be done until the component is started anyway?

atchertchian: Done changes in separate commits to ease up re-review (will fixup commits after validation)

efge (approved):

kevinleturc (dismissed):

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/api/blobholder/TestBlobHolderAdapterService.java
```diff
@@ -45,25 +45,28 @@
...
     public void testContrib() throws Exception {
-        BlobHolderAdapterComponent bhac = (BlobHolderAdapterComponent) service;
+        BlobHolderAdapterComponent bhac = (BlobHolderAdapterComponent) getService();
 
         assertEquals(0, bhac.getFactoryNames().size());
 
         hotDeployer.deploy(Constants.CORE_TEST_TESTS_BUNDLE + ":test-blob-holder-adapters-contrib.xml");
 
+        // retrieve service again after hotreload
+        bhac = (BlobHolderAdapterComponent) getService();
```
efge, line 69: Or you could keep all the injection code and just do `service = Framework.getService(BlobHolderAdapterService.class)` here (keeping the comment)

efge (approved):

kevinleturc (approved):

PR #4591, opened 2021-01-14
NXP-30050: write generic message for exception mapped to HTTP error code >= 500

troger:


modules/platform/nuxeo-webengine/nuxeo-webengine-core/src/main/java/org/nuxeo/ecm/webengine/app/JsonWebengineWriter.java
```diff
@@ -102,6 +104,19 @@ protected static void writeException(JsonGenerator jg, Throwable t, MediaType me
...
+     * @since 11.5
+     */
+    protected static String getExceptionMessage(String exceptionMessage, int statusCode) {
+        if (statusCode < SC_INTERNAL_SERVER_ERROR) {
+            return exceptionMessage;
+        }
+
+        boolean stackDisplay = jsonFactoryManager.isStackDisplay();
+        boolean isDevMode = Framework.isDevModeSet();
+        return stackDisplay || isDevMode ? exceptionMessage : "Internal Server Error";
```
efge, line 117: Why not simply

    if (statusCode < SC_INTERNAL_SERVER_ERROR || jsonFactoryManager.isStackDisplay() || Framework.isDevModeSet()) {
        return exceptionMessage;
    } else {
        return "Internal Server Error";
    }

efge (dismissed):

efge (approved):

modules/platform/nuxeo-webengine/nuxeo-webengine-core/src/main/java/org/nuxeo/ecm/webengine/app/JsonWebengineWriter.java
```diff
@@ -102,6 +104,17 @@ protected static void writeException(JsonGenerator jg, Throwable t, MediaType me
         jg.flush();
     }
 
+    /**
+     * @since 11.5
+     */
+    protected static String getExceptionMessage(String exceptionMessage, int statusCode) {
+        if (statusCode < SC_INTERNAL_SERVER_ERROR || jsonFactoryManager.isStackDisplay() || Framework.isDevModeSet()) {
```
BoboTiG, line 111: Maybe a stupid question: what about 502, 503, 504 errors? They wil have the same `"Internal Server Error"` message?

BoboTiG: Well, it is the desired behavior. :)

kevinleturc (approved):

modules/platform/nuxeo-webengine/nuxeo-webengine-core/src/main/java/org/nuxeo/ecm/webengine/app/JsonWebengineWriter.java
troger: So, no question? 😬

efge: Indeed we could try to keep better specific messages for >500 but the http code is what's important to clients. And there's very few places in Nuxeo where we return something else than 500.

efge (approved):

kevinleturc (approved):

PR #4593, opened 2021-01-18
NXP-29984: Runtime Registries Migrations #3

atchertchian:

PR #4601, opened 2021-01-22
NXP-29984: Runtime Registries Migrations #4

atchertchian:


modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/XValueFactory.java
```diff
@@ -196,6 +197,9 @@ public String serialize(Context context, Object value) {
     public static final XValueFactory CLASS = new XValueFactory() {
         @Override
         public Object deserialize(Context context, String value) {
+            if (StringUtils.isBlank(value)) {
```
atchertchian, line 200: This change was needed to allow disabling a query maker without redefining the class: see https://jira.nuxeo.com/browse/NXP-30092 for proper handling of such cases, with addition of unit tests.

atchertchian:
(force-pushed after merge of previous PR #4593 on master)

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/thumbnail/ThumbnailServiceImpl.java
```diff
@@ -41,38 +38,39 @@
...
 public class ThumbnailServiceImpl extends DefaultComponent implements ThumbnailService {
 
-    private static final Log log = LogFactory.getLog(BlobHolderAdapterComponent.class);
-
     public static final String THUMBNAILFACTORY_EP = "thumbnailFactory";
 
     protected ThumbnailFactory defaultFactory;
 
-    protected final Map<String, ThumbnailFactory> factoriesByDocType = new HashMap<>();
+    protected Map<String, ThumbnailFactory> factoriesByDocType = new HashMap<>();
```
efge, line 45: Can be `null` given that you set them in `start`

modules/core/nuxeo-core-api/src/test/java/org/nuxeo/ecm/core/api/pathsegment/PathSegmentServiceTest.java
```diff
@@ -91,7 +92,7 @@ public void testContrib() throws Exception {
     @Test
     @Deploy("org.nuxeo.ecm.core.api.tests:OSGI-INF/test-pathsegment-contrib.xml")
     public void testContribOverride() throws Exception {
-        PathSegmentService service = Framework.getService(PathSegmentService.class);
+        assertNotNull(service);
```
efge, line 95: Not needed, test framework will throw if it cannot inject. Same in other methods.

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/pathsegment/PathSegmentComponent.java
```diff
@@ -18,135 +18,53 @@
...
+    public void start(ComponentContext context) {
+        PathSegmentService confService = this.<PathSegmentServiceDescriptor> getRegistryContribution(XP).map(desc -> {
+            try {
+                return desc.className.getDeclaredConstructor().newInstance();
+            } catch (ReflectiveOperationException e) {
+                throw new RuntimeServiceException(e);
+            }
+        }).orElse(null);
+        if (confService == null) {
+            confService = new PathSegmentServiceDefault();
```
efge, line 45: Coud be predefined in a constant, and in the `orElse`

modules/core/nuxeo-core-persistence/src/main/java/org/nuxeo/ecm/core/persistence/PersistenceComponent.java
```diff
@@ -100,7 +60,7 @@ public PersistenceProvider newProvider(String name) {
 
     @Override
     public HibernateConfiguration getHibernateConfiguration(String name) {
-        HibernateConfiguration config = registry.get(name);
+        HibernateConfiguration config = this.<HibernateConfiguration> getRegistryContribution(XP, name).orElse(null);
         if (config == null) {
             throw new PersistenceError("no hibernate configuration identified by '" + name + "' is available");
```
efge, line 65: `orElseThrow`?

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/jdbc/QueryMakerServiceImpl.java
```diff
@@ -20,88 +20,43 @@
...
-        }
-    }
-
-    @Override
-    public void unregisterContribution(Object contribution, String extensionPoint, ComponentInstance contributor) {
-        if (XP.equals(extensionPoint)) {
-            unregisterQueryMaker((QueryMakerDescriptor) contribution);
+    public void start(ComponentContext context) {
+        queryMakers = new ArrayList<>(2);
+        queryMakers.addAll(this.<QueryMakerDescriptor> getRegistryContributions(XP)
```
efge, line 42: Could probably assign directly. `Collectors.toList` in practice returns an `ArrayList` and this won't change (as too many people already rely on that, per OpenJDK dev discussions).

modules/core/nuxeo-core-storage/src/main/java/org/nuxeo/ecm/core/storage/lock/LockManagerService.java
```diff
@@ -41,109 +38,48 @@
...
+    public void start(ComponentContext context) {
+        lockManagers = new ConcurrentHashMap<>();
+        this.<LockManagerDescriptor> getRegistryContributions(XP_LOCKMANAGER).forEach(desc -> {
+            if (desc.klass != null) {
+                try {
+                    Constructor<? extends LockManager> ctor = desc.klass.getConstructor(String.class);
+                    lockManagers.put(desc.name, ctor.newInstance(name));
+                } catch (ReflectiveOperationException e) {
+                    addRuntimeMessage(Level.ERROR, e.getMessage());
+                    throw new NuxeoException(e);
```
efge, line 63: `RuntimeServiceException` for consistency?

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/pathsegment/PathSegmentComponent.java
atchertchian: Sadly using orElse directly does not work here:
```
The method orElse(capture#1-of ? extends PathSegmentService) in the type Optional<capture#1-of ? extends PathSegmentService> is not applicable for the arguments (PathSegmentService)
```

I'll change the assignment still

atchertchian:
(also reviewed MultiTenantServiceImpl, tests showed that applicationStarted deprecated method was not called anymore, due to the start method impl, now fixed)

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/pathsegment/PathSegmentComponent.java
efge: With an additional cast this seems to compile;

        service = this.<PathSegmentServiceDescriptor> getRegistryContribution(XP).map(desc -> {
            try {
                return (PathSegmentService) desc.className.getDeclaredConstructor().newInstance();
            } catch (ReflectiveOperationException e) {
                throw new RuntimeServiceException(e);
            }
        }).orElse(DEFAULT_SERVICE);

And all the lambda could be in a helper `newInstance(PathSegmentServiceDescriptor)` method to make this nicer.

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/test/resources/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,3 @@
+Manifest-Version: 1.0
+Bundle-Name: org.nuxeo.ecm.core.storage.sql.test
```
atchertchian, line 2: this is making eclipse confuse this bundle and the nuxeo-core-storage-sql-test bundle: will rename it (see https://jira.nuxeo.com/browse/NXP-30098)

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/pathsegment/PathSegmentComponent.java
atchertchian: I'll keep that in mind when working on https://jira.nuxeo.com/browse/NXP-30092: there are too many places where we do such instantiations (and i'd still like to throw better error messages in cases where the className would be null, which is not the case here).
The related error management should be consistent in all cases (some add runtime error messages, some not, etc...)

kevinleturc: Just for my knowledge, in such a situation, we prefer to declare the default service as a static member VS doing a `orElseGet(PathSegmentService::new)` for instance?

atchertchian: Not necessarily, at least that's not what I understood, it's just syntax-related.
Here i do agree we'd better instantiate at start, it even if it should not make a big difference (as the default class does not hold any state).
It's a bit different when this default value is non-null after the component is stopped (in ExceptionHandler component typically), where using a static member makes more sense to me, or when the instantiated class holds a state (in which case re-instantiating makes more sense).

efge: Agreed that here it does not make a big difference.

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/MapRegistry.java
```diff
@@ -129,4 +129,12 @@ protected String computeId(Context ctx, XAnnotatedObject xObject, Element elemen
         return (T) contrib;
     }
 
+    protected void logError(String message, Throwable t, String extensionId, String id) {
+        log.error(message, t);
+    }
+
+    protected void logWarn(String message, String extensionId, String id) {
+        log.warn(message);
+    }
```
kevinleturc, line 138: Logging shouldn't exist in methods.

Also, this increases the complexity as we're now having two more interfaces to provide default methods + two more classes to link the abstract and the interface.

I would prefer to use a decorator pattern for this.

atchertchian: I don't understand the "shouldn't exist" part. Please tell if you have another implementation in mind: i started defining a "RuntimeLogger" class and thought it was overkill after all.
I can remove the whole feature for now if you prefer: the point is to easily check that we have no implicit merges by default, but it's also to prepare for consistent exception management on classes instantiation (https://jira.nuxeo.com/browse/NXP-30092)

kevinleturc: I wanted to say that method referring to logging shouldn't exist in code.

I really think a decorator class would be more elegant.

After a Slack talk, we agree to postpone this.

kevinleturc (approved):

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/MapRegistry.java
atchertchian: Commit moved to the wip branch for NXP-30092

efge (approved):

PR #4603, opened 2021-01-24
Feature NXP-29920 dynamic template property

kevinleturc:
I removed all deprecated methods, fields, and constants that I encountered. 
Don't hesitate if you think it makes sense to keep some of them.


modules/runtime/nuxeo-connect-standalone/src/main/java/org/nuxeo/connect/update/task/guards/PlatformHelper.java
```diff
@@ -81,7 +79,7 @@ public boolean isJetty() {
      */
     @Deprecated
     public boolean isWindows() {
-        return System.getProperty("os.name").toLowerCase().indexOf("win") >= 0;
+        return System.getProperty("os.name").toLowerCase().contains("win");
```
efge, line 81: Unrelated to your changes but `toLowerCase` without a locale is wrong. This will fail in a Turkish default locale. I'd just do `System.getProperty("os.name").contains("Windows")`. Or just use Apache Lang `SystemUtils.IS_OS_WINDOWS`

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
```diff
@@ -146,9 +148,11 @@
      */
     public static final String PARAM_TEMPLATES_NAME = "nuxeo.templates";
 
+    @Deprecated(forRemoval = true) // not used
```
efge: Can you add `since  = "11.5"` too in all the new `@Deprecated`?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationHolder.java
```diff
@@ -0,0 +1,246 @@
...
+        this(home, home.resolve("bin").resolve("nuxeo.conf"));
+    }
+
+    public ConfigurationHolder(Path home, Path nuxeoConf) {
+        this.home = requireNonNull(home).toAbsolutePath();
+        this.nuxeoConf = requireNonNull(nuxeoConf).toAbsolutePath();
+        // initialize some defaults
+        var basicConfig = new Properties();
+        UnaryOperator<String> homeResolver = k -> getHomePath().resolve(k).toString();
+        UnaryOperator<String> runtimeResolver = k -> getRuntimeHomePath().resolve(k).toString();
```
efge, line 77: Why not regular member functions?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
```diff
@@ -92,23 +94,37 @@
...
-    private static final String NEW_FILES = ConfigurationGenerator.TEMPLATES + File.separator + "files.list";
+    private static final String NEW_FILES = "files.list";
 
     protected final ConfigurationGenerator generator;
 
+    /** @since 11.5 */
+    protected final ConfigurationHolder configHolder;
+
+    @Deprecated
     protected File dataDir = null;
```
efge: Could you remove all the useless `= null`?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
```diff
@@ -349,31 +338,35 @@ public void setPidDir(String pidDirStr) {
      * @since 5.4.2
      */
     public void checkPaths() throws ConfigurationException {
-        File badInstanceClid = new File(generator.getNuxeoHome(),
-                getDefaultDataDir() + File.separator + "instance.clid");
-        if (badInstanceClid.exists() && !getDataDir().equals(badInstanceClid.getParentFile())) {
-            log.warn("Moving {} to {}.", () -> badInstanceClid, this::getDataDir);
+        Path badInstanceClid = configHolder.getRuntimeHomePath()
+                                           .resolve(Environment.DEFAULT_DATA_DIR)
+                                           .resolve("instance.clid");
```
efge: You could have a constant `INSTANCE_CLID_PATH = Path.of(Environment.DEFAULT_DATA_DIR, "instance.clid")` and do `.resolve(INSTANCE_CLID_PATH)`

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
```diff
@@ -347,7 +350,7 @@ public ConfigurationGenerator(boolean quiet, boolean debug) {
         }
         String nuxeoConfPath = System.getProperty(NUXEO_CONF);
         if (nuxeoConfPath == null) {
-            configHolder = new ConfigurationHolder(nuxeoHome);
+            configHolder = new ConfigurationHolder(nuxeoHome, nuxeoHome.resolve("bin").resolve("nuxeo.conf"));
```
efge: The resolve part coud be a constant Path.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
```diff
@@ -392,6 +395,7 @@ protected boolean isGenerationOnce() {
     /**
      * @see #PARAM_FORCE_GENERATION
      */
+    @Deprecated(since = "11.1") // not used
```
efge: 11.5 ?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationHolder.java
```diff
@@ -115,6 +114,10 @@ public Path getLogPath() {
         return getPropertyAsPath(Environment.NUXEO_LOG_DIR);
     }
 
+    public Path getLogConfigPath() {
+        return getHomePath().resolve("lib").resolve("log4j2.xml");
```
efge, line 125: Resolve part could be a single constant Path.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/info/InstanceInfo.java
```diff
@@ -54,4 +71,74 @@ public InstanceInfo() {
...
+     */
+    public static InstanceInfo from(ConfigurationHolder configHolder, String clid, List<LocalPackage> pkgs) {
+        InstanceInfo nxInstance = new InstanceInfo();
+        nxInstance.NUXEO_CONF = configHolder.getNuxeoConfPath().toString();
+        nxInstance.NUXEO_HOME = configHolder.getHomePath().toString();
+        // distribution
+        Path distFile = configHolder.getConfigurationPath().resolve("distribution.properties");
+        if (Files.notExists(distFile)) {
+            // fallback in the file in templates
+            distFile = configHolder.getTemplatesPath().resolve(Path.of("common", "config", "distribution.properties"));
```
efge, line 95: Path could be a constant at class level.

server/nuxeo-launcher/src/main/java/org/nuxeo/launcher/NuxeoLauncher.java
```diff
@@ -701,7 +707,7 @@ private File getJavaExecutable() {
     }
 
     protected String getClassPath() {
-        File binDir = configurationGenerator.getNuxeoBinDir();
+        File binDir = configurationGenerator.getConfigurationHolder().getHomePath().resolve("bin").toFile();
```
efge, line 710: Could `"bin"` be a constant Path defined at class level?

efge:
The ticket mentions that this will be backported. What's the plan here? I'm not confortable at all with backporting all the refactoring.

kevinleturc:
> The ticket mentions that this will be backported. What's the plan here? I'm not confortable at all with backporting all the refactoring.

I didn't plan to backport these changes, indeed there're too many 'break the compatibility' things on java code.

At this moment, I plan to try to backport the last commit, but as far as I saw, it won't be enough to make our whole system works with the new way to declare templates. For instance, checkers and showconf won't be able to get templates declared with the new way. My big concern is around the `writeConfiguration` method which could mess up the `nuxeo.conf` file due to the de-sync between the `templates` field and the `nuxeo.templates` parameter.

I'll be happy to schedule a meeting to talk about this.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
kevinleturc: These were added to help me during the refactoring, I will clean intermediate commits and remove them.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationHolder.java
kevinleturc: Mainly because I can't make them static as I need the holder instance, thus making them as regular fields will cause IDE warns saying these can be a local variable.

Also, I didn't want to make them a class method, otherwise, they will be accessible to the whole package and I didn't want to expose a method returning a Path as a String.

dmetzler:
The backport is needed because this feature is required to make true Docker images that don't contain runtime configuration. We are currently relying on a hack at the cloud level to include the `mongodb` template in the image configuration that is not satisfying: it should be done at the binding (`conf.d`) level, but there is currently no way for composing the `nuxeo.templates` value from different sources. 

Depending on the timeline for the 11.x, LTS or 10.10, we'll have to discuss about this.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
kevinleturc: The whole class is deleted in the end, I will just make sure I didn't add intermediate changes.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
kevinleturc: Sure, I will name the constant `BAD_INSTANCE_CLID_PATH` to avoid confusion as the clid must follow the data path.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationHolder.java
```diff
@@ -0,0 +1,246 @@
...
+    public Path getHomePath() {
+        return home;
+    }
+
+    public Path getNuxeoConfPath() {
+        return nuxeoConf;
+    }
+
+    public Path getTemplatesPath() {
+        return home.resolve("templates");
```
atchertchian, line 98: Use constant?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationHolder.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /**
+     * Returns the Home of NuxeoRuntime (same as {@code Framework.getRuntime().getHome()})
+     */
+    public Path getRuntimeHomePath() {
+        return home.resolve("nxserver");
```
atchertchian, line 105: Use constant?
Also it was pointed out recently on slack doc channel that nuxeo.conf property `templateName.target` default value should not be `server/default/deploy/nuxeo.ear` but `nxserver`, but it seems this property does not exist anymore (?) 
https://doc.nuxeo.com/nxdoc/configuration-parameters-index-nuxeoconf/#templatenametarget

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
```diff
@@ -594,19 +572,16 @@ public String getDefaultPackagesDir() {
...
     public InstanceInfo getInfo(String clid, List<LocalPackage> pkgs) {
         InstanceInfo nxInstance = new InstanceInfo();
-        nxInstance.NUXEO_CONF = generator.getNuxeoConf().getPath();
-        nxInstance.NUXEO_HOME = generator.getNuxeoHome().getPath();
+        nxInstance.NUXEO_CONF = configHolder.getNuxeoConfPath().toString();
+        nxInstance.NUXEO_HOME = configHolder.getHomePath().toString();
         // distribution
-        File distFile = new File(generator.getConfigDir(), "distribution.properties");
-        if (!distFile.exists()) {
+        Path distFile = configHolder.getConfigurationPath().resolve("distribution.properties");
```
atchertchian: Use constant?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
kevinleturc: Removed from appropriate commit as it is an intermediate thing.

modules/runtime/nuxeo-launcher-commons/src/test/java/org/nuxeo/launcher/config/ConfigurationGeneratorTest.java
```diff
@@ -561,22 +527,4 @@ protected boolean isTemplateIncluded(String template) {
         return configGenerator.getIncludedTemplates().stream().map(File::getName).anyMatch(isEqual(template));
     }
 
-    /**
-     * Checks environment variable replacement within templates (NXP-29392).
-     *
-     * @since 11.2
-     */
-    @Test
-    public void testEnvironmentVariableInNuxeoDefaults() {
```
jcarsique, line 570: is this feature covered in another test?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ServerConfigurator.java
```diff
@@ -349,31 +338,35 @@ public void setPidDir(String pidDirStr) {
...
     public void checkPaths() throws ConfigurationException {
-        File badInstanceClid = new File(generator.getNuxeoHome(),
-                getDefaultDataDir() + File.separator + "instance.clid");
-        if (badInstanceClid.exists() && !getDataDir().equals(badInstanceClid.getParentFile())) {
-            log.warn("Moving {} to {}.", () -> badInstanceClid, this::getDataDir);
+        Path badInstanceClid = configHolder.getRuntimeHomePath()
+                                           .resolve(Environment.DEFAULT_DATA_DIR)
+                                           .resolve("instance.clid");
+        Path dataPath = configHolder.getDataPath();
+        if (Files.exists(badInstanceClid) && !dataPath.startsWith(badInstanceClid)) {
```
jcarsique: Isn't it rather `!badInstanceClid.startsWith(dataPath)` ?

atchertchian (commented):
Nice work :+1: 
I reviewed commit by commit and was a bit confused by some changes done over several commits (move of some tests between commit for "Extract configuration holder" and "Extract configuration loader" for instance).
Although this cleanup is great on master, I'm also worried about backport so I hope the changes can be reduced to handle only the main purpose.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationHolder.java
```diff
@@ -0,0 +1,246 @@
...
+
+    protected void clear() {
+        defaultConfig.clear();
+        userConfig.clear();
+    }
+
+    protected void setConfigurable(boolean configurable) {
+        this.configurable = configurable;
+        if (!configurable) {
+            // clear the configuration - probably an error case
```
jcarsique: you can use the above clear() method.
where is used the setConfigurable() method and why clearing the config when it's false?
maybe is it a remnant of the "static server" with read-only configuration?

jcarsique (dismissed):
Nice improvement.

Looks like it changed while I was reviewing, thus some comments about "configurable" and "clear()" look outdated. There was a inversion on `dataPath.startsWith(badInstanceClid)` that I don't see neither.

Is there no FreeMarker overlap between "nuxeo.templates" and "nuxeo.templates.*" ?

For the mentioned backport, not an easy task!, I would tend to add the feature with some bridging without breaking the API instead of a refactoring.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationHolder.java
kevinleturc: Indeed the documentation should be updated to reflect that and also to reflect changes made in this PR.

The property still exists, you can find it [there](https://github.com/nuxeo/nuxeo/pull/4603/files#diff-2b3d798ff8d6cb0c3fff75b05e57d57a0f49a53e660107aa30d6a0db459a7c3cR308).

modules/runtime/nuxeo-launcher-commons/src/test/java/org/nuxeo/launcher/config/ConfigurationGeneratorTest.java
kevinleturc: Yes, it has been moved [there](https://github.com/nuxeo/nuxeo/pull/4603/files#diff-f201073b7409397d423bfc3eeaee09a2c409e7fe392e2638788d14e34b5eeaecR67-R85).

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationHolder.java
kevinleturc: I didn't find any usage of `ConfigurationGenerator#isConfigurable`, I can't tell for what it was used. Maybe it was for the static server indeed.

At this point of the changes, I was clearing the config when it was false because it was the behavior the generator was doing when this flag was false. I will try to clean this change from the intermediate commit as it doesn't make sense because the whole thing was replaced by has the config been loaded.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
```diff
@@ -300,12 +301,24 @@ private void loadConfiguration(boolean evalDynamicProperties) throws Configurati
...
+        // include nuxeo.templates.*
+        String templatesWildcard = configHolder.stringPropertyNames()
+                                               .stream()
+                                               .filter(k -> k.startsWith(PARAM_TEMPLATES_NAME + '.'))
+                                               .sorted()
+                                               .map(configHolder::getProperty)
+                                               .collect(Collectors.joining(TEMPLATE_SEPARATOR));
+        if (isNotBlank(templatesWildcard)) {
+            templates += TEMPLATE_SEPARATOR + templatesWildcard;
+        }
```
kevinleturc, line 320: I'm continuing the discussion here in order to have a thread, @jcarsique:
> Is there no FreeMarker overlap between "nuxeo.templates" and "nuxeo.templates.*" ?

It seems there is a freemarker WARN log when both properties exist, see below:

```lang-none
WARN  [TextTemplate] FreeMarker variables: ignored 'nuxeo.templates.cloud' conflicting with 'nuxeo.templates'
```

If I understand correctly, it means that `nuxeo.templates.cloud` won't be available for variable substitution in freemarker templating.

I don't think it is something we want, but I may be wrong.

As it is our code, and depending on what we want for these variables, do you think we can ignore the log in `TextTemplate`?

jcarsique: Yes: FreeMarker does not like its variables to overlap ("a.b" vs "a.b.c") and that makes it buggy if used in a substitution.
I would rather try to migrate and remove the default "nuxeo.templates" to "nuxeo.templates.all" or anything similar.
Or use "nuxeo.template.new" instead of "nuxeo.templates.new".

jcarsique: Confirmed again today: that will make `nuxeo.templates` being a Map instead of String, thus breaking any current usage in the templates.

kevinleturc: Ok, it doesn't like a way to go.

Migrating the current to `nuxeo.templates.all` seems to me a huge breaking change, even if I break everything regarding java code, we should be good regarding _nuxeo.conf_.

I'll check with the team what we can do. 

Thanks for the details.

modules/runtime/nuxeo-launcher-commons/src/test/java/org/nuxeo/launcher/config/ConfigurationMarshallerTest.java
```diff
@@ -0,0 +1,308 @@
...
+    protected Path nuxeoConf;
+
+    protected ConfigurationMarshaller marshaller = new ConfigurationMarshaller(System.getProperties());
+
+    protected ConfigurationHolder configHolder;
+
+    @Before
+    public void setUp() throws Exception {
+        testDirectory = getResourcePath("configuration/marshaller").resolve(name.getMethodName());
+        // copy the file, some IDE doesn't like file rewriting, so keep a source
```
ataillefer, line 64: some IDEs don't like

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/DBCheck.java
```diff
@@ -0,0 +1,160 @@
...
+        Properties conProps = new Properties();
+        conProps.put("user", dbUser);
+        conProps.put("password", dbPassword);
+        log.debug("Testing URL: {} with: {}", url, conProps);
+        Connection con = driver.connect(url, conProps);
+        con.close();
+    }
+
+    /**
+     * Build an {@link URLClassLoader} for the given databaseTemplate looking in the templates directory and in the
```
ataillefer, line 126: Builds

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationChecker.java
```diff
@@ -0,0 +1,483 @@
...
+    /**
+     * Check the java version present in system properties compared to compliant ones.
+     */
+    protected void checkJavaVersionIsCompliant() throws ConfigurationException {
+        String version = systemProperties.getProperty("java.version");
+        checkJavaVersion(version, COMPLIANT_JAVA_VERSIONS);
+    }
+
+    /**
+     * Check the java version present in system properties compared to compliant ones.
```
ataillefer, line 134: Checks the given java version compared to the given compliant ones.

server/nuxeo-launcher/src/test/java/org/nuxeo/launcher/TestNuxeoLauncher.java
```diff
@@ -41,45 +43,43 @@
...
     @Rule
-    public final ExpectedException thrown = ExpectedException.none();
+    public final ConfigurationRule rule = new ConfigurationRule("launcher");
 
-    @Override
     @Before
-    public void setUp() throws Exception {
+    public void setUp() {
+        System.setProperty(NUXEO_HOME, rule.getNuxeoHome().toString());
+        // re-init default env to allow different nuxeo home path for each tests
```
ataillefer, line 76: for each test

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/info/InstanceInfo.java
```diff
@@ -54,4 +78,74 @@ public InstanceInfo() {
     @XmlElement(name = "configuration")
     public ConfigurationInfo config;
 
+    /**
+     * Introspect the server and builds the instance info.
```
ataillefer, line 87: Introspects

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
```diff
@@ -304,12 +305,24 @@ private void loadConfiguration(boolean evalDynamicProperties) throws Configurati
...
      */
     protected void includeTemplates() throws ConfigurationException {
+        // include nuxeo.templates
         String templates = configHolder.getProperty(PARAM_TEMPLATES_NAME);
         if (isBlank(templates)) {
             log.warn("No template found in configuration! Fallback on 'default'.");
             templates = "default";
             configHolder.put(PARAM_TEMPLATES_NAME, templates);
         }
+        // include nuxeo.templates.*
```
ataillefer, line 315: Isn't it `nuxeo.append.templates.*`?

modules/runtime/nuxeo-launcher-commons/src/test/java/org/nuxeo/launcher/config/ConfigurationGeneratorTest.java
```diff
@@ -329,6 +330,23 @@ public void testIncludeProfile() throws Exception {
...
+    @Test
+    public void testIncludeTemplatesWildcard() throws Exception {
+        ConfigurationGenerator generator = generatorBuilder().init(true).build();
+        ConfigurationHolder configHolder = generator.getConfigurationHolder();
+        // test templates are all included
+        List<String> includedTemplates = configHolder.getIncludedTemplateNames();
+        assertTrue(includedTemplates.contains("default"));
+        assertTrue(includedTemplates.contains("test"));
+        assertTrue(includedTemplates.contains("mongodb"));
+        assertTrue(includedTemplates.contains("cloud-directupload"));
```
ataillefer, line 342: Are these needed given the ordered assertion below?

ataillefer (commented):
**Monumental** work!
Nitpicking mostly :smile:

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
```diff
@@ -643,25 +605,6 @@ protected void includeTemplates() throws IOException {
         log.debug(includedTemplates);
     }
 
-    private void logDebugInformation() {
-        String devPropValue = userConfig.getProperty(NUXEO_DEV_SYSTEM_PROP);
-        if (Boolean.parseBoolean(devPropValue)) {
-            log.debug("Nuxeo Dev mode enabled");
-        } else {
-            log.debug("Nuxeo Dev mode is not enabled");
-        }
```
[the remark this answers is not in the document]
kevinleturc: Not really, they appear when the launcher has the option `--debug`, I didn't find them really useful. Wdyt?

modules/runtime/nuxeo-launcher-commons/src/test/java/org/nuxeo/launcher/config/ConfigurationGeneratorTest.java
kevinleturc: Indeed, let leverages the ordered assertion.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
kevinleturc: We've decided to move from `nuxeo.templates.*` to `nuxeo.append.templates.*` to avoid this issue.

I'm resolving this thread, don't hesitate to re-open it if needed.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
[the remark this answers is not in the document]
jcarsique: maybe to be moved to trace; I remember that the dev mode introduced many code branches, maybe making it worth to warn the developer about the dev mode status?

The debug launcher option is a distinct subject, very much useful and missing logs since the move to Log4j2.

I would keep those logs unless the dev mode was removed. That's the purpose of a debug option to log everything relevant.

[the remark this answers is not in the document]
kevinleturc: Moving them to trace will definitively hide them.

We're warning the developer when the dev mode is enabled in the docker packaging, as it is the target I would say it is good to leverage the docker warn.

If it's not enough, then we can keep the _Nuxeo Dev mode is enabled_ log but in this case, I would log it at info level.

Wdyt?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationHolder.java
```diff
@@ -0,0 +1,322 @@
...
+     * associated with the given {@code key}.
+     */
+    public String getProperty(String key) {
+        return userConfig.getProperty(key);
+    }
+
+    /**
+     * Returns the property value associated with the given {@code key} from the user configuration.
+     * <p>
+     * The method falls back on the default configuration if the user configuration doesn't contain a property value
```
ataillefer, line 170: on the given {@code defaultValue} no?

kevinleturc: after falling back on default configuration, will add a comment.

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/config/ConfigurationGenerator.java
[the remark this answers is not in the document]
atchertchian: I kept these some (long) time ago when cleaning up the different "dev mode" notions (with "test", "debug", and "strict" modes...).
When in dev mode, some of the code logic is not optimized so i think it makes sense to keep one log when it's enabled, and remove the other one, as you're suggesting :+1:.
I'd use the warn level instead of info (and it would be consistent with the docker packaging behaviour)

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/DBCheck.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @Override
+    public void check(ConfigurationHolder configHolder) throws ConfigurationException {
+        try {
+            checkDatabaseConnection(configHolder);
+        } catch (DatabaseDriverException e) {
+            log.debug(e, e);
+            log.error(e.getMessage());
+            throw new ConfigurationException("Could not find database driver: " + e.getMessage());
```
atchertchian, line 82: why not showing the full stack trace on error, not only with debug logs? (same just below)

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/DBCheck.java
```diff
@@ -0,0 +1,160 @@
...
+            throws DatabaseDriverException {
+        File[] files = ArrayUtils.addAll( //
+                databaseTemplateDir.resolve("lib").toFile().listFiles(), //
+                configHolder.getHomePath().resolve("lib").toFile().listFiles());
+        List<URL> urlsList = new ArrayList<>();
+        if (files != null) {
+            for (File file : files) {
+                if (file.getName().endsWith("jar")) {
+                    try {
+                        urlsList.add(new URL("jar:file:" + file.getPath() + "!/"));
```
atchertchian, line 144: Could this be cleaned up following similar code on the NuxeoDeployer, using a PathMatcher to lookup libs? (https://github.com/nuxeo/nuxeo/blob/master/modules/runtime/nuxeo-runtime-tomcat-adapter/src/main/java/org/nuxeo/runtime/tomcat/NuxeoDeployer.java#L100)

modules/platform/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/main/java/org/nuxeo/elasticsearch/ElasticSearchChecker.java
```diff
@@ -67,8 +60,10 @@ public boolean accepts(ConfigurationGenerator cg) {
     }
 
     @Override
-    public void check(ConfigurationGenerator cg) throws ConfigurationException {
-        ElasticSearchClientConfig config = getConfig(cg);
+    public void check(ConfigurationHolder configHolder) throws ConfigurationException {
+        ElasticSearchClientConfig config = getDescriptor(configHolder, CONFIG_NAME, ElasticSearchClientConfig.class,
+                // avoid XMap to fail when trying to load class value by removing class attribute
```
atchertchian, line 65: I'm interested to know more about this problem linked to xmap and classes introspection (for https://jira.nuxeo.com/browse/NXP-30092), although i realized you simply moved that comment.
Checking the class can be loaded does not make sense in this context maybe?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/info/InstanceInfo.java
```diff
@@ -19,19 +19,48 @@
...
+import org.nuxeo.common.codec.Crypto;
+import org.nuxeo.connect.update.LocalPackage;
+import org.nuxeo.launcher.config.ConfigurationHolder;
+
 @XmlAccessorType(XmlAccessType.NONE)
 @XmlRootElement(name = "instance")
 public class InstanceInfo {
 
+    /** @since 11.5 */
+    protected static final String DISTRIBUTION_PROPS = "distribution.properties";
```
atchertchian, line 54: Why not defining it in ConfigurationConstants too?

atchertchian (approved):
:+1: Added a few comments for questions/details.

As a warning :grimacing:: changes linked to ConfigurationGenerator might have an impact on nuxeo-jsf-ui code (for seam hot reload management and some admin center screens)

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/DBCheck.java
kevinleturc: I think this is because we want a nice logging in the console, as we call this code through nuxeoctl. We get the stack in debug mode when we pass the --debug option.

modules/platform/nuxeo-elasticsearch/nuxeo-elasticsearch-core/src/main/java/org/nuxeo/elasticsearch/ElasticSearchChecker.java
kevinleturc: If I remember correctly we don't have everything in the class loader to load what the class attribute represents. Maybe @bdelbosc has more details?

modules/runtime/nuxeo-launcher-commons/src/main/java/org/nuxeo/launcher/info/InstanceInfo.java
kevinleturc: Indeed 🙂.

atchertchian (dismissed):

jcarsique (approved):

efge (approved):

efge:
What's new to review in this latest force push?

kevinleturc:
> What's new to review in this latest force push?

This [part](https://github.com/nuxeo/nuxeo/pull/4603/files#diff-11dfa3adfafcd1517aaf45daf14ccfdcdddc2733131093aa0138655b72021668R141-R151), the code wasn't NPE proof enough.

atchertchian (approved):

efge (approved):

jcarsique (approved):

ataillefer (approved):

PR #4606, opened 2021-01-26
Improvement nxp 30097 error handling async operation adapter

RSalem07:
These changes are required by the initial PR [ELEMENTS#345](https://github.com/nuxeo/nuxeo-elements/pull/345/), and the need on retention to manage the errors. As discussed it with Nelson, I have updated the backend part to throw an exception in a case where we face any issue with the bulk action, even completed.

### DIFF
--- a/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/AsyncStatus.java
+++ b/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/AsyncStatus.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright 2018 Nuxeo (http://nuxeo.com/) and others.
+ * (C) Copyright 2018-2021 Nuxeo (http://nuxeo.com/) and others.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,7 +24,6 @@ import java.io.Serializable;
  * Interface to be implemented by asynchronous services' statuses.
  *
  * @param <K> type of task id
- *
  * @since 10.3
  */
 public interface AsyncStatus<K extends Serializable> extends Serializable {
--- a/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/AsyncStatus.java
+++ b/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/AsyncStatus.java
@@ -18,6 +18,8 @@
  */
 package org.nuxeo.ecm.core.api;
 
+import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
+
 import java.io.Serializable;
 
 /**
@@ -37,4 +39,34 @@ public interface AsyncStatus<K extends Serializable> extends Serializable {
      * Returns true if command is completed.
      */
     boolean isCompleted();
+
+    /**
+     * Gets the error message if any.
+     *
+     * @return the error message if any, {@code null} otherwise
+     * @since 11.5
+     */
+    default String getErrorMessage() {
+        return null;
+    }
+
+    /**
+     * Gets the error code if any.
+     *
+     * @return the error code if any, {@code 0} otherwise
+     * @since 11.5
+     */
+    default int getErrorCode() {
+        return hasError() ? SC_INTERNAL_SERVER_ERROR : 0;
+    }
+
+    /**
+     * Checks if there is any error.
+     *
+     * @return {@code true} if there is any error, {@code false} otherwise
+     * @since 11.5
+     */
+    default boolean hasError() {
+        return getErrorMessage() != null;
+    }
 }
--- a/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
+++ b/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
@@ -20,6 +20,9 @@
 package org.nuxeo.ecm.core.bulk.computation;
 
 import static java.lang.Math.min;
+import static java.util.Objects.requireNonNullElse;
+import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
+import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
 import static org.nuxeo.ecm.core.bulk.BulkServiceImpl.STATUS_STREAM;
 import static org.nuxeo.ecm.core.bulk.message.BulkStatus.State.ABORTED;
 import static org.nuxeo.ecm.core.bulk.message.BulkStatus.State.COMPLETED;
@@ -207,11 +210,11 @@ public class BulkScrollerComputation extends AbstractComputation {
             }
         } catch (IllegalArgumentException | QueryParseException | DocumentNotFoundException e) {
             log.error("Invalid query results in an empty document set: {}", command, e);
-            updateStatusAfterScroll(context, commandId, "Invalid query");
+            updateStatusAfterScroll(context, commandId, "Invalid query", SC_BAD_REQUEST);
         } catch (NuxeoException e) {
             if (command != null) {
                 log.error("Invalid command produces an empty document set: {}", command, e);
-                updateStatusAfterScroll(context, command.getId(), "Invalid command");
+                updateStatusAfterScroll(context, command.getId(), "Invalid command", e.getStatusCode());
             } else {
                 log.error("Discard invalid record: {}", record, e);
             }
@@ -284,20 +287,21 @@ public class BulkScrollerComputation extends AbstractComputation {
                 BulkCodecs.getStatusCodec().encode(delta));
     }
 
-    protected void updateStatusAfterScroll(ComputationContext context, String commandId, String errorMessage) {
-        updateStatusAfterScroll(context, commandId, 0, errorMessage, false);
+    protected void updateStatusAfterScroll(ComputationContext context, String commandId, String errorMessage,
+            Integer errorCode) {
+        updateStatusAfterScroll(context, commandId, 0, errorMessage, errorCode, false);
     }
 
     protected void updateStatusAfterScroll(ComputationContext context, String commandId, long documentCount,
             boolean limited) {
-        updateStatusAfterScroll(context, commandId, documentCount, null, limited);
+        updateStatusAfterScroll(context, commandId, documentCount, null, null, limited);
     }
 
     protected void updateStatusAfterScroll(ComputationContext context, String commandId, long documentCount,
-            String errorMessage, boolean limited) {
+            String errorMessage, Integer errorCode, boolean limited) {
         BulkStatus delta = BulkStatus.deltaOf(commandId);
         if (errorMessage != null) {
-            delta.inError(errorMessage);
+            delta.inError(errorMessage, requireNonNullElse(errorCode, SC_INTERNAL_SERVER_ERROR));
         }
         if (documentCount == 0) {
             delta.setState(COMPLETED);
--- a/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
+++ b/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright 2018 Nuxeo SA (http://nuxeo.com/) and others.
+ * (C) Copyright 2018-2021 Nuxeo (http://nuxeo.com/) and others.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
+++ b/modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
@@ -19,6 +19,8 @@
 
 package org.nuxeo.ecm.core.bulk.message;
 
+import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
+
 import java.io.Serializable;
 import java.time.Instant;
 import java.util.Collections;
@@ -81,6 +83,10 @@ public class BulkStatus implements AsyncStatus<String> {
 
     @Nullable
     protected String errorMessage;
+
+    /** @since 11.5 **/
+    @Nullable
+    protected Integer errorCode;
 
     @Nullable
     protected Long processed;
@@ -211,6 +217,9 @@ public class BulkStatus implements AsyncStatus<String> {
         }
         if (update.errorMessage != null && errorMessage == null) {
             errorMessage = update.errorMessage;
+        }
+        if (update.errorCode != null && errorCode == null) {
+            errorCode = update.errorCode;
         }
         if (update.queryLimitReached) {
             queryLimitReached = true;
@@ -435,6 +444,7 @@ public class BulkStatus implements AsyncStatus<String> {
         this.processingDurationMillis = processingDurationMillis;
     }
 
+    @Override
     public boolean hasError() {
         return errorCount > 0;
     }
@@ -461,12 +471,22 @@ public class BulkStatus implements AsyncStatus<String> {
      * An error occurred during the processing
      */
     public void inError(String message) {
+        inError(message, SC_INTERNAL_SERVER_ERROR);
+    }
+
+    /**
+     * An error occurred during the processing
+     * 
+     * @since 11.5
+     */
+    public void inError(String message, int code) {
         if (isDelta()) {
             errorCount = 1;
         } else {
             errorCount++;
         }
         this.errorMessage = message;
+        this.errorCode = code;
     }
 
     @Override
@@ -484,4 +504,11 @@ public class BulkStatus implements AsyncStatus<String> {
         return ToStringBuilder.reflectionToString(this);
     }
 
+    @Override
+    public int getErrorCode() {
+        if (errorCode == null) {
+            return 0;
+        }
+        return errorCode;
+    }
 }
--- a/modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestBulkProcessor.java
+++ b/modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestBulkProcessor.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright 2018 Nuxeo (http://nuxeo.com/) and others.
+ * (C) Copyright 2018-2021 Nuxeo (http://nuxeo.com/) and others.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -69,6 +69,7 @@ import org.nuxeo.runtime.transaction.TransactionHelper;
 @RunWith(FeaturesRunner.class)
 @Features(CoreFeature.class)
 public class TestBulkProcessor {
+
     private static final Logger log = org.apache.logging.log4j.LogManager.getLogger(TestBulkProcessor.class);
 
     @Inject
@@ -324,7 +325,7 @@ public class TestBulkProcessor {
         // now append first bucket
         service.appendExternalBucket(new BulkBucket(commandId, docs.subList(0, nbDocs / 2)));
         // wait for the status
-        await().atMost(ONE_MINUTE) .until(() -> service.getStatus(commandId).getProcessed() == nbDocs / 2);
+        await().atMost(ONE_MINUTE).until(() -> service.getStatus(commandId).getProcessed() == nbDocs / 2);
         status = service.getStatus(commandId);
         assertEquals(SCROLLING_RUNNING, status.getState());
         assertEquals(0, status.getTotal());
@@ -365,7 +366,7 @@ public class TestBulkProcessor {
         // prepare and submit the command with an external scroller and submit it
         String commandId = service.submit(
                 new BulkCommand.Builder(SetPropertiesAction.ACTION_NAME, "ignored", "system").useExternalScroller()
-                        .build());
+                                                                                             .build());
         // wait for the scroller
         await().atMost(ONE_MINUTE).until(() -> service.getStatus(commandId).getState() == SCROLLING_RUNNING);
         BulkStatus status = service.getStatus(commandId);
--- a/modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestBulkProcessor.java
+++ b/modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestBulkProcessor.java
@@ -18,6 +18,7 @@
  */
 package org.nuxeo.ecm.core.bulk;
 
+import static org.apache.http.HttpStatus.SC_BAD_REQUEST;
 import static org.awaitility.Awaitility.await;
 import static org.awaitility.Duration.ONE_MINUTE;
 import static org.junit.Assert.assertEquals;
@@ -81,6 +82,9 @@ public class TestBulkProcessor {
     @Inject
     public StreamService stream;
 
+    @Inject
+    public CoreFeature coreFeature;
+
     @Test
     public void testEmptyQuery() throws InterruptedException {
         @SuppressWarnings("resource")
@@ -128,6 +132,7 @@ public class TestBulkProcessor {
         assertEquals(COMPLETED, status.getState());
         assertEquals(0, status.getTotal());
         assertTrue(status.hasError());
+        assertEquals(SC_BAD_REQUEST, status.getErrorCode());
 
         // query with error
         nxql = "SELECT * FROM Document WHERE ecm:path = 'non/existing/path'";
@@ -138,6 +143,7 @@ public class TestBulkProcessor {
         assertEquals(COMPLETED, status.getState());
         assertEquals(0, status.getTotal());
         assertTrue(status.hasError());
+        assertEquals(SC_BAD_REQUEST, status.getErrorCode());
     }
 
     @Test
--- a/modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
+++ b/modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
@@ -44,6 +44,7 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.nuxeo.ecm.automation.AutomationService;
@@ -333,11 +334,38 @@ public class AsyncOperationAdapter extends DefaultAdapter {
 
     public int getErrorStatus(String executionId) {
         Long status = (Long) getTransientStore().getParameter(executionId, TRANSIENT_STORE_ERROR_STATUS);
-        return status == null ? 0 : status.intValue();
+        if (status != null) {
+            return status.intValue();
+        }
+
+        // check the errors, if the execution is a bulk action
+        Serializable taskId = getAsyncTaskId(executionId);
+        if (taskId != null) {
+            AsyncStatus<?> asyncStatus = getAsyncService(executionId).getStatus(taskId);
+            if (asyncStatus != null) {
+                return asyncStatus.getErrorCode();
+            }
+        }
+
+        return 0;
     }
 
     public String getError(String executionId) {
-        return (String) getTransientStore().getParameter(executionId, TRANSIENT_STORE_ERROR);
+        String error = (String) getTransientStore().getParameter(executionId, TRANSIENT_STORE_ERROR);
+        if (StringUtils.isNotEmpty(error)) {
+            return error;
+        }
+
+        // check the errors, if the execution is a bulk action
+        Serializable taskId = getAsyncTaskId(executionId);
+        if (taskId != null) {
+            AsyncStatus<?> asyncStatus = getAsyncService(executionId).getStatus(taskId);
+            if (asyncStatus != null) {
+                return asyncStatus.getErrorMessage();
+            }
+        }
+
+        return null;
     }
 
     protected void setOutput(String executionId, Serializable output) {
--- a/modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
+++ b/modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright 2018 Nuxeo (http://nuxeo.com/) and others.
+ * (C) Copyright 2018-2021 Nuxeo (http://nuxeo.com/) and others.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -113,11 +113,11 @@ public class AsyncOperationAdapter extends DefaultAdapter {
 
     protected static final String TRANSIENT_STORE_OUTPUT_BLOB = "blob";
 
-    protected static final String STATUS_PATH= "status";
+    protected static final String STATUS_PATH = "status";
 
-    protected static final String RUNNING_STATUS= "RUNNING";
+    protected static final String RUNNING_STATUS = "RUNNING";
 
-    protected static final String RESULT_URL_KEY= "url";
+    protected static final String RESULT_URL_KEY = "url";
 
     protected static final ObjectMapper MAPPER = new ObjectMapper();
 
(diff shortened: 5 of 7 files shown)


modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
```diff
@@ -318,7 +323,20 @@ protected void setError(String executionId, String error) {
...
+            return error;
+        }
+
+        if (!isAsync(executionId)) {
+            return null;
+        }
+
+        // Error on the bulk action
+        Serializable taskId = getTaskId(executionId);
+        BulkStatus status = Framework.getService(BulkService.class).getStatus((String) taskId);
```
nelsonsilva, line 338: Relying directly on `BulkStatus` here breaks the `AsyncStatus` abstraction. We should introduce a `getError()` on `AsynStatus` instead and rely on `getAsyncService(executionId).getStatus(taskId).getError()`.

modules/platform/nuxeo-automation/nuxeo-automation-test/src/test/java/org/nuxeo/ecm/automation/server/test/AsyncOperationAdapterTest.java
```diff
@@ -207,6 +207,14 @@ public void testError() throws Exception {
                             .set("rollback", true)
                             .executeReturningExceptionEntity(SC_INTERNAL_SERVER_ERROR);
         assertEquals("Internal Server Error", error);
+
+        // Use an invalid query to ensure we get a bulk action error, even it's completed
```
ataillefer, line 211:
```suggestion
        // Use an invalid query to ensure we get a bulk action error, even if it's completed
```

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
RSalem07: Yep, I see you point, but I am wondering about the backward compatibility, If we do, I thing we have to make something like:

```

public interface AsyncStatus<K extends Serializable> extends Serializable {

    /**
     * Returns the number of errors encountered.
     */
    public default long getErrorCount() {
        return 0;
    }

    /**
     * Returns the first error message if any or null.
     */
    public default String getErrorMessage() {
        return null;
    }

    /**
     * Returns the number of errors encountered.
     */
    public default long getErrorCount() {
        return 0;
    }

    /**
     * Returns the first error message if any or null.
     */
    public default String getErrorMessage() {
        return null;
    }
....
```

Do you agree on this approche 

thx

nelsonsilva: You mean in terms of possible implementations that would be missing this new method ? I'm ok with the default interface imlpementations as you proposed. As for the methods themselves up to you really, I'd introduce only the methods you'll actually use, having an error count without a way to access the other errors seems pretty pointless to me tbh. So, if in the end you only use one error message, which is the first error in the bulkstatus, this is an implementation detail imho and I'd have a simple `getError()` in `AsyncStatus` returning it

RSalem07: > You mean in terms of possible implementations that would be missing this new method ? 

exactly, sorry I didn't mention it, but yes

> I'm ok with the default interface implementations as you proposed.

yep

> I'd introduce only the methods you'll actually use, having an error count without a way to access the other errors 

seems pretty pointless to me tbh. So, if in the end you only use one error message, which is the first error in the bulkstatus, this is an implementation detail imho and I'd have a simple getError() in AsyncStatus returning it
you are right, completely agree with you, I'll only keep `getError` one, sorry I don't I added the two ones

thx

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/AsyncStatus.java
```diff
@@ -37,4 +39,24 @@
      * Returns true if command is completed.
      */
     boolean isCompleted();
+
+    /**
+     * Gets the message error if any.
```
ataillefer, line 44: error message

ataillefer (dismissed):

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
```diff
@@ -318,7 +323,20 @@ protected void setError(String executionId, String error) {
...
+            return error;
+        }
+
+        if (!isAsync(executionId)) {
+            return null;
+        }
+
+        // Error on the bulk action
+        Serializable taskId = getTaskId(executionId);
+        AsyncStatus<String> status = Framework.getService(BulkService.class).getStatus((String) taskId);
```
nelsonsilva, line 338:
```suggestion
        AsyncStatus<String> status = getAsyncService(executionId).getStatus(taskId);
```

modules/platform/nuxeo-automation/nuxeo-automation-test/src/test/java/org/nuxeo/ecm/automation/server/test/AsyncOperationAdapterTest.java
```diff
@@ -207,6 +207,14 @@ public void testError() throws Exception {
                             .set("rollback", true)
                             .executeReturningExceptionEntity(SC_INTERNAL_SERVER_ERROR);
         assertEquals("Internal Server Error", error);
+
+        // Use an invalid query to ensure we get a bulk action error, even if it's completed
+        JsonNode jsonNode= async.newRequest(BulkRunAction.ID)
```
nelsonsilva, line 212: nit: missing space before `=`

nelsonsilva (approved):

efge:
#4659, which is has many changes around error management, also touches some of the same classes. Please wait for it to be merge before rebasing this on it and checking all tests still pass.

RSalem07:
@efge thx, I'll wait :)

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
```diff
@@ -333,11 +334,41 @@ protected void setError(String executionId, Throwable t) {
...
+        // check the errors, if the execution is a bulk action
+        AsyncStatus<Serializable> asyncStatus = getAsyncStatus(executionId);
+        if (asyncStatus != null) {
+            return asyncStatus.getError();
+        }
+
+        return null;
+    }
+
+    protected AsyncStatus<Serializable> getAsyncStatus(String executionId) {
```
efge, line 365: I think this should be `AsyncStatus<String>` everywhere to avoid casting later

RSalem07: If we want to update that, we have to modify the signature of [AsyncOperationAdapter#AsyncService<Serializable, ?, ?> getAsyncService(String executionId)](https://github.com/nuxeo/nuxeo/blob/00bc07d1f2ff374a703b1babaaa22f9931f8fa1b/modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java#L445) Which will has an impact, I am not sure that the type will be always `String`.

[AsyncService(https://github.com/nuxeo/nuxeo/blob/00bc07d1f2ff374a703b1babaaa22f9931f8fa1b/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/AsyncService.java#L38)

efge: Ok but in the same file all the other calls to `getAsyncService(executionId).getStatus(taskId)` don't need a cast, so I don't know why you have one...

RSalem07: ah yes, I think I'll remove this function. And, keep the same logic than the others call.

Thanks,

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
```diff
@@ -333,11 +334,38 @@ protected void setError(String executionId, Throwable t) {
...
     public int getErrorStatus(String executionId) {
         Long status = (Long) getTransientStore().getParameter(executionId, TRANSIENT_STORE_ERROR_STATUS);
-        return status == null ? 0 : status.intValue();
+        if (status != null) {
+            return status.intValue();
+        }
+
+        // check the errors, if the execution is a bulk action
+        if (isAsync(executionId)) {
+            Serializable taskId = getAsyncTaskId(executionId);
```
efge, line 343: Checking `isAsync` before calling `getAsyncTaskId` is redundant. The new pattern in this code after the latest refactoring/cleanup is just:

            Serializable taskId = getAsyncTaskId(executionId);
            if (taskId != null) {
                ...
            }

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/AsyncStatus.java
```diff
@@ -37,4 +39,24 @@
...
+    }
+
+    /**
+     * Checks if there is any error.
+     *
+     * @return {@code true} if there is any error, {@code false} otherwise
+     * @since 11.5
+     */
+    default boolean hasError() {
+        return StringUtils.isNotEmpty(getError());
```
efge, line 60: For a default method that gets overridden anyway in `BulkStatus` and is here only for the vanishingly unlikely case that someone has their own implementation of `AsyncStatus`, you should just do `return getError() != null`

modules/platform/nuxeo-automation/nuxeo-automation-test/src/test/java/org/nuxeo/ecm/automation/server/test/AsyncOperationAdapterTest.java
```diff
@@ -208,6 +208,18 @@ public void testError() throws Exception {
...
+        Map<String, Serializable> automationParams = Map.of("properties", "dc:title=folder_title_updated");
+        Map<String, Serializable> actionParams = Map.of(AutomationBulkAction.OPERATION_ID, "Document.Update", //
+                AutomationBulkAction.OPERATION_PARAMETERS, (Serializable) automationParams);
+        JsonNode jsonNode = async.newRequest(BulkRunAction.ID)
+                                 .set("action", AutomationBulkAction.ACTION_NAME)
+                                 .set("query", "SELCT * FROM Folder") // Invalid Query
+                                 .set("bucketSize", "10")
+                                 .set("batchSize", "5")
+                                 .set("parameters", MAPPER.writeValueAsString(actionParams))
+                                 .execute(SC_INTERNAL_SERVER_ERROR);
```
efge, line 222: I don't like that this is returning `SC_INTERNAL_SERVER_ERROR`; the latest changes were careful to return `SC_BAD_REQUEST` for all errors that are due to user input.

But this may mean slightly deeper changes to have `BulkStatus.inError()` take and record also an error status, and change its callers to provide one when available. So that typically `BulkScrollerComputation.processRecord` can provide one when calling `updateStatusAfterScroll` for an invalid query. And `AbstractBulkComputation.processFailure` should check if the `failure` is a `NuxeoException` and reuse its status code in that case.

modules/platform/nuxeo-automation/nuxeo-automation-server/src/main/java/org/nuxeo/ecm/automation/server/jaxrs/adapters/AsyncOperationAdapter.java
```diff
@@ -249,8 +250,8 @@ public Object result(@PathParam("executionId") String executionId) throws IOExce
             // cleanup after result is accessed
             cleanup(executionId);
 
-            if (status != 0) {
-                throw new NuxeoException(error, status);
+            if (status != 0 || StringUtils.isNotEmpty(error)) {
```
efge, line 253: Do we have legitimate cases where `error` can be just `""`? If not, I'd prefer a simple `null` check. Same in `getError`

efge (changes requested):

modules/platform/nuxeo-automation/nuxeo-automation-test/src/test/java/org/nuxeo/ecm/automation/server/test/AsyncOperationAdapterTest.java
RSalem07: I agree On that. And, as you has mentioned it, It needs a deeper changes on Bulk part. I'll go deeper and rework things.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
```diff
@@ -212,6 +217,9 @@ public void merge(@NotNull BulkStatus update) {
         if (update.errorMessage != null && errorMessage == null) {
             errorMessage = update.errorMessage;
         }
+        if (update.getErrorCode() != 0) {
+            errorCode = update.getErrorCode();
```
ataillefer, line 221: Worth putting `update.getErrorCode()` in a variable?

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestBulkProcessor.java
```diff
@@ -138,6 +145,8 @@ public void testInvalidQuery() throws InterruptedException {
         assertEquals(COMPLETED, status.getState());
         assertEquals(0, status.getTotal());
         assertTrue(status.hasError());
+        int expectedStatus = coreFeature.getStorageConfiguration().isDBS() ? SC_NOT_FOUND : SC_INTERNAL_SERVER_ERROR;
```
ataillefer, line 148: Why do we have a different behavior if DBS? Can you please add a comment?

RSalem07: Yes, The tests are down, On Mongo this is related to the fact, that for SBS the http response is not `500` but `404`

```
[2021-03-18T17:55:12.373Z] [ERROR] testInvalidQuery(org.nuxeo.ecm.core.bulk.TestBulkProcessor)  Time elapsed: 0.425 s  <<< FAILURE!
[2021-03-18T17:55:12.373Z] java.lang.AssertionError: expected:<500> but was:<404>
```
More on https://jenkins.platform.dev.nuxeo.com/blue/organizations/jenkins/nuxeo%2Fnuxeo/detail/PR-4606/9/pipeline/109

efge: I think the point is to understand why the server itself returns different error codes, and see if that's a bug (to fix) or something we want (to document). In general that's not something we want, we prefer identical behavior, including error behavior, from backends.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
```diff
@@ -206,12 +208,16 @@ protected void processRecord(ComputationContext context, Record record) {
                 updateStatusAfterScroll(context, commandId, documentCount, limitReached);
             }
         } catch (IllegalArgumentException | QueryParseException | DocumentNotFoundException e) {
+            int statusCode = HttpStatus.SC_INTERNAL_SERVER_ERROR;
+            if (e instanceof NuxeoException) {
+                statusCode = ((NuxeoException) e).getStatusCode();
+            }
```
efge, line 214: I think we want to turn `IllegalArgumentException` into a 400 as well.
For the `DocumentNotFoundException` I think there's a discussion to have to decide if we want to keep a 404 or turn it into a 400. I don't know when this exception is generated.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
```diff
@@ -35,6 +35,8 @@
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.nuxeo.ecm.core.api.AsyncStatus;
 
+import static org.apache.http.HttpStatus.SC_INTERNAL_SERVER_ERROR;
```
efge, line 38: The formatter puts static imports before other imports.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
```diff
@@ -435,6 +443,10 @@ public void setProcessingDurationMillis(long processingDurationMillis) {
         this.processingDurationMillis = processingDurationMillis;
     }
 
+    public void setErrorCode(int errorCode) {
```
efge, line 446: Please add `/** @since 11.5 */`

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
```diff
@@ -461,12 +473,20 @@ public void setErrorCount(long errorCount) {
      * An error occurred during the processing
      */
     public void inError(String message) {
+        inError(message, SC_INTERNAL_SERVER_ERROR);
+    }
+
+    /**
+     * An error occurred during the processing
+     */
+    public void inError(String message, int code) {
```
efge, line 482: `@since 11.5`

efge (changes requested):

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
RSalem07: sure, I completely forgot it

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
RSalem07: yes, agree on that. But I told myself if we find the issue and we align it to the same error code. It can break changes. But I'll take a look.

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestBulkProcessor.java
RSalem07: yes, agree on that. But I told myself if we find the issue and we align it to the same error code. It can break changes. But I'll take a look.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
RSalem07: Currently, the exception is raised [Here](https://github.com/nuxeo/nuxeo/blob/master/modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/SessionImpl.java#L576). I did a change on this one to make it more consistent with DBS. But I think we have to discuss the others ones, where we raise `IllegalArgumentException` for example [Here](https://github.com/nuxeo/nuxeo/blob/master/modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/SessionImpl.java#L584)

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/bulk/TestBulkProcessor.java
RSalem07: I did a quick thing to use the `DocumentNotFoundException` too for `VCS` more on [Here](https://github.com/nuxeo/nuxeo/pull/4606#discussion_r598898734).

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
efge: Hm changing code in `SessionImpl` is going too far for this commit. And there'd be many other similar changes to do if we want to be consistent, and probably unit tests to update etc.
Here you should just in the `catch` clause check if we have an `IllegalArgumentException` and then set the status to SC_BAD_REQUEST.

RSalem07: ah sorry, I see your point. I though we need to do it a low level.

efge: My " I don't know when this exception is generated" was about the `DocumentNotFoundException` as I don't know if it's an internal one that we catch for unrelated reason (in which case 500 is ok), or if it's a direct consequence of the user passing a reference to a doc that doesn't exist (in which case it's a 400).

RSalem07: yep, 
 For the query `SELECT * FROM Document WHERE ecm:path = 'non/existing/path'` we have two different logic: 

- DBS low level is raising a `DocumentNotFoundException` which is `404` [DBS#404](https://github.com/nuxeo/nuxeo/blob/master/modules/core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSSession.java#L266)
- VCS on other side is raising a `IllegalArgumentException` which is [converted to 500](https://github.com/nuxeo/nuxeo/pull/4606/files#diff-1723bb85f2e3ac2876b96b87404ede9ffee1f4d9747e576e0cfdb80bfee5b1bbR170) by myself. That's why I pointed you initially to the `SessionImpl` and I did the changes at its level. which I agree has more impact. 

Perhaps, I am wrong, but for me we have to make more consistency on low level VSC vs DBS and the raising exception. Otherwise we can rework [Here](https://github.com/nuxeo/nuxeo/pull/4606/files#diff-1723bb85f2e3ac2876b96b87404ede9ffee1f4d9747e576e0cfdb80bfee5b1bbR170-R173) to turn into the one(s) that we want:

- `IllegalArgumentException` into `DocumentNotFoundException` => both of them will be `404`
- We consider `IllegalArgumentException` and `DocumentNotFoundException` as `400`
 
for me the two options make sens as `ecm:path = 'non/existing/path'` can be considered as not exiting or a bad request.

efge: Ok but fixing all this is too much for this PR which is about error handling in the async adapter. Here turning `IllegalArgumentException` into a 400 in the catch clause is enough and makes sense semantically. We'll have 400 vs 404 for VCS/DBS but I think it's good enough for a client. Unifying this at low level is another topic.

RSalem07: I made more tests and it seems that for DBS (mongo) if we change the query to `nxql = "SELECT * FROM Document WHERE ecm:path = '/non/existing/path'";` I mean adding `/` at the beginning of `ecm:path` predicate. We don't have the error at all, because it returns `null` [Here](https://github.com/nuxeo/nuxeo/blob/master/modules/core/nuxeo-core-storage-dbs/src/main/java/org/nuxeo/ecm/core/storage/dbs/DBSSession.java#L328)

RSalem07: sorry, perhaps I am missing things. But https://github.com/nuxeo/nuxeo/pull/4606#discussion_r598933252 will not be enough, we have to make consistency between both of `IllegalArgumentException` and `DocumentNotFoundException ` as DBS is returning `DocumentNotFoundException`.  Otherwise my test will not pass. If want to remove the original check that I have added. Can I convert `IllegalArgumentException` to 404. Will make consistency between the two types (VCS and DBS).

efge: Turning `IllegalArgumentException` to 404 is wrong. However we could very well convert `DocumentNotFoundException` to a 400 here in this `catch` block, I think the semantics still work. @nuxeo/platform any comment?

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/AsyncStatus.java
```diff
@@ -38,4 +39,34 @@
...
      */
     boolean isCompleted();
+
+    /**
+     * Gets the error message if any.
+     *
+     * @return the error message if any, {@code null} otherwise
+     * @since 11.5
+     */
+    default String getError() {
```
kevinleturc, line 49: I would prefer a `getErrorMessage` to have the same method name as `BulkStatus` and avoid a bridge method.

efge (approved):

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
```diff
@@ -284,17 +286,38 @@ protected void updateStatusAsScrolling(ComputationContext context, String comman
...
 
+    /**
+     * @deprecated since 11.5, use {@link #updateStatusAfterScroll(ComputationContext, String, String, int)} instead
+     **/
+    @Deprecated(since = "11.5", forRemoval = true)
     protected void updateStatusAfterScroll(ComputationContext context, String commandId, String errorMessage) {
-        updateStatusAfterScroll(context, commandId, 0, errorMessage, false);
+        int errorCode = StringUtils.isEmpty(errorMessage) ? 0 : HttpStatus.SC_INTERNAL_SERVER_ERROR;
+        updateStatusAfterScroll(context, commandId, errorMessage, errorCode);
+    }
```
kevinleturc, line 296: I think deprecation is not needed there, such APIs were already modified when we introduce the `limited` parameter.

modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/AsyncStatus.java
```diff
@@ -20,11 +20,12 @@
 
 import java.io.Serializable;
 
+import org.apache.http.HttpStatus;
```
troger, line 23: Should use `HttpServletResponse` 🙏

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
```diff
@@ -284,17 +286,38 @@ protected void updateStatusAsScrolling(ComputationContext context, String comman
...
+     **/
+    @Deprecated(since = "11.5", forRemoval = true)
     protected void updateStatusAfterScroll(ComputationContext context, String commandId, long documentCount,
             String errorMessage, boolean limited) {
+        int errorCode = StringUtils.isEmpty(errorMessage) ? 0 : HttpStatus.SC_INTERNAL_SERVER_ERROR;
+        updateStatusAfterScroll(context, commandId, documentCount, errorMessage, errorCode, limited);
+    }
+
+    protected void updateStatusAfterScroll(ComputationContext context, String commandId, long documentCount,
+            String errorMessage, int errorCode, boolean limited) {
```
kevinleturc, line 320: `errorMessage` and `errorCode` should be both provided or not, we should remove cases where we can have one or the other.

As it, we could use `BulkStatus#inError(String, int)` and remove `BulkStatus#setErrorCode`.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
```diff
@@ -82,6 +84,9 @@
     @Nullable
     protected String errorMessage;
 
+    /** @since 11.5 **/
+    protected int errorCode;
```
kevinleturc, line 88: For old BulkStatus presents in streams, this field should be an Integer and have a `@Nullable` annotation.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
```diff
@@ -212,6 +217,9 @@ public void merge(@NotNull BulkStatus update) {
         if (update.errorMessage != null && errorMessage == null) {
             errorMessage = update.errorMessage;
         }
+        if (update.getErrorCode() != 0) {
```
kevinleturc, line 220: Once the field will be nullable, we should have:
```suggestion
        if (update.errorCode != null && errorCode == null) {
```

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
```diff
@@ -435,6 +443,11 @@ public void setProcessingDurationMillis(long processingDurationMillis) {
         this.processingDurationMillis = processingDurationMillis;
     }
 
+    /** @since 11.5 */
+    public void setErrorCode(int errorCode) {
+        this.errorCode = errorCode;
+    }
+
     public boolean hasError() {
```
kevinleturc, line 451: Could you add a `@Override` to this method?

kevinleturc (changes requested):

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
kevinleturc: It should be fine as it seems to be an edge case for me.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
RSalem07: > errorMessage and errorCode should be both provided or not, we should remove cases where we can have one or the other.

Yep, for now I removed them. Is it ok for you. let me know wdyt

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
```diff
@@ -435,6 +444,12 @@ public void setProcessingDurationMillis(long processingDurationMillis) {
         this.processingDurationMillis = processingDurationMillis;
     }
 
+    /** @since 11.5 */
+    public void setErrorCode(Integer errorCode) {
+        this.errorCode = errorCode;
+    }
```
kevinleturc, line 450: Should be removed.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
```diff
@@ -284,20 +286,21 @@ protected void updateStatusAsScrolling(ComputationContext context, String comman
...
+        updateStatusAfterScroll(context, commandId, documentCount, null, 0, limited);
     }
 
     protected void updateStatusAfterScroll(ComputationContext context, String commandId, long documentCount,
-            String errorMessage, boolean limited) {
+            String errorMessage, int errorCode, boolean limited) {
         BulkStatus delta = BulkStatus.deltaOf(commandId);
         if (errorMessage != null) {
-            delta.inError(errorMessage);
+            delta.inError(errorMessage, errorCode);
```
kevinleturc, line 303: I would say that errorCode could be an `Integer`, and call the inError method as below:
```java
delta.inError(errorMessage, requireNonNullElse(errorCode, SC_INTERNAL_SERVER_ERROR));
```
What do you think?

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/message/BulkStatus.java
kevinleturc: We should access the field directly like for `errorMessage`.

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/computation/BulkScrollerComputation.java
```diff
@@ -284,20 +287,21 @@ protected void updateStatusAsScrolling(ComputationContext context, String comman
...
-        updateStatusAfterScroll(context, commandId, 0, errorMessage, false);
+    protected void updateStatusAfterScroll(ComputationContext context, String commandId, String errorMessage,
+            Integer errorCode) {
+        updateStatusAfterScroll(context, commandId, 0, errorMessage, errorCode, false);
     }
 
     protected void updateStatusAfterScroll(ComputationContext context, String commandId, long documentCount,
             boolean limited) {
-        updateStatusAfterScroll(context, commandId, documentCount, null, limited);
+        updateStatusAfterScroll(context, commandId, documentCount, null, 0, limited);
```
kevinleturc, line 297:
```suggestion
        updateStatusAfterScroll(context, commandId, documentCount, null, null, limited);
```

RSalem07: yep, thanks Kevin. I completely forgot it.

kevinleturc: No problem 😉

kevinleturc (approved):

ataillefer (approved):

PR #4607, opened 2021-01-26
NXP-29984: Runtime Registries Migrations #5

atchertchian:


modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/service/FilterConfigDescriptor.java
```diff
@@ -34,11 +37,13 @@
...
  */
 @XObject(value = "filterConfig")
+@XRegistry(compatWarnOnMerge = true)
 public class FilterConfigDescriptor {
 
     public static final String DEFAULT_CACHE_DURATION = "3599";
 
     @XNode("@name")
+    @XRegistryId
     protected String name;
```
efge, line 47: `getName()` defaults to the pattern so we have to be careful here maybe (unless it's a feature nobody uses)

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/service/FilterConfigDescriptor.java
```diff
@@ -113,7 +104,7 @@ public boolean isPrivate() {
     }
 
     public String getCacheTime() {
-        if (cacheTime == null || cacheTime.equals("")) {
+        if (StringUtils.isBlank(cacheTime)) {
             cacheTime = DEFAULT_CACHE_DURATION;
```
efge, line 108: Can't this be put on the field itself now? I thought your changes allowed that.

modules/runtime/nuxeo-runtime-management/src/main/java/org/nuxeo/runtime/management/ResourceFactoryDescriptor.java
```diff
@@ -20,23 +20,21 @@
...
 
-    public ResourceFactoryDescriptor(Class<? extends ResourceFactory> factoryClass) {
-        this.factoryClass = factoryClass;
-    }
-
-    public ResourceFactoryDescriptor() {
-    }
-
+    /** @since 11.5 */
     public Class<? extends ResourceFactory> getFactoryClass() {
```
efge, line 38: This isn't a new method, the `@since` is wrong.

modules/platform/nuxeo-target-platforms/nuxeo-target-platforms-core/src/main/java/org/nuxeo/targetplatforms/core/service/TargetPlatformServiceImpl.java
```diff
@@ -76,60 +76,21 @@
...
 
     @Override
-    public void activate(ComponentContext context) {
-        platforms = new TargetPlatformRegistry();
-        packages = new TargetPackageRegistry();
-        conf = new ServiceConfigurationRegistry();
+    public void start(ComponentContext context) {
+        overrideDirectory = this.<ServiceConfigurationDescriptor> getRegistryContribution(XP_CONF)
+                                .map(ServiceConfigurationDescriptor::getOverrideDirectory)
+                                .filter(Predicate.not(StringUtils::isBlank))
```
efge, line 87: `.filter(StringUtils::isNotBlank)`

modules/platform/nuxeo-target-platforms/nuxeo-target-platforms-core/src/main/java/org/nuxeo/targetplatforms/core/descriptors/TargetDescriptor.java
```diff
@@ -39,13 +41,15 @@
...
+    private static final Logger log = LogManager.getLogger(TargetDescriptor.class);
 
     @XNode("@id")
+    @XRegistryId
     String id;
 
-    @XNode("@enabled")
-    Boolean enabled;
+    @XNode(value = XEnable.ENABLE, fallback = "@enabled", defaultAssignment = "true")
+    // registry needs to list enabled descriptors
```
efge, line 51: Clarify the comment: `// no @XEnable because the registry needs to know which descriptors are enabled or not`

modules/platform/nuxeo-platform-task/nuxeo-platform-task-api/src/main/java/org/nuxeo/ecm/platform/task/TaskProvider.java
```diff
@@ -31,7 +30,7 @@
  * @author Laurent Doguin
  * @since 5.5
  */
-public interface TaskProvider extends Serializable {
+public interface TaskProvider {
```
efge, line 33: I think I had an issue somewhere last time I tried to remove the `Serializable` in these descriptors but I don't remember what... Maybe JSF?

modules/platform/nuxeo-platform-task/nuxeo-platform-task-api/src/main/java/org/nuxeo/ecm/platform/task/TaskService.java
```diff
@@ -35,13 +35,20 @@
...
      * Property used to pass task in the notified events properties
      */
     String TASK_INSTANCE_EVENT_PROPERTIES_KEY = "taskInstance";
 
+    /**
+     * Default task root path.
+     *
+     * @since 11.5
+     */
+    static final String DEFAULT_PARENT_PATH = "/task-root";
```
efge, line 50: `static final` is already implicit in interfaces.

modules/platform/nuxeo-invite/src/main/java/org/nuxeo/ecm/user/invite/UserInvitationComponent.java
```diff
@@ -782,23 +762,20 @@ public UserRegistrationConfiguration getConfiguration(DocumentModel requestDoc)
...
             }
-
-            if (!configurations.containsKey(configurationName)) {
-                throw new NuxeoException("Configuration " + configurationName + " is not registered");
-            }
-            return configurations.get(configurationName);
+            return getConfiguration(configurationName);
         } catch (NuxeoException e) {
-            log.info("Unable to get request parent document: " + e.getMessage());
+            log.info("Unable to get request parent document: {}", e.getMessage());
```
efge, line 767: `e::getMessage`

modules/platform/nuxeo-invite/src/main/java/org/nuxeo/ecm/user/invite/UserInvitationComponent.java
```diff
@@ -782,23 +762,20 @@ public UserRegistrationConfiguration getConfiguration(DocumentModel requestDoc)
...
     @Override
     public UserRegistrationConfiguration getConfiguration(String name) {
-        if (!configurations.containsKey(name)) {
-            throw new NuxeoException("Trying to get unknown user registration configuration.");
+        Optional<UserRegistrationConfiguration> conf = getRegistryContribution(XP, name);
+        if (conf.isEmpty()) {
+            throw new NuxeoException("Configuration " + name + " is not registered");
         }
-        return configurations.get(name);
+        return conf.get();
```
efge, line 778: Could use a simple `orElseThrow`

efge (changes requested):

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/service/FilterConfigDescriptor.java
atchertchian: Not sure it's used but thanks for catching this, i added pattern as a fallback on the annotation directly :)

modules/platform/nuxeo-platform-web-common/src/main/java/org/nuxeo/ecm/platform/web/common/requestcontroller/service/FilterConfigDescriptor.java
atchertchian: Indeed, and there's a "trim" attribute too so no need for the check :+1:

modules/platform/nuxeo-platform-task/nuxeo-platform-task-api/src/main/java/org/nuxeo/ecm/platform/task/TaskProvider.java
atchertchian: I don't see why, although i'll check if all these changes have any impact on JSF.
Maybe a confusion with Seam beans in conversation scope, that need to be Serializable?

modules/platform/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/service/TemplateProcessorComponent.java
```diff
@@ -173,58 +138,44 @@ public void addContextExtensions(DocumentModel currentDocument, DocumentWrapper
...
+                                                              .findFirst();
+        if (res.isEmpty() && !candidates.isEmpty()) {
+            res = Optional.of(candidates.get(0));
         }
-        if (candidates.size() > 0) {
-            return candidates.get(0);
-        }
-        return null;
+        return res;
     }
```
kevinleturc, line 167: What do think about sorting the stream just after giving the `predicate`, the boolean natural order is _false_ then _true_, we will end with something like:
```
return this.<TemplateProcessorDescriptor> getRegistryContributions(PROCESSOR_XP)
           .stream()
           .filter(predicate)
           .sorted(Comparator.comparing(TemplateProcessorDescriptor::isDefaultProcessor).reversed()) // true first
           .findFirst();
```

kevinleturc (approved):
Great work 👍 

Some details

modules/platform/nuxeo-platform-audit/nuxeo-platform-audit-core/src/main/resources/OSGI-INF/nxaudit-service.xml
```diff
@@ -127,7 +127,6 @@
     <event name="documentLocked" />
     <event name="documentUnlocked" />
     <event name="documentSecurityUpdated" />
-    <event name="retentionActiveChanged" />
```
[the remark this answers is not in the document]
atchertchian: I modified the commit message: it's already there, with extendedInfos, on another contribution in the same module: https://github.com/nuxeo/nuxeo/blob/master/modules/platform/nuxeo-platform-audit/nuxeo-platform-audit-core/src/main/resources/OSGI-INF/user-audit-contrib.xml#L40

modules/platform/nuxeo-template-rendering/nuxeo-template-rendering-core/src/main/java/org/nuxeo/template/service/TemplateProcessorComponent.java
atchertchian: Thanks, i wondered how to even more simplify this. The only potential difference is that original order of contributions is not respected, i don't know if this should taken into account.

atchertchian: Let's see how tests behave with this change: maybe this edge case should not happen in practice (?)

kevinleturc: As far as I know, the sort won't change the original order of contributions having the same value for `isDefaultProcessor` 🤔. That's an interesting interrogation, not sure we need to respect it as we don't (yet) provide any ordering mechanism. Wdyt?

atchertchian: Well we do provide some kind of ordering mechanism, we could imagine a component requiring the other just to make sure its contributions are registered after.
Anyway this would not change the fact that first is taken in any case here, so if the sort does not change the original order for those, we should be good.

efge (approved):

kevinleturc (approved):

PR #4613, opened 2021-01-29
NXP-30045: Align Mongo maxTime with transaction timeout

bdelbosc:


modules/runtime/nuxeo-runtime-jtajca/src/main/java/org/nuxeo/runtime/transaction/TransactionHelper.java
```diff
@@ -216,6 +216,36 @@ public static boolean isTransactionTimedOut() {
...
+            return -1;
+        }
+        try {
+            Transaction tx = tm.getTransaction();
+            if (tx == null || tx.getStatus() != Status.STATUS_ACTIVE) {
+                return -1;
+            }
+            if (tx instanceof org.apache.geronimo.transaction.manager.TransactionImpl) {
+                // Geronimo Transaction Manager
+                return (Long) GERONIMO_TRANSACTION_TIMEOUT_FIELD.get(tx);
```
efge, line 238: You must substract `System.currentTimeMillis()` to get a TTL, as Geronimo stores the expiration time of the transaction.
Also we could normalize zero and negative values (tx expired) to `0`. Behavior when tx is expired should be documented in the Javadoc.

modules/runtime/nuxeo-runtime-jtajca/src/main/java/org/nuxeo/runtime/transaction/TransactionHelper.java
```diff
@@ -216,6 +216,36 @@ public static boolean isTransactionTimedOut() {
         }
     }
 
+
```
efge, line 219: Remove extra empty line before and after the method.

modules/runtime/nuxeo-runtime-jtajca/src/main/java/org/nuxeo/runtime/transaction/TransactionHelper.java
```diff
@@ -216,6 +216,36 @@ public static boolean isTransactionTimedOut() {
...
+    /**
+     * Gets the transaction TTL.
+     *
+     * @return the time to live in millis or {@code -1} when outside of a transaction.
+     * @since 11.5
+     */
+    public static long getTransactionTimeToLiveMs() {
```
efge, line 226: `startTransaction(int)` uses seconds as a unit (because that's what `TransactionManager.setTransactionTimeout` uses) so for consistency I would prefer if the TTL was returned in seconds, and name this method just `getTransactionTimeToLive`.

efge (changes requested):

modules/runtime/nuxeo-runtime-jtajca/src/main/java/org/nuxeo/runtime/transaction/TransactionHelper.java
```diff
@@ -216,6 +216,36 @@ public static boolean isTransactionTimedOut() {
...
+        TransactionManager tm = NuxeoContainer.getTransactionManager();
+        if (tm == null) {
+            return -1;
+        }
+        try {
+            Transaction tx = tm.getTransaction();
+            if (tx == null || tx.getStatus() != Status.STATUS_ACTIVE) {
```
efge, line 233: `STATUS_MARKED_ROLLBACK` should be allowed too, it's still in a transaction.

modules/runtime/nuxeo-runtime-jtajca/src/main/java/org/nuxeo/runtime/transaction/TransactionHelper.java
bdelbosc: yes but this is a different usage and ms could be important

efge: I'm not sure why milliseconds are important.
Note that even in Geronimo's `TransactionImpl` the `commit()` transaction timeout check is implemented with ` if (TransactionTimer.getCurrentTime() > timeout)` and the current time in `TransactionTimer` is updated only every second, there's no finer granularity.
So we have to consider all timeouts and TTLs as an approximation up to the second.

bdelbosc: it matters because in one second we could run thousands of queries and rounding 499ms to 0 can be a problem.
Now if geronimo has a second granularity I will see that next week.

efge: But all this happens only if we're at the last second of the original transaction timeout anyway, so if it fails half a second earlier or later it doesn't matter.

kevinleturc (approved):

efge (approved):

PR #4614, opened 2021-01-29
NXP-29984: Runtime Registries Migrations #6

atchertchian:


modules/platform/nuxeo-platform-mail/src/main/java/org/nuxeo/ecm/platform/mail/action/MessageActionPipeRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+                    "Usage of the \"override\" attribute on MessageActionPipeDescriptor contribution '%s',"
+                            + " in extension '%s', is deprecated: use the \"merge\" attribute instead.",
+                    id, extensionId);
+            DeprecationLogger.log(message, "11.5");
+            return false;
+        }
+        boolean shouldMerge = super.shouldMerge(ctx, xObject, element, extensionId, id, existing);
+        if (existing != null && !shouldMerge) {
+
+        }
```
kevinleturc, line 50: Empty?

modules/platform/nuxeo-platform-query-api/src/main/java/org/nuxeo/ecm/platform/query/core/PageProviderServiceImpl.java
```diff
@@ -44,31 +49,104 @@
...
+
+    @Override
+    public void activate(ComponentContext context) {
+        programmaticDefinitions = new ConcurrentHashMap<>();
+    }
+
+    @Override
+    public void deactivate(ComponentContext context) {
+        programmaticDefinitions = null;
+    }
```
kevinleturc, line 74: Could it be placed after stop method? I prefer to follow the order of runtime steps, not a strong opinion, wdyt?

modules/platform/nuxeo-platform-query-api/src/main/java/org/nuxeo/ecm/platform/query/core/PageProviderServiceImpl.java
```diff
@@ -44,31 +49,104 @@
...
+        // returns clones as definition API accepts setters
+        PageProviderDefinition def = programmaticDefinitions.get(name);
+        if (def != null) {
+            return Optional.of(def.clone());
+        }
+        Optional<PageProviderDefinition> optDef = getRegistryContribution(PROVIDER_EP, name);
+        if (optDef.isPresent()) {
+            return Optional.of(optDef.get().clone());
+        }
+        return Optional.empty();
```
kevinleturc, line 136:
```suggestion
        return Optional.ofNullable(programmaticDefinitions.get(name))
                       .or(() -> getRegistryContribution(PROVIDER_EP, name))
                       .map(PageProviderDefinition::clone);
```

modules/platform/nuxeo-platform-query-api/src/main/java/org/nuxeo/ecm/platform/query/core/PageProviderServiceImpl.java
```diff
@@ -209,13 +286,15 @@ public PageProviderDefinition getPageProviderDefinition(String name) {
         return ret;
     }
 
+    protected PageProviderDefinition getPageProviderDefinitionOrElseThrow(String name) {
+        return getOptionalPageProviderDefinition(name).orElseThrow(
+                () -> new NuxeoException(String.format("Could not resolve page provider with name '%s'", name)));
+    }
+
```
kevinleturc, line 293: Could it be put next to `getOptionalPageProviderDefinition`?

modules/platform/nuxeo-platform-video/src/main/java/org/nuxeo/ecm/platform/video/service/VideoServiceImpl.java
```diff
@@ -161,11 +123,11 @@ public void launchAutomaticConversions(DocumentModel doc, boolean onlyMissing) {
 
     @Override
     public TranscodedVideo convert(Video originalVideo, String conversionName) {
-        if (!videoConversions.registry.containsKey(conversionName)) {
-            throw new NuxeoException(String.format("'%s' is not a registered video conversion.", conversionName));
-        }
+        VideoConversion conversion = this.<VideoConversion> getRegistryContribution(VIDEO_CONVERSIONS_EP,
```
kevinleturc, line 126: Maybe `var`?

modules/platform/nuxeo-web-resources-core/src/main/java/org/nuxeo/ecm/web/resources/core/service/WebResourceManagerImpl.java
```diff
@@ -36,142 +35,70 @@
...
     public List<Processor> getProcessors(String type) {
-        return processors.getProcessors(type);
+        List<Processor> res = new ArrayList<>();
+        for (Processor proc : getProcessors()) {
+            if (type == null || proc.getTypes().contains(type)) {
+                res.add(proc);
+            }
+        }
+        Collections.sort(res);
+        return res;
```
kevinleturc, line 95: Could be:
```
return getProcessor().stream().filter(p -> type ==  null || p.getTypes().contains(type)).sorted().collect(Collectors.toList());
```

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/MapRegistry.java
```diff
@@ -92,40 +92,90 @@ protected String computeId(Context ctx, XAnnotatedObject xObject, Element elemen
...
+            return false;
+        }
+        // checks no attribute other than id and enablement
+        int nbAttr = element.getAttributes().getLength();
+        if (nbAttr > 2) {
+            return false;
+        }
+        XAnnotatedMember enable = xObject.getEnable();
+        return (enable != null && enable.hasValue(ctx, element));
+    }
```
kevinleturc, line 133: Can't it be put in the Abstract? Maybe other methods are concerned too.

modules/core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/StreamWorkManager.java
```diff
@@ -296,45 +295,41 @@ public void init() {
...
-            }
-        }
+                @Override
+                public void afterRuntimeStart(ComponentManager mgr, boolean isResume) {
+                    if (isProcessingDisabled()) {
+                        log.warn("WorkManager processing has been disabled on this node");
+                        return;
+                    }
+                    streamProcessor.start();
+                    getWorkQueueIds().forEach(id -> activateQueueMetrics(id));
```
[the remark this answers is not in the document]
atchertchian: The fact that it's inside the `new ComponentManager.Listener` was making that behavior buggy: happy to know if there's a way to reference the "super" method

modules/core/nuxeo-core-management/src/main/java/org/nuxeo/ecm/core/management/CoreManagementComponent.java
```diff
@@ -56,20 +55,65 @@
...
+    @Override
+    public void deactivate(ComponentContext context) {
+        EventStatsHolder.clearStats();
+    }
+
+    @Override
+    public void start(ComponentContext context) {
+        globalManager = new GlobalAdministrativeStatusManagerImpl();
+        this.<AdministrableServiceDescriptor> getRegistryContributions(SERVICE_DEF_EP)
+            .forEach(globalManager::registerService);
```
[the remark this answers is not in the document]
atchertchian: I agree, i've removed register API on services as much as I could, but the pattern was too recurrent here, and I was indeed wondering if those could be registered "on the fly"...

modules/core/nuxeo-core-query/src/main/java/org/nuxeo/ecm/core/query/sql/parser/parser.java
```diff
@@ -8,7 +8,6 @@
 
 import java_cup.runtime.*;
 import org.nuxeo.ecm.core.query.sql.model.*;
-import org.nuxeo.ecm.core.query.sql.*;
```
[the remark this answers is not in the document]
atchertchian: It looks like it was generated, I saw the "unused imports" warn so cleaned it up but I don't need to, I will revert the change.

modules/core/nuxeo-platform-commandline-executor/src/main/java/org/nuxeo/ecm/platform/commandline/executor/service/CommandLineExecutorComponent.java
```diff
@@ -166,58 +151,52 @@ public ExecResult execCommand(String commandName, CmdParameters params) throws C
...
-
-        CommandLineDescriptor cmdDesc = commandDescriptors.get(commandName);
-        Executor executor = executors.get(cmdDesc.getExecutor());
-        EnvironmentDescriptor environment = new EnvironmentDescriptor().merge(env).merge(
-                envDescriptors.getOrDefault(commandName, envDescriptors.get(cmdDesc.getCommand())));
-        return executor.exec(cmdDesc, params, environment);
+        CommandLineDescriptor cmdDesc = this.<CommandLineDescriptor> getRegistryContribution(EP_CMD, commandName)
+                                            // should not happen
+                                            .orElseThrow(() -> new RuntimeException(
+                                                    "Command " + commandName + " is not be available"));
```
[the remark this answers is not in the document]
atchertchian: :+1: (not my code :D)

modules/core/nuxeo-platform-commandline-executor/src/main/java/org/nuxeo/ecm/platform/commandline/executor/service/CommandLineExecutorComponent.java
```diff
@@ -166,58 +151,52 @@ public ExecResult execCommand(String commandName, CmdParameters params) throws C
...
+        EnvironmentDescriptor globalEnv = getEnvironmentDescriptor("");
+        EnvironmentDescriptor commandEnv = getEnvironmentDescriptor(commandName);
+        if (commandEnv == null) {
+            commandEnv = getEnvironmentDescriptor(cmdDesc.getCommand());
+        }
```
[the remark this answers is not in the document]
atchertchian: I'll have another look, this one was not the easiest to migrate so maybe I gave up on some simplifications.

modules/core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/StreamWorkManager.java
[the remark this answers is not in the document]
kevinleturc: We could use `StreamWorkManager.this::activateQueueMetrics` but as it is longer than the lambda I prefer your solution.

modules/core/nuxeo-platform-commandline-executor/src/main/java/org/nuxeo/ecm/platform/commandline/executor/service/EnvironmentDescriptor.java
```diff
@@ -80,4 +74,27 @@ public String getName() {
...
+        }
+        if (other != null) {
+            res.name = other.name;
+            if (other.workingDirectory != null) {
+                res.workingDirectory = other.workingDirectory;
+            }
+            res.parameters.putAll(other.getParameters());
+        }
+        return res;
+    }
```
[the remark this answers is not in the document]
atchertchian: I ended up doing this because i removed the "merge" method (that was impacting existing descriptor instances), I'll have a look at what you suggest but note there was no unit test, so I hope this will not break the expected behaviour.

modules/platform/nuxeo-platform-mail/src/main/java/org/nuxeo/ecm/platform/mail/action/MessageActionPipeRegistry.java
atchertchian: I went back and forth with the fix, this clearly needs to be cleaned up, but i'll make another change here anyway (will point out the new commit once done)

modules/platform/nuxeo-platform-query-api/src/main/java/org/nuxeo/ecm/platform/query/core/PageProviderServiceImpl.java
atchertchian: At this point of migrations, clearly not a strong opinion on my end for that kind of refinement :D
I placed it after the activation because it makes it easier to check that it "undoes" what activation does, I can change it if you prefer but I might have followed this pattern in a lot of other cases already.

modules/platform/nuxeo-platform-query-api/src/main/java/org/nuxeo/ecm/platform/query/core/PageProviderServiceImpl.java
atchertchian: :+1: I'm not accustomed to using or()

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/MapRegistry.java
atchertchian: Actually i need to fix this: the implementation should differ between map and single (single registry should not expect an id attribute).
Other similar methods also have that constraint (need to extract the contribution and maybe expose the item id), but I'll have another look to see if some of this can be factored out.

modules/core/nuxeo-platform-commandline-executor/src/main/java/org/nuxeo/ecm/platform/commandline/executor/service/EnvironmentDescriptor.java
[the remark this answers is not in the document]
atchertchian: I added back the old merge method

modules/platform/nuxeo-platform-mail/src/main/java/org/nuxeo/ecm/platform/mail/action/MessageActionPipeRegistry.java
atchertchian: (cleanup done)

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/MapRegistry.java
atchertchian: I factored out the shouldEnable and shouldRemove methods in the end, kept the merge methods apart (as the compat warn message needs the contribution id anyway, and the "onlyHandlesEnablement" method should be specific as explained above.

atchertchian:
Handled feedback, also added a4a99852f74b667aa62e0dd382c36274b4c46447 (as it is going to be useful for further migrations), this impacted changes on MailService registry at ca2c8160ca76aaf1884fdb417c01195c24901c6c (when handling the "override" boolean on descriptor + associated deprecation warm)

modules/platform/nuxeo-platform-query-api/src/main/java/org/nuxeo/ecm/platform/query/core/PageProviderServiceImpl.java
kevinleturc: We can keep the current code.

atchertchian:
Removed previous ("reverse fallback") boolean behavior, not as useful as i hoped, and reviewed the default registries API again to help with further migrations when writing custom registries.
Commits changed: f257e2a52510a5395548a1bc3a411b103f1de842 and ff5da18de67443ed5e9f3180dd17b42b695fa5e8.
This impacted only MessageActionPipeRegistry after that: https://github.com/nuxeo/nuxeo/pull/4614/commits/9f59a4c0e73c4354d15c5f340fc9b3cc3e5fc873#diff-7345de20cca3a4be514e1aea2b6e750bc91d88d9bf796586a0de7dc33863c490R32

modules/platform/nuxeo-platform-mail/src/main/java/org/nuxeo/ecm/platform/mail/action/MessageActionPipeDescriptor.java
```diff
@@ -15,102 +15,105 @@
...
-     */
-    public void merge(MessageActionPipeDescriptor descriptor) {
-        for (String actionName : descriptor.getActions().keySet()) {
-            actionDescriptorsRegistry.put(actionName, descriptor.getActions().get(actionName));
-        }
-        pipe = new MessageActionPipe();
+    public Map<String, MessageActionDescriptor> getActionsMap() {
+        return Stream.of(actions)
+                     .collect(Collectors.toMap(MessageActionDescriptor::getId, Function.identity(), (e1, e2) -> e2,
+                             LinkedHashMap::new));
```
efge, line 76: We have lots of those Map collectors now, maybe a good pattern would be to define static methods in the corresponding descriptors like:

    public static final Collector<...> COLLECTOR_TO_ID_DESC_MAP = Collectors.toMap(MessageActionDescriptor::getId, Function.identity(),
               (e1, e2) -> e2, LinkedHashMap::new);

modules/platform/nuxeo-platform-mail/src/main/java/org/nuxeo/ecm/platform/mail/service/MailServiceImpl.java
```diff
@@ -85,46 +73,37 @@
...
-        }
+    public void start(ComponentContext context) {
+        configuredFetchers = new HashMap<>();
+        this.<SessionFactoryDescriptor> getRegistryContributions(SESSION_FACTORY).forEach(descriptor -> {
+            String fetcherName = descriptor.getFetcherName();
+            Class<? extends PropertiesFetcher> clazz = this.<PropertiesFetcherDescriptor> getRegistryContribution(
+                    PROPERTIES_FETCHER, fetcherName).map(PropertiesFetcherDescriptor::getFetcher).orElse(null);
+            if (clazz == null) {
+                addRuntimeMessage(Level.ERROR, String.format("Unknown properties fetcher: %s", fetcherName));
+            } else {
```
efge, line 84: Could add a `return` here after the error case and deindent the `else` block.

modules/platform/nuxeo-platform-mail/src/main/java/org/nuxeo/ecm/platform/mail/service/MailServiceImpl.java
```diff
@@ -85,46 +73,37 @@
...
+                PropertiesFetcher fetcher;
+                try {
+                    fetcher = clazz.getDeclaredConstructor().newInstance();
```
efge, line 87: `fetcher` declaration can be moved down into the `try`

modules/core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/StreamWorkManager.java
```diff
@@ -296,45 +295,41 @@ public void init() {
...
-    }
 
-    class ComponentListener implements ComponentManager.Listener {
-        @Override
-        public void beforeStop(ComponentManager mgr, boolean isStandby) {
-            if (!shutdown(10, TimeUnit.SECONDS)) {
-                log.error("Some processors are still active");
-            }
-        }
+            Framework.getRuntime().getComponentManager().addListener(new ComponentManager.Listener() {
```
efge, line 307: Could you add a comment explaining why it's not possible to make the component itself implement `ComponentManager.Listener`?

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/AbstractRegistry.java
```diff
@@ -108,4 +109,41 @@ public void unregister(String tag) {
...
+
+    protected boolean shouldMerge(Context ctx, XAnnotatedObject xObject, Element element, String extensionId) {
+        XAnnotatedMember merge = xObject.getMerge();
+        return merge != null && Boolean.TRUE.equals(merge.getValue(ctx, element));
+    }
+
+    @SuppressWarnings("unchecked")
+    protected <T> T getMergedInstance(Context ctx, XAnnotatedObject xObject, Element element, Object existing) {
+        return (T) xObject.newInstance(ctx, element, existing);
+    }
```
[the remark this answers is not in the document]
atchertchian: Indeed, but xmap is used in other cases than runtime contributions so I'd rather de-correlate this change: maybe create a dedicated Jira isue?

modules/core/nuxeo-platform-commandline-executor/src/main/java/org/nuxeo/ecm/platform/commandline/executor/service/CommandLineExecutorComponent.java
```diff
@@ -166,58 +164,47 @@ public ExecResult execCommand(String commandName, CmdParameters params) throws C
...
-            return new CommandAvailability();
-        } else {
-            return new CommandAvailability(desc.getInstallationDirective(), desc.getInstallErrorMessage());
+        CommandAvailability avail = unavailableCommands.get(commandName);
+        if (avail != null) {
+            return avail;
         }
+        return this.<CommandLineDescriptor> getRegistryContribution(EP_CMD, commandName)
+                   .map(desc -> new CommandAvailability())
+                   .orElse(new CommandAvailability(commandName + " is not a registered command"));
```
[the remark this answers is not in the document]
atchertchian: I agree this old code is quite complicated, and migrating it gave me quite a headache.
The only reason the service stores unavailable commands it to preserve the install message held by descriptors.
I thought it was valuable to keep that information (which is tested), although it may not be used by callers, but since it is wrapped by the CommandNotAvailable exception` #getMessage` method, it is hard to check whether some caller code is relying on this information to be provided...

modules/platform/nuxeo-platform-oauth/src/main/java/org/nuxeo/ecm/platform/oauth2/providers/OAuth2ServiceProviderRegistryImpl.java
```diff
@@ -21,65 +21,68 @@
...
+        persistProviders();
+    }
 
-            List<DocumentModel> providers = queryProviders(filter, 1);
-            return providers.isEmpty() ? null : providers.get(0);
-        } catch (DirectoryException e) {
-            log.error("Unable to read provider from Directory backend", e);
+    protected DocumentModel getPersistedProvider(String serviceName) {
+        if (StringUtils.isBlank(serviceName)) {
+            log.warn("Cannot fetch provider without a serviceName");
```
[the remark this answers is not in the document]
atchertchian: Maybe, but there are multiple places where the null provider is checked: sometimes null is indeed expected (sometimes not), I preserved the existing behavior to avoid getting into too much cleanup, please tell if you'd rather have me do this cleanup (which would avoid getting a NPE only in jsf-ui code).

modules/core/nuxeo-core-event/src/main/java/org/nuxeo/ecm/core/work/StreamWorkManager.java
atchertchian: I gave up on making that change in a preceding PR (see https://github.com/nuxeo/nuxeo/pull/4544#issuecomment-749534696). The fact that StreamWorkManager is supposed to replace the WorkManagerImpl component is highly unusual for the runtime, and my understanding is that works should be removed anyway so I did not dig further. I can still add this as a comment.

modules/platform/nuxeo-platform-oauth/src/main/java/org/nuxeo/ecm/platform/oauth2/providers/OAuth2ServiceProviderRegistryImpl.java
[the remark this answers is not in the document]
efge: We can do other cleanups later

modules/core/nuxeo-platform-commandline-executor/src/main/java/org/nuxeo/ecm/platform/commandline/executor/service/CommandLineExecutorComponent.java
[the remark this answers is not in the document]
efge: For later

modules/runtime/nuxeo-common/src/main/java/org/nuxeo/common/xmap/registry/AbstractRegistry.java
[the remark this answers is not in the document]
efge: Yes that was just an observation, not for immediate action

efge (approved):

kevinleturc (approved):

atchertchian:
Merging dismissing failure on ftests and docker image build + tests: validation went through ok, except random error log being handled with #4633

PR #4615, opened 2021-01-31
NXP-29516/NXP-30044: async blob digest computation (10.10)

efge:

PR #4625, opened 2021-02-03
NXP-29516/NXP-30044: async blob digest computation

efge:
Forward-port of #4615


efge:
(rebased)

efge:
(fixed one Javadoc issue + rebased)

efge:
Added one commit to allow simple config with `nuxeo.core.blobstore.digestAsync=true`

efge:
Added one commit to configure async digest simply.
Added two more commits to make path strategies accept more keys, and improve CachingBlobStore to better use path strategies (don't depend on LRUFileCache); with added caching-related tests.

efge:
Rebased + added commit for NXP-30168

PR #4629, opened 2021-02-04
NXP-30133: fix CSV Import tests on huge files taking too long

efge:


efge:
(force pushed to add `@since`)

troger (approved):

efge:
(rebased)

PR #4636, opened 2021-02-08
NXP-30140: upgrade keycloak

efge:


modules/platform/login/nuxeo-platform-login-keycloak/pom.xml
```diff
@@ -58,8 +55,8 @@
     <!-- #############################################################################  KEYCLOAK -->
     <dependency>
       <groupId>org.keycloak</groupId>
-      <artifactId>keycloak-tomcat8-adapter</artifactId>
-      <version>${keycloak.version}</version>
+      <artifactId>keycloak-tomcat-adapter</artifactId>
+      <version>12.0.2</version>
```
kevinleturc, line 59: Should we move the version declaration to the root pom?

kevinleturc (approved):

modules/platform/login/nuxeo-platform-login-keycloak/pom.xml
efge: I'll do that with a few others when we move to dependabot so that all deps are in the toplevel POM.

PR #4638, opened 2021-02-10
Fix NXP-30078 csv export multi line

kevinleturc:


modules/platform/nuxeo-platform-csv-export/src/main/java/org/nuxeo/ecm/platform/csv/export/io/DocumentPropertyCSVWriter.java
```diff
@@ -168,17 +174,25 @@ protected String getTranslatedValue(Session session, String schema, String lang,
...
     protected void writeUnsupported(Type type, CSVPrinter printer) throws IOException {
         printer.print(String.format("type %s is not supported", type.getName()));
     }
 
+    protected String replaceCarriageReturn(String value) {
+        String replacement = ctx.getParameter(CARRIAGE_RETURN_REPLACEMENT_CTX_DATA);
+        if (value == null || replacement == null) {
+            return value;
+        }
+        return value.replace(System.lineSeparator(), replacement);
```
efge, line 195: I'd prefer something more general as a file with one kind of newline may be treated on a platform with another kind.

    protected static final Pattern NEWLINE = Pattern.compile("\\R");

    NEWLINE.matcher(value).replaceAll(replacement);

efge (changes requested):

modules/platform/nuxeo-platform-csv-export/src/main/java/org/nuxeo/ecm/platform/csv/export/io/DocumentPropertyCSVWriter.java
kevinleturc: Thanks for the code 👍

efge (approved):

troger (dismissed):

efge (dismissed):

kevinleturc:
I had to force push because the test uses `$` as carriage return replacement and as we're now using a pattern matcher java tries to interpret it as a regex group.

modules/platform/nuxeo-platform-csv-export/src/main/java/org/nuxeo/ecm/platform/csv/export/io/DocumentPropertyCSVWriter.java
```diff
@@ -167,19 +177,26 @@ protected String getTranslatedValue(Session session, String schema, String lang,
...
     protected void writeUnsupported(Type type, CSVPrinter printer) throws IOException {
         printer.print(String.format("type %s is not supported", type.getName()));
     }
 
+    protected String replaceCarriageReturn(String value) {
+        String replacement = ctx.getParameter(CARRIAGE_RETURN_REPLACEMENT_CTX_DATA);
+        if (value == null || replacement == null) {
+            return value;
+        }
+        return NEWLINE.matcher(value).replaceAll(replacement);
```
efge, line 199: In order to avoid escaping in the caller and avoid too much work (not quote when there's nothing to replace, which is common) we could do:

    Matcher m = NEWLINE.matcher(value);
    if (m.find()) {
        value = m.replaceAll(Matcher.quoteReplacement(replacement));
    }
    return value;

efge (changes requested):

efge (approved):

troger (approved):

troger (approved):

efge (approved):

bdelbosc (approved):

PR #4641, opened 2021-02-11
Task NXP-30153 improve pipeline build/stage result

kevinleturc:


ci/Jenkinsfiles/build.groovy
```diff
@@ -527,7 +527,7 @@ pipeline {
           ----------------------------------------"""
           runFunctionalTests('ftests')
         }
-        findText regexp: ".*ERROR.*", fileSet: "ftests/**/log/server.log"
+        findText regexp: ".*ERROR.*", fileSet: "ftests/**/log/server.log", buildResult: 'UNSTABLE'
```
atchertchian, line 530: Maybe adapt the slack notif accordingly (?) see https://github.com/nuxeo/nuxeo-explorer/blob/master/ci/Jenkinsfiles/build.groovy#L412 (where ERROR in logs are frequent too)

kevinleturc:
Trying another approach by declaring the plugin like mentioned [here](https://plugins.jenkins.io/text-finder/)

kevinleturc:
Trying another approach by putting findText in the post/always step.

ci/Jenkinsfiles/build.groovy
kevinleturc: This change was more a test to help me to understand what is going on. In the end, this stage should make the test FAILURE, the change has been removed.

ci/Jenkinsfiles/build.groovy
```diff
@@ -262,7 +260,7 @@ def buildUnitTestStage(env) {
 }
 
 void archiveKafkaLogs(namespace, logFile) {
-  sh "kubectl logs ${TEST_KAFKA_POD_NAME} --namespace=${namespace} > ${logFile}"
+  sh "kubectl logs ${TEST_KAFKA_POD_NAME} --namespace=${namespace} > ${logFile} || echo"
```
efge, line 263: What's the `|| echo` for? If it's to not stop if the previous command returns an error, the traditional way is `|| true`

kevinleturc: It is to avoid the error indeed, I think I tried the echo way without success, let do another try 🤞.

efge: `echo` and `true` both returns a status 0 (ok) so they should be equivalent in terms of error recovery, it's just that there's no need to output a newline.

troger (approved):

ataillefer (approved):

efge (approved):

PR #4644, opened 2021-02-12
NXP-30146: Fix random ftest error log on AuditStorageLogWriterPool

atchertchian:


atchertchian:
Not sure we want to check the component manager status on each stream computation -> created #4645 as an alternative

atchertchian:
Will be handled by #4646

PR #4645, opened 2021-02-12
NXP-30146: Fix random ftest error log on AuditStorageLogWriterPool (alternative)

atchertchian:
Alternative to #4644 where we can have "post start" listeners


atchertchian:
Will be handled by #4646

PR #4654, opened 2021-02-16
NXP-30095: fix Auth.LoginAs and Auth.Logout when there is no session in context

efge:


kevinleturc (dismissed):

modules/platform/nuxeo-automation/nuxeo-automation-core/src/main/java/org/nuxeo/ecm/automation/core/operations/login/LoginAs.java
```diff
@@ -52,7 +52,8 @@ public void run() throws LoginException, OperationException {
         NuxeoLoginContext lc;
         if (name == null) {
             Principal origPrincipal = ctx.getPrincipal();
-            lc = Framework.loginSystem(origPrincipal.getName());
+            String origUser = origPrincipal == null ? null : origPrincipal.getName();
+            lc = Framework.loginSystem(origUser);
```
ataillefer, line 56: Wouldn't it be clearer to call `Framework.loginSystem()` in the null case?

The Javadoc form `Framework#loginSystem(String originatingUser)` doesn't say what is expected when the `originatingUser` parameter is null, neither does the underlying `LoginService#loginAs`.

You have to go down to `LoginComponent#SystemID(String origUser)` to understand what is happening.

efge: `Framework.loginSystem(null)` means login as system user with no originating user, so to me it makes sense that it would be the same as ``Framework.loginSystem()`.

ataillefer: Maybe, but it's still not clear in the Javadoc.

efge:
Added one commit to clarify Framework.loginSystem Javadoc

ataillefer (approved):

kevinleturc (approved):

PR #4658, opened 2021-02-17
NXP-28660: avoid popup at startup when testing ps2pdf converter

atchertchian:


nuxeo-core/nuxeo-core-convert-plugins/src/main/resources/OSGI-INF/commandline-pdf2text-contrib.xml
```diff
@@ -16,6 +16,7 @@
       <winCommand>gswin64c</winCommand>
       <parameterString>-dNOPAUSE -dBATCH -sDEVICE=pdfwrite -dEPSFitPage -sOutputFile=#{targetFilePath}
         #{sourceFilePath}</parameterString>
+      <testParameterString>-dNODISPLAY</testParameterString>
```
efge, line 19: Why only in tests? I guess we want it also in regular usage...

efge (changes requested):

nuxeo-core/nuxeo-core-convert-plugins/src/main/resources/OSGI-INF/commandline-pdf2text-contrib.xml
atchertchian: My understanding is that the display happens only when executing the command "gs" without any other parameters: in regular usage, the "parameterString" above prevents the problem.

atchertchian: Also, since I seem to be the only annoyed by this issue :D : I don't want to introduce any misbehaviour in the regular usage in a backport.

efge: Ah yes sorry I misremembered what `testParameterString` is for, it's for testing if a command is present, not for unit test context...

efge (approved):

atchertchian:
(T&P in progress here: https://qa2.nuxeo.org/jenkins/view/TestAndPush/job/TestAndPush/job/ondemand-testandpush-atchertchian-10.10/15/)

tmartins (approved):

PR #4673, opened 2021-02-25
NXP-30182: Keep a bulk status for 24h in case of error

bdelbosc:


kevinleturc (approved):

modules/core/nuxeo-core-bulk/src/main/java/org/nuxeo/ecm/core/bulk/BulkServiceImpl.java
```diff
@@ -99,7 +99,11 @@
     public static final long COMPLETED_TTL_SECONDS = 3_600;
 
     // How long we keep the command and its status in the kv store once aborted
-    public static final long ABORTED_TTL_SECONDS = 7_200;
+    public static final long ABORTED_TTL_SECONDS = 43_200;
+
+    // @since 11.5
+    // How long we keep the command and its status in the kv store once completed with an error
+    public static final long COMPLETED_IN_ERROR_TTL_SECONDS = 86_400;
```
troger, line 106: Worth using `Duration` here to get something easier to read? :)

Something like
```
public static final long ABORTED_TTL_SECONDS =  Duration.ofHours(1).toSeconds()
```
?

bdelbosc: 86_400 is 2 times easier than 43_200, but using duration will make it hard to backport in java8 :)

efge: Duration is in Java 8 :)

bdelbosc: yes but with different methods getSeconds vs toSeconds

efge: Yes toSeconds was added after 8.

efge: But getSeconds still exists in 11 and does the same thing.

troger (approved):

PR #4677, opened 2021-03-01
NXP-30135: extend startup cluster-wide lock to Workflow model import

efge:


modules/platform/nuxeo-platform-document-routing/nuxeo-routing-core/src/main/java/org/nuxeo/ecm/platform/routing/core/listener/RouteModelsInitializator.java
```diff
@@ -32,13 +40,43 @@
...
+            session.save();
+        });
+    }
+
+    protected Duration getLockDuration() {
+        Duration duration;
+        try {
+            String prop = Framework.getProperty(CLUSTER_START_DURATION_PROP);
+            if (isBlank(prop)) {
+                duration = Duration.ZERO;
```
kevinleturc, line 69: What do you think about adding the `isBlank` check to `DurationUtils` and throw a DateTimeParseException in such a case?

efge: Actually I already added `DurationUtils.parsePositive` specifically for that in the past, see  https://github.com/nuxeo/nuxeo/commit/99c78f67c5bb4eca7f43743359f074c6f257790d but I don't know why I copy/paste old code in this case...

efge:
(force-pushed to update to use `DurationUtils.parsePositive`)

kevinleturc (approved):

troger (approved):

PR #4692, opened 2021-03-03
NXP-30227: Add picker page provider definition

mnixo:
The changes are based on what Studio generates for this type of configuration, but of course we can adapt them to better fit the existing contributions.

### DIFF
--- a/modules/platform/nuxeo-platform-imaging/nuxeo-platform-imaging-core/src/main/resources/OSGI-INF/imaging-pageprovider-contrib.xml
+++ b/modules/platform/nuxeo-platform-imaging/nuxeo-platform-imaging-core/src/main/resources/OSGI-INF/imaging-pageprovider-contrib.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0"?>
+
+<component name="org.nuxeo.ecm.platform.picture.pageproviders">
+
+  <extension target="org.nuxeo.ecm.platform.query.api.PageProviderService" point="providers">
+
+    <coreQueryPageProvider name="document_picker">
+      <trackUsage>true</trackUsage>
+      <whereClause docType="AdvancedSearch">
+        <predicate parameter="dc:title" operator="FULLTEXT">
+          <field schema="advanced_search" name="title" />
+        </predicate>
+        <fixedPart>
+          ecm:mixinType = 'Picture' AND
+          file:content IS NOT NULL AND
+          ecm:mixinType != 'HiddenInNavigation' AND
+          ecm:isVersion = 0 AND
+          ecm:isTrashed = 0
+        </fixedPart>
+      </whereClause>
+      <pageSize>20</pageSize>
+    </coreQueryPageProvider>
+
+  </extension>
+
+</component>
--- a/modules/platform/nuxeo-platform-imaging/nuxeo-platform-imaging-core/src/main/resources/META-INF/MANIFEST.MF
+++ b/modules/platform/nuxeo-platform-imaging/nuxeo-platform-imaging-core/src/main/resources/META-INF/MANIFEST.MF
@@ -24,6 +24,7 @@ Bundle-ManifestVersion: 2
 Nuxeo-Component: OSGI-INF/imaging-service-framework.xml,
  OSGI-INF/imaging-service-contrib.xml,
  OSGI-INF/imaging-bulk-contrib.xml,
+ OSGI-INF/imaging-pageprovider-contrib.xml,
  OSGI-INF/picture-schemas-contrib.xml,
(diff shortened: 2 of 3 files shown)


modules/platform/nuxeo-platform-default-config/src/main/resources/OSGI-INF/document-pageprovider-contrib.xml
```diff
@@ -136,6 +136,25 @@
       <pageSize>20</pageSize>
     </coreQueryPageProvider>
 
+    <genericPageProvider class="org.nuxeo.elasticsearch.provider.ElasticSearchNxqlPageProvider" name="picker">
```
troger, line 139: Tests are failing because `nuxeo-platform-default-config` is deployed on some tests but not `nuxeo-elasticsearch-*` modules.

Moreover, you don't need a new `Picker` document type as it contains only a `title` field, there is already a title field in the `AdvancedSearch` doc type.

The page provider is linked to the `Picture` facet, so I would move it to `nuxeo-platform-imaging-core` module in a new contribution file, and not `nuxeo-platform-default-config`.

I would have declared it like the other page providers (not strictly linked to ES as it's not needed):
```
  <coreQueryPageProvider name="picker">
      <trackUsage>true</trackUsage>
      <whereClause docType="AdvancedSearch">
        <predicate parameter="dc:title" operator="FULLTEXT">
          <field schema="advanced_search" name="title" />
        </predicate>
        <fixedPart>
          ecm:mixinType = 'Picture' AND
          file:content IS NOT NULL AND
          ecm:mixinType != 'HiddenInNavigation' AND
          ecm:isVersion = 0 AND
          ecm:isTrashed = 0
        </fixedPart>
      </whereClause>
      <pageSize>20</pageSize>
  </coreQueryPageProvider>
```

An then, add it to the list here https://github.com/nuxeo/nuxeo/blob/e2150ea02a11905fe564315d7e91036153cb6476/server/nuxeo-nxr-server/src/main/resources/templates/common-base/nuxeo.defaults#L105 so that it can be used with ES.

@bdelbosc do you confirm that's the good approach?

efge: It cannot be generic as `file:content IS NOT NULL` only works on ES.
Adding things to `elasticsearch.override.pageproviders` when it can be declared in the XML instead is weird. The property is for overrides, we shouldn't use it as part of the general config.

efge: Note for the future: if we had introduced a new document type / schema I would have insisted that it be called `NuxeoFilePicker` and same for the schema, as there's a risk of collision with customer types otherwise.

mnixo:
I tried to meet somewhere in the middle. In summary:
- Moved the page provider contribution to `nuxeo-platform-imaging-core`.
- Kept `<genericPageProvider>` instead of `<coreQueryPageProvider>` + override on `elasticsearch.override.pageproviders`.
- Renamed the page provider as `nuxeoPicker` instead of `picker`. We (UI) don't have anything final just yet, so we can take the opportunity to use a better id (if `nuxeoPicker` is not too inconsistent with the other default page provider ids, of course).

modules/platform/nuxeo-platform-imaging/nuxeo-platform-imaging-core/src/main/resources/OSGI-INF/imaging-pageprovider-contrib.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0"?>
+
+<component name="org.nuxeo.ecm.platform.picture.pageproviders">
+
+  <extension target="org.nuxeo.ecm.platform.query.api.PageProviderService" point="providers">
+
+    <genericPageProvider class="org.nuxeo.elasticsearch.provider.ElasticSearchNxqlPageProvider" name="nuxeoPicker">
```
kevinleturc, line 7: Tests are failing because nuxeo-platform-imaging-core doesn't depend on `nuxeo-elasticsearch-*` modules, thus we have a ClassNotFound error.

mnixo:
Reverted back to Thomas' original suggestion, which includes the addition to `elasticsearch.override.pageproviders`.
This should clear up the tests.
Still open to suggestions.

kevinleturc (approved):

troger (approved):

troger (approved):

mnixo:
Ended up renaming as `document_picker` for better consistency with the existing page provider.

kevinleturc (approved):

PR #4720, opened 2021-03-19
NXP-30123: new default cors configuration

npeacock88:


server/nuxeo-nxr-server/src/main/resources/templates/common-base/nuxeo.defaults
```diff
@@ -321,3 +321,7 @@ tracing.datadog.timeout=10s
 tracing.stackdriver.enabled=false
 tracing.stackdriver.timeout=30s
 tracing.stackdriver.gcpProjectId=
+
+# Cors config
+nuxeo.cors.allowSubdomains=true
+nuxeo.cors.supportedMethods=GET,POST,HEAD,OPTIONS,DELETE,PUT
```
efge, line 327: I think it's worth also adding an empty

    nuxeo.cors.urls=

to make explicit what can be configured. You may want to change the comment to `# CORS config (active if nuxeo.cors.urls has a value)`

Then in the nxftl file you can test presence with `<#if (nuxeo.cors.urls)?has_content>`

server/nuxeo-nxr-server/src/main/resources/templates/common-base/nxserver/config/cors-config.xml.nxftl
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<component name="org.nuxeo.ecm.platform.ui.web.cors.config">
```
efge, line 2: It will be very important to update the CORS documentation as soon as this is merged, in order to explain the new properties, and to stress that if people provide their own XML contribution they will now have to add `<require>org.nuxeo.ecm.platform.ui.web.cors.config</require>` as well.

efge (changes requested):

kevinleturc (changes requested):
Hello, could you make one single commit with the _NXP-30123_ prefix?

kevinleturc (approved):

server/nuxeo-nxr-server/src/main/resources/templates/common-base/nxserver/config/cors-config.xml.nxftl
```diff
@@ -0,0 +1,12 @@
...
+
+<#if (nuxeo.cors.urls)?has_content>
+  <extension target="org.nuxeo.ecm.platform.web.common.requestcontroller.service.RequestControllerService" point="corsConfig">
+      <corsConfig name="cors" allowOrigin="${nuxeo.cors.urls}" allowSubdomains="${nuxeo.cors.allowSubdomains}" supportedMethods="${nuxeo.cors.supportedMethods}">
+        <pattern>/.*</pattern>
```
ataillefer, line 7: Do we want to make the pattern configurable?

server/nuxeo-nxr-server/src/main/resources/templates/common-base/nxserver/config/cors-config.xml.nxftl
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<component name="org.nuxeo.ecm.platform.ui.web.cors.config">
+
+<#if (nuxeo.cors.urls)?has_content>
+  <extension target="org.nuxeo.ecm.platform.web.common.requestcontroller.service.RequestControllerService" point="corsConfig">
+      <corsConfig name="cors" allowOrigin="${nuxeo.cors.urls}" allowSubdomains="${nuxeo.cors.allowSubdomains}" supportedMethods="${nuxeo.cors.supportedMethods}">
```
ataillefer, line 6: What about the other parameters described in the documentation: https://doc.nuxeo.com/nxdoc/cross-origin-resource-sharing-cors/.
Do we want to make them configurable?
@efge

efge: Why not yes. I don't know what's most often configured. It's up to whoever asked for this configurability to decide what's best for them.

gspradlin: Hi Nikki, pointed me here to answer your question.  The configuration scope in NXP-30123 is the result of reviewing the CORS configurations implemented for many clients. I reviewed all CORS configurations in the PS team's Git repo (nuxeo-projects) which makes up most of Nuxeo Cloud as well as some other private repositories for self-managed customers. Additionally, pen tests run by customers generally have findings about CORS and the CS team helps to make the CORS policy secure. Here are my findings regarding all CORS parameters.
 
- allowOrigin: Whitespace-separated list of origins that the CORS filter must allow. 
The most common configuration is the list of origins to either allow all origins or only a whitelist of origins.

- allowSubdomains: If true the CORS filter will allow requests from any origin which is a subdomain origin of the allowed origins. 
The next most common configuration is to allow subdomains which is why we will default it to true when this CORS config is enabled, but it can be set to false (which is the default in the code).

- supportedMethods: List of the supported HTTP methods. 
In some cases I have observed a restriction on the supported methods which is why this is also configurable.  

- allowGenericHttpRequests: If false, only valid and accepted CORS requests that be allowed (strict CORS filtering). 
The default in the code is true.  I have never observed this to be set to false.

- supportedHeaders: The names of the supported author request headers. 
I have never observed this to be set.

- exposedHeaders: List of the response headers other than simple response headers that the browser should expose to the author of the cross-domain request through the XMLHttpRequest.getResponseHeader() method. 
I have never observed this to be set.

- supportsCredentials: Indicates whether user credentials, such as cookies, HTTP authentication or client-side certificates, are supported. 
The default in the code is true and I have only observed this to be set to true. 

- maxAge: Indicates how long the results of a preflight request can be cached by the web browser, in seconds.
I've never observed this to be set on purpose, but on a rare occasion it may be set to 3600 only because there is an example in Explorer with this value.

server/nuxeo-nxr-server/src/main/resources/templates/common-base/nxserver/config/cors-config.xml.nxftl
gspradlin: I have observed some slight variations in the pattern.  While I don't think the differences have much purpose, it seems a good idea to make this configurable (just in case).  @npeacock88 Can you please make the pattern configurable with the default pattern nuxeo.cors.pattern=/.*

ataillefer (approved):

kevinleturc (approved):

efge (approved):

PR #4723, opened 2021-03-22
Fix nxp 30274 timeout windows

bdelbosc:


kevinleturc (approved):

modules/platform/nuxeo-thumbnail/src/main/java/org/nuxeo/ecm/platform/thumbnail/listener/UpdateThumbnailListener.java
```diff
@@ -94,9 +95,17 @@ protected void processDoc(CoreSession session, DocumentModel doc) {
...
             session.saveDocument(doc);
+            newTransaction();
         }
     }
 
+    protected void newTransaction() {
+        if (TransactionHelper.isTransactionActiveOrMarkedRollback()) {
+            TransactionHelper.commitOrRollbackTransaction();
+        }
+        TransactionHelper.startTransaction();
```
ataillefer, line 106: Here we start a transaction even in the case where there wasn't any transaction active or marked for rollback. Though this case is it's unlikely, can it be an issue?

ataillefer (approved):

modules/platform/nuxeo-thumbnail/src/main/java/org/nuxeo/ecm/platform/thumbnail/listener/UpdateThumbnailListener.java
bdelbosc: It should not be an issue, this async listener is run by a ListenerWork so inside a transaction like all `AbstractWork#run`,
even if run as PostCommit listener or as a bundle bulk events `EventBundleBulkRunner#call` it is inside a transaction.

PR #4736, opened 2021-04-02
NXP-30311: Unregister services whose component fails to start

kevinleturc:


modules/runtime/nuxeo-runtime/src/main/java/org/nuxeo/runtime/model/impl/ComponentManagerImpl.java
```diff
@@ -387,11 +387,16 @@ public void removeListener(ComponentManager.Listener listener) {
...
         if (ri == null) {
+            log.debug("The component exposing the service: {} doesn't exist", serviceClass);
             return null;
         }
         ComponentInstance ci = ri.getComponent();
         if (ci == null) {
-            log.debug("The component exposing the service {} is not resolved or not started", serviceClass);
+            log.debug("The component exposing the service: {} is not resolved", serviceClass);
+        }
+        if (Boolean.getBoolean("nuxeo.start.strict") && ri.getState() == RegistrationInfo.START_FAILURE) {
```
bdelbosc, line 397: Maybe `ri.getState() == RegistrationInfo.START_FAILURE` first because strict mode will always be true and getBoolean is slower

ataillefer (approved):

bdelbosc (approved):

PR #4739, opened 2021-04-06
Fix nxp 30312 npe audio file thumbnail

ataillefer:


bdelbosc:
ok but the mp3 file is 772 KB way too big

ataillefer:
> ok but the mp3 file is 772 KB way too big

Right, I trimmed it to 3.49 KB.

troger (approved):

bdelbosc (approved):

PR #4753, opened 2021-04-12
NXP-30314: Wait for topic creation

bdelbosc:
It may take several seconds after createTopics returns success for all the brokers
to become aware that the topics have been created, especially in the case of
massive creation with replications


bdelbosc:
t&p in progress

kevinleturc (approved):

ataillefer (approved):

PR #4759, opened 2021-04-20
NXP-30265: allow to hide nuxeoctl logs

troger:


server/nuxeo-server-tomcat/src/main/resources/tomcat/bin/nuxeoctl
```diff
@@ -25,6 +25,11 @@ if [[ "x $@" =~ " -d" || "x $@" =~ " --debug" || -n "$LAUNCHER_DEBUG" ]]; then
   set -x
 fi
 
+VERBOSE=true
+if [[ "x $@" =~ " -q" || "x $@" =~ " --quiet" ]]; then
```
efge, line 29: FYI you don't need the `"x$1"` trick to protect values when using `[[`, this is only needed for `[`.
See https://news.ycombinator.com/item?id=26776956 for more.

(also applies to the `-d` test above)

efge (approved):

kevinleturc (approved):

server/nuxeo-server-tomcat/src/main/resources/tomcat/bin/nuxeoctl
troger: Noted for future work. I will keep the code as it to be consistent with the existing.

PR #4772, opened 2021-05-06
NXP-30357: Fix blob MIME type with S3 direct upload

ataillefer:
Split Content-Type header into MIME type and charset, dispatch them respectively to the blob's MIME type and encoding properties.

### DIFF
--- a/modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
+++ b/modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
@@ -44,6 +44,9 @@ import java.util.Objects;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.http.HeaderElement;
+import org.apache.http.NameValuePair;
+import org.apache.http.message.BasicHeaderValueParser;
 import org.nuxeo.ecm.automation.server.jaxrs.batch.Batch;
 import org.nuxeo.ecm.automation.server.jaxrs.batch.handler.AbstractBatchHandler;
 import org.nuxeo.ecm.automation.server.jaxrs.batch.handler.BatchFileInfo;
@@ -155,6 +158,34 @@ public class S3DirectBatchHandler extends AbstractBatchHandler {
 
     // public for tests
     public String blobProviderId;
+
+    /**
+     * @since 11.5
+     */
+    public static String getMimeType(String contentType) {
+        Objects.requireNonNull(contentType);
+        HeaderElement headerElement = BasicHeaderValueParser.parseHeaderElement(contentType, null);
+        return headerElement.getName();
+    }
+
+    /**
+     * @since 11.5
+     */
+    public static String getCharset(String contentType) {
+        Objects.requireNonNull(contentType);
+        HeaderElement headerElement = BasicHeaderValueParser.parseHeaderElement(contentType, null);
+        String encoding = null;
+        for (NameValuePair param : headerElement.getParameters()) {
+            if (param.getName().equalsIgnoreCase("charset")) {
+                String s = param.getValue();
+                if (!isBlank(s)) {
+                    encoding = s;
+                }
+                break;
+            }
+        }
+        return encoding;
+    }
 
     @Override
     protected void initialize(Map<String, String> properties) {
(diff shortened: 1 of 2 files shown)


modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
```diff
@@ -156,6 +160,27 @@
     // public for tests
     public String blobProviderId;
 
+    /**
+     * Returns the MIME type related to the given {@code contentType}, stripping the charset if present and not a text
+     * content type. For instance, "application/pdf; charset=UTF-8" results in "application/pdf".
```
bdelbosc, line 165: A Content-Type is composed of Mime Type and eventually a Charset, if the method is supposed to return a mime type no need to explain that we remove the charset part, this is the purpose of the method

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
```diff
@@ -156,6 +160,27 @@
...
+     * time, the Nuxeo code doesn't deal with such a MIME type.
+     *
+     * @since 11.5
+     */
+    public static String getBlobMimeType(String contentType) {
+        Objects.requireNonNull(contentType);
+        if (contentType.startsWith("text/")) {
+            return contentType;
+        }
+        Matcher m = CHARSET_CONTENT_TYPE_PATTERN.matcher(contentType);
```
bdelbosc, line 177: not sure we need a regex to split on ';' but ok

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
efge: That's a good point, in Nuxeo Blob's `blob.getMimeType` is expected to return the MIME type in all cases. When there's a charset it should be put in the "encoding" part, available with `blob.getEncoding`. 

I didn't understand why the S3 metadata has a `metadata.getContentEncoding()` that's redundant with the `metadata.getContentType()` but actually we use `metadata.getContentEncoding()` wrong, it's something else related to the HTTP transfer itself... So there's another bug lurking here, and we never extract properly the charset(=encoding) from the S3 metadata info

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
```diff
@@ -127,6 +129,8 @@
 
     public static final String INFO_USE_S3_ACCELERATE = "useS3Accelerate";
 
+    public static final Pattern CHARSET_CONTENT_TYPE_PATTERN = Pattern.compile("(.*);.*");
```
efge, line 132: In theory this is not generic enough, for instance this is a valid `Content-Type` header: `text/plain  ;  CHARSET = UTF-8 ; foo=bar`

efge (changes requested):

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
ataillefer: We've agreed to:
- not use `metadata.getContentEncoding()`
- split `metadata.getContentType()` into the MIME type and charset, then dispatch them respectively to the blob's MIME type and encoding properties

efge (approved):

modules/core/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/core/storage/sql/S3DirectBatchHandler.java
```diff
@@ -156,6 +159,40 @@
     // public for tests
     public String blobProviderId;
 
+    /**
+     * @since 11.5
+     */
+    public static HeaderElement getHeaderElement(String contentType) {
```
bdelbosc, line 165: What is the point of exposing the an org.apache.http.HeaderElement object? 
Can't we just have 
String getMimeType(String contentType)
String getCharset(String contentType)

ataillefer: For the tests, though we could also do `BasicHeaderValueParser.parseHeaderElement(contentType, null)` in the tests directly...

efge: Yes we could inline the code of getHeaderElement directly in the two methods, this is not a time-critical method that's called a lot so we can afford to parse the header twice.

efge (approved):

efge (approved):

troger (approved):

PR #4794, opened 2021-05-21
NXP-30401: Log the download event for each document when downloading a container

kevinleturc:


modules/platform/rest-api/nuxeo-rest-api-test/src/test/java/org/nuxeo/ecm/restapi/test/BulkDownloadTest.java
```diff
@@ -0,0 +1,73 @@
...
+@RunWith(FeaturesRunner.class)
+@Features({ RestServerFeature.class, CollectionFeature.class })
+@RepositoryConfig(cleanup = Granularity.METHOD, init = RestServerInit.class)
+@Deploy("org.nuxeo.ecm.actions")
+@Deploy("org.nuxeo.ecm.platform.rendition.api")
+@Deploy("org.nuxeo.ecm.platform.rendition.core")
+public class BulkDownloadTest extends BaseTest {
+
+    @Inject
+    protected CoreSession session;
```
efge, line 55: There's already a session injected in the base class

efge (changes requested):

modules/platform/rest-api/nuxeo-rest-api-test/src/test/java/org/nuxeo/ecm/restapi/test/BulkDownloadTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    // NXP-30401
+    @Test
+    public void testAuditEventOnFolderDownload() {
```
efge, line 59: It may be worth explaining in a comment (if I understand correctly) that the Blob.BulkDownload of a zip will trigger a zip rendition which itself is going to call the Document.GetContainerRendition operation. It's not obvious at all

kevinleturc:
I pushed review fixes and I added another `logDownload` call to log download in case of selection.

efge (approved):

ataillefer (approved):

PR #4847, opened 2021-06-17
NXP-30243 NXP-30432 NXP-30281 NXP-30215: add indexes

kevinleturc:


ataillefer (approved):

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/jdbc/SQLInfo.java
```diff
@@ -867,6 +867,10 @@ protected void initFragmentSQL(String tableName) {
         maker.newColumn(Model.MAIN_KEY, type);
         maker.newFragmentFields();
         maker.postProcess();
+        // NXP-30243
+        if ("collection:documentIds".equals(tableName)) {
+            maker.table.addIndex("item");
```
efge, line 872: I'd prefer `Model.COLL_TABLE_VALUE_KEY` instead of `"item"` here

efge (changes requested):

modules/core/nuxeo-core-storage-sql/nuxeo-core-storage-sql/src/main/java/org/nuxeo/ecm/core/storage/sql/jdbc/SQLInfo.java
efge: (I assume that `collection:documentIds` is correct and has been tested — I'm not sure if the table name is already mangled from the fragment name at this point)

kevinleturc: I tested it on an H2 DB and the tableName is as the condition, not mangled.

efge (approved):

ataillefer (approved):

PR #4853, opened 2021-06-18
NXP-30192: always check Read permission in DocumentModelResolver

troger:


kevinleturc (dismissed):

modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/model/DocumentModelResolver.java
```diff
@@ -197,6 +199,14 @@ public Object fetch(Object value, Object context) {
         return docHolder.getValue();
     }
 
+    protected boolean hasReadPermission(CoreSession session, DocumentRef docRef) {
```
efge, line 202: Maybe add a comment explaining why we don't use `session.exists` here.

modules/core/nuxeo-core-test/src/test/java/org/nuxeo/ecm/core/resolver/TestDocumentModelResolver.java
```diff
@@ -671,6 +681,32 @@ public void testUserWithNoAccessToDocument() {
...
+        acp.addACE(ACL.LOCAL_ACL, new ACE("bob", "Read", true));
+        fooDoc.setACP(acp, true);
+        transactionalFeature.nextTransaction();
+
+        CoreSession bobSession = coreFeature.getCoreSession("bob");
+        DocumentModel resolvedDoc = (DocumentModel) fooDoc.getProperty(REPO_AND_ID_XPATH)
+                                                          .getObjectResolver()
+                                                          .fetch(bobSession);
+        // bob cannot read the referenced document
+        assertNull(resolvedDoc);
```
efge, line 707: Could you also test explicitly what happens on a list? I guess here the list will contain a `null`, but as we discussed another option would be to skip the value from the list completely.

kevinleturc (approved):

ataillefer (approved):

efge (approved):

PR #4857, opened 2021-06-23
NXP-30170: Don't re-create workflow if it hasn't change

kevinleturc:


troger (approved):

efge:
Commit message: `change` -> `changed`

I'm a bit disappointed, I expected that the fix for this would be generic at the importer level, to not re-create a document if an identical one already exists. Here you save the hash of the original workflow zip to not re-do the import if previously done. But this is not going to help if for some reason the workflow has been changed in storage by other means (rogue document update, code bug, database-level change, etc.).

ataillefer (dismissed):

troger (approved):

ataillefer (dismissed):

kevinleturc:
Rebased to take into account skaffold changes.

troger (approved):

PR #4909, opened 2021-08-04
NXP-30518: Continue on errors in automationUi action

kevinleturc:


modules/platform/nuxeo-automation/nuxeo-automation-features/src/main/java/org/nuxeo/ecm/automation/core/operations/services/bulk/AutomationBulkActionUi.java
```diff
@@ -26,7 +26,7 @@
  *
  * @since 11.5
  */
-public class AutomationBulkActionUi extends AutomationBulkAction {
+public class AutomationBulkActionUi extends AbstractAutomationBulkAction {
```
bdelbosc, line 29: is this wanted?

kevinleturc: Yes it is, otherwise, the abstract is not useful and we could merge `AutomationBulkAction` & `AbstractAutomationBulkAction`.

bdelbosc (approved):

PR #4945, opened 2021-09-23
NXP-30624: Set blob length at lower level

bdelbosc:
To avoid trigger any listener and changing the blob key.


bdelbosc:
ready to be reviewed, successfully tested on local envs.

efge (approved):

troger (approved):

kevinleturc (approved):

PR #4963, opened 2021-10-13
NXP-22842: Add Redis checking at startup

tfowley:


kevinleturc (approved):

nuxeo-runtime/nuxeo-runtime-redis/src/main/java/org/nuxeo/ecm/core/redis/RedisChecker.java
```diff
@@ -0,0 +1,82 @@
...
+            String pong = executor.execute(Jedis::ping);
+            if (!"PONG".equals(pong)) {
+                throw new RuntimeException("Unable to ping Redis, received " + pong); // NOSONAR
+            }
+        } catch (RuntimeException e) {
+            throw new ConfigurationException("Unable to reach Redis on prefix: " + config.prefix, e);
+        }
+    }
+
+    protected RedisPoolDescriptor getConfig(ConfigurationGenerator cg) throws ConfigurationException {
```
ataillefer, line 60: Why can't we extend the BackingChecker as in 2021?

nuxeo-runtime/nuxeo-runtime-redis/src/main/java/org/nuxeo/ecm/core/redis/RedisChecker.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+public class RedisChecker implements BackingChecker {
+
+    private static final String TEMPLATE_NAME = "redis";
+
+    private static final String CONFIG_NAME = "redis-config.xml";
+
+    @Override
+    public boolean accepts(ConfigurationGenerator cg) {
+        return cg.getTemplateList().contains(TEMPLATE_NAME);
```
ataillefer, line 43: In 2021 we call `getIncludedTemplateNames` on the config holder, the `ConfigurationGenerator` also has this method in 10.10, is the difference intended?

tfowley: I can't find the getIncludedTemplateNames in ConfigurationGenerator in 10.10, I could only find a getIncludedTemplates but this returns a list of files instead of Strings so I used getTemplateList because it was also used in the MongoDBChecker class.

nuxeo-runtime/nuxeo-runtime-redis/src/main/java/org/nuxeo/ecm/core/redis/RedisChecker.java
tfowley: @kevinleturc had mentioned something about this but I think it was out of scope for this task, I might be wrong about that though 🤷‍♂️

kevinleturc: In 10.10, we don't have the generic method to resolve the descriptors. Thus we need to have this kind of method in each checkers.

ataillefer (approved):

PR #4985, opened 2021-11-16
NXP-30550: Throws an error if document structure has changed on DBS

kevinleturc:


guirenard (dismissed):

efge:
The commit message seems strange to me. I would write "Don't throw an error if...".

kevinleturc:
> The commit message seems strange to me. I would write "Don't throw an error if...".

But I'm throwing an error in such a case, a Nuxeo one instead of the ClassCastException.

Would _Throw a typed error..._ fit?

efge:
> But I'm throwing an error in such a case, a Nuxeo one instead of the ClassCastException.
> Would Throw a typed error... fit?

Ah yes my mistake sorry. So maybe "throw a better error if..."?

ataillefer (approved):

bdelbosc (approved):

PR #4987, opened 2021-11-17
NXP-30702: Update the loading of the Groovy class

glefevre:


addons/nuxeo-segment.io-connector/src/main/java/org/nuxeo/segment/io/SegmentIOMapper.java
```diff
@@ -78,32 +81,39 @@ public boolean isScreen() {
...
+            if (groovyMapping != null && !groovyMapping.isEmpty()) {
+                sb.append(groovyMapping);
+            }
+            try (GroovyClassLoader loader = new GroovyClassLoader(this.getClass().getClassLoader())) {
+                klass = loader.parseClass(sb.toString());
+            } catch (IOException e) {
+                throw new NuxeoException(String.format("Error during Groovy script execution for the '%s' segmentIO mapper", name), e);
+            }
+        }
+        return klass;
```
kevinleturc: Should we do a double-checked locking on the initialization? It may be interesting if the method is concurrently called. This will end with:
```suggestion
        if (klass == null) {
            synchronized(this) {
                if (klass == null) {
                    // Define Groovy class based on mapper contributions
                    StringBuffer sb = new StringBuffer();
                    for (String key : parameters.keySet()) {
                        sb.append("mapping.put(\"");
                        sb.append(key);
                        sb.append("\", ");
                        sb.append(parameters.get(key));
                        sb.append(");\n");
                    }

                    if (groovyMapping != null && !groovyMapping.isEmpty()) {
                        sb.append(groovyMapping);
                    }
                    try (GroovyClassLoader loader = new GroovyClassLoader(this.getClass().getClassLoader())) {
                        klass = loader.parseClass(sb.toString());
                    } catch (IOException e) {
                        throw new NuxeoException(String.format("Error during Groovy script execution for the '%s' segmentIO mapper", name), e);
                    }
                }
            }
        }
        return klass;
```

kevinleturc (dismissed):
If my comment is not accurate we can go for the merge.

addons/nuxeo-segment.io-connector/src/main/java/org/nuxeo/segment/io/SegmentIOMapper.java
```diff
@@ -78,32 +81,41 @@ public boolean isScreen() {
         return ACTIONS.screen.name().equalsIgnoreCase(target);
     }
 
-    public Map<String, Serializable> getMappedData(Map<String, Object> context) {
+    public Class<?> getGroovyClazz() {
+        if (klass == null) {
+            synchronized (this) {
+                // Define Groovy class based on mapper contributions
```
kevinleturc: A second nullity check should be done to avoid re-creating the resource by the threads which were waiting on the synchronized block.

glefevre: You're right, let me add it

akervern (approved):

kevinleturc (approved):

PR #4988, opened 2021-11-18
Fix nxp 30687 retention bulk actions 1010

guirenard:


kevinleturc (approved):

ataillefer (approved):
Sounds like some good JS :D

PR #5004, opened 2021-12-07
NXP-30560: prevent GC from deleting extra blobs when several blob pro…

guirenard:
…viders share storage

### DIFF
--- a/nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
+++ b/nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
@@ -20,6 +20,7 @@ package org.nuxeo.ecm.core.blob;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Deque;
 import java.util.LinkedList;
@@ -69,6 +70,13 @@ public class DocumentBlobManagerComponent extends DefaultComponent implements Do
     protected static final String MAIN_BLOB_XPATH = "content";
 
     protected Deque<BlobDispatcherDescriptor> blobDispatcherDescriptorsRegistry = new LinkedList<>();
+
+    protected volatile List<BinaryGarbageCollector> garbageCollectors;
+
+    /**
+     * true when several blob providers share a same storage.
+     */
+    protected volatile boolean sharedStorage;
 
     @Override
     public void deactivate(ComponentContext context) {
@@ -361,22 +369,48 @@ public class DocumentBlobManagerComponent extends DefaultComponent implements Do
     // find which GCs to use
     // only GC the binary managers to which we dispatch blobs
     protected List<BinaryGarbageCollector> getGarbageCollectors() {
-        List<BinaryGarbageCollector> gcs = new LinkedList<>();
-        for (String providerId : getBlobDispatcher().getBlobProviderIds()) {
-            BlobProvider blobProvider = getBlobProvider(providerId);
-            BinaryGarbageCollector gc = blobProvider.getBinaryGarbageCollector();
-            if (gc != null) {
-                gcs.add(gc);
+        if (garbageCollectors == null) {
+            synchronized (this) {
+                if (garbageCollectors == null) {
+                    List<BinaryGarbageCollector> gcs = new ArrayList<>();
+                    for (String providerId : getBlobDispatcher().getBlobProviderIds()) {
+                        BlobProvider blobProvider = getBlobProvider(providerId);
+                        BinaryGarbageCollector gc = blobProvider.getBinaryGarbageCollector();
+                        if (gc != null) {
+                            gcs.add(gc);
+                        }
+                    }
+                    long idCount = gcs.stream().map(BinaryGarbageCollector::getId).distinct().count();
+                    sharedStorage = idCount < gcs.size();
+                    garbageCollectors = gcs;
+                }
             }
         }
-        return gcs;
+        return garbageCollectors;
+    }
+
+    /**
+     * Get the list of garbage collectors.
+     *
+     * @param refresh if true the list is recomputed, use latest computation otherwise
+     * @return a list of garbage collectors
+     * @since 10.10-HF56
+     */
+    protected List<BinaryGarbageCollector> getGarbageCollectors(boolean refresh) {
+        if (refresh) {
+            synchronized (this) {
+                garbageCollectors = null;
+            }
+        }
+        return getGarbageCollectors();
     }
 
     @Override
     public BinaryManagerStatus garbageCollectBinaries(boolean delete) {
         // do the GC in a long-running transaction to avoid timeouts
         return runInTransaction(() -> {
-            List<BinaryGarbageCollector> gcs = getGarbageCollectors();
+            // Get a fresh list of collectors to initiate garbage collection
+            List<BinaryGarbageCollector> gcs = getGarbageCollectors(true);
             // start gc
             long start = System.currentTimeMillis();
             for (BinaryGarbageCollector gc : gcs) {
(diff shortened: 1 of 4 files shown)


guirenard:
This is a simplified backport of https://jira.nuxeo.com/browse/NXP-30560 to 10.10. It prevents GC from deleting extra blobs when several blob providers share storage but without switching to a callback system for the GC mark stage in order to do not break API on HF. 

Note that I am storing the list of Garbage Collectors in this [variable](https://github.com/nuxeo/nuxeo/pull/5004/files#diff-9436fed104a7062892d1b3f4f37fca058700f64e943f835213d774c7ec484ddbR75) for performances reason as it will be needed in each call of [markReferencedBinary](https://github.com/nuxeo/nuxeo/pull/5004/files#diff-9436fed104a7062892d1b3f4f37fca058700f64e943f835213d774c7ec484ddbR448)

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -434,13 +450,20 @@ public BinaryManagerStatus garbageCollectBinaries(boolean delete) {
 
     @Override
     public void markReferencedBinary(String key, String repositoryName) {
-        BlobProvider blobProvider = getBlobProvider(key, repositoryName);
-        BinaryGarbageCollector gc = blobProvider.getBinaryGarbageCollector();
-        if (gc != null) {
-            key = stripBlobKeyPrefix(key);
-            gc.mark(key);
+        List<BinaryGarbageCollector> gcs = getGarbageCollectors();
+        Set<String> ids = gcs.stream().map(BinaryGarbageCollector::getId).collect(toSet());
```
kevinleturc: Could it be interesting to just compute the count of unique ids?

It could be achieved like this:

```suggestion
        long idCount = gcs.stream().map(BinaryGarbageCollector::getId).distinct().count();
```

kevinleturc (dismissed):
Pretty nice to not break the API 👌.

The commit message needs to be reworded to use the [NXP-30739](https://jira.nuxeo.com/browse/NXP-30739) instead of [NXP-30560](https://jira.nuxeo.com/browse/NXP-30560)

kevinleturc (dismissed):

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -434,13 +450,20 @@ public BinaryManagerStatus garbageCollectBinaries(boolean delete) {
 
     @Override
     public void markReferencedBinary(String key, String repositoryName) {
-        BlobProvider blobProvider = getBlobProvider(key, repositoryName);
-        BinaryGarbageCollector gc = blobProvider.getBinaryGarbageCollector();
-        if (gc != null) {
-            key = stripBlobKeyPrefix(key);
-            gc.mark(key);
+        List<BinaryGarbageCollector> gcs = getGarbageCollectors();
+        long idCount = gcs.stream().map(BinaryGarbageCollector::getId).distinct().count();
```
efge: I think doing this for every call to `markReferencedBinary` is a bit too much, in a large repo there will be millions. It's better to move the computation of the `sharedStorage` boolean and the `gcs` list to the `garbageCollectBinaries` method, which is the one that starts the whole process.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -361,12 +371,18 @@ protected String stripBlobKeyPrefix(String key) {
...
-        List<BinaryGarbageCollector> gcs = new LinkedList<>();
-        for (String providerId : getBlobDispatcher().getBlobProviderIds()) {
-            BlobProvider blobProvider = getBlobProvider(providerId);
-            BinaryGarbageCollector gc = blobProvider.getBinaryGarbageCollector();
-            if (gc != null) {
-                gcs.add(gc);
+        if (gcs == null) {
+            synchronized (this) {
+                if (gcs == null) {
+                    gcs = new LinkedList<>();
```
efge: For the double-checked locking idiom you want to assign `gcs` only once and at then _end_, when the value is completely determined. So 

    List<BinaryGarbageCollector> list = new ArrayList<>(); // btw we don't use LinkedList 
    ...
        if (gc != null) { list.add(gc); }
    ...
    gcs = list;

efge (changes requested):

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -361,22 +375,47 @@ protected String stripBlobKeyPrefix(String key) {
...
+    /**
+     * Get the list of garbage collectors.
+     *
+     * @param refresh if true the list is recomputed, use latest computation otherwise
+     * @return a list of garbage collectors
+     * @since 10.10-HF56
+     */
+    protected List<BinaryGarbageCollector> getGarbageCollectors(boolean refresh) {
+        if (refresh) {
+            synchronized(this) {
```
efge: Space before `(`

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -99,10 +107,16 @@ public void unregisterContribution(Object contribution, String extensionPoint, C
 
     protected void registerBlobDispatcher(BlobDispatcherDescriptor descr) {
         blobDispatcherDescriptorsRegistry.add(descr);
+        synchronized (this) {
+            garbageCollectors = null;
+        }
```
efge: I don't think these are needed.

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -361,22 +375,47 @@ protected String stripBlobKeyPrefix(String key) {
...
-        List<BinaryGarbageCollector> gcs = new LinkedList<>();
-        for (String providerId : getBlobDispatcher().getBlobProviderIds()) {
-            BlobProvider blobProvider = getBlobProvider(providerId);
-            BinaryGarbageCollector gc = blobProvider.getBinaryGarbageCollector();
-            if (gc != null) {
-                gcs.add(gc);
+        if (garbageCollectors == null) {
+            synchronized (this) {
+                if (garbageCollectors == null) {
+                    List<BinaryGarbageCollector> list = new ArrayList<>();
```
efge: Could be called `gcs` now that it's not a global field anymore :)

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -434,19 +473,24 @@ public BinaryManagerStatus garbageCollectBinaries(boolean delete) {
 
     @Override
     public void markReferencedBinary(String key, String repositoryName) {
-        BlobProvider blobProvider = getBlobProvider(key, repositoryName);
-        BinaryGarbageCollector gc = blobProvider.getBinaryGarbageCollector();
-        if (gc != null) {
-            key = stripBlobKeyPrefix(key);
-            gc.mark(key);
+        List<BinaryGarbageCollector> gcs = getGarbageCollectors();
```
efge: This line can be moved down in the `if (sharedStorage)` section

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -361,22 +375,47 @@ protected String stripBlobKeyPrefix(String key) {
...
+        }
+        return getGarbageCollectors();
     }
 
     @Override
     public BinaryManagerStatus garbageCollectBinaries(boolean delete) {
         // do the GC in a long-running transaction to avoid timeouts
         return runInTransaction(() -> {
-            List<BinaryGarbageCollector> gcs = getGarbageCollectors();
+            List<BinaryGarbageCollector> gcs = getGarbageCollectors(true);
```
efge, line 413: We should explain in a comment why we need a "refreshed" list of GCs here and not elsewhere. Essentially, it's an optimization to avoid recomputing things too often while in the mark phase of a GC process.

efge (changes requested):

nuxeo-core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/blob/DocumentBlobManagerComponent.java
```diff
@@ -70,6 +71,13 @@
 
     protected Deque<BlobDispatcherDescriptor> blobDispatcherDescriptorsRegistry = new LinkedList<>();
 
+    protected volatile List<BinaryGarbageCollector> garbageCollectors;
+
+    /**
+     * true when several blob providers share a same storage.
+     */
+    protected volatile boolean sharedStorage = false;
```
efge: Sorry a last one: ` = false` is implicit and should be removed.

efge (changes requested):

efge (approved):

kevinleturc (approved):

PR #5018, opened 2021-12-17
NXP-30709: Do not terminate computation in case of checkpoint error

bdelbosc:


efge:
Commit message has XNP instead of NXP.

nuxeo-runtime/nuxeo-stream/src/main/java/org/nuxeo/lib/stream/computation/log/ComputationRunner.java
```diff
@@ -174,51 +188,70 @@ public boolean waitForAssignments(Duration timeout) throws InterruptedException
...
-            if (normalTermination || interrupted) {
-                log.debug(metadata.name() + ": Terminated");
+        }
+    }
+
+    protected ReturnCode runOnce() {
+        try {
+            processLoop();
+        } catch (CheckPointException e) {
+            log.error(metadata.name() + ": Error during checkpoint, processing will be duplicated: " + e.getMessage(),
```
kevinleturc, line 229: As it is an expected failure, shouldn't we log at warn level?

kevinleturc: The `e.getMessage()` will also contains the metadata name, don't we want to remove it from the exception in `checkpointIfNecessary` method?

bdelbosc: I think the error level is ok because there is a risk of duplicate processing which can be a problem if the processing is not idempotent.
for the message the format is now unified using the computation name as a prefix to ease searching in log,

kevinleturc (approved):

guirenard (approved):

PR #5023, opened 2021-12-22
NXP-30684: Shortened pah strategy by default for local blob store

yjulienne-nuxeo:


nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/LocalBlobStoreConfiguration.java
```diff
@@ -45,6 +45,15 @@
...
     public static final String DEFAULT_PATH = "binaries";
 
+    /** In the initialization properties, the property for the store path strategy type (full|shortened). */
+    public static final String PROP_PATH_STRATEGY_TYPE = "pathStrategyType";
+
+    public static final String PATH_STRATEGY_TYPE_FULL = "full";
+
+    public static final String PATH_STRATEGY_TYPE_SHORTENED = "shortened";
+
+    public static final String DEFAULT_PATH_STRATEGY_TYPE = PATH_STRATEGY_TYPE_SHORTENED;
```
kevinleturc: I think we should keep PATH_STRATEGY_TYPE_FULL as default for 10.10.

efge: Yes this is necessary for compat. But I even suggest not doing that part of the change at all, see my comment in https://github.com/nuxeo/nuxeo-lts/pull/331#discussion_r773973855

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/LocalBlobStoreConfiguration.java
```diff
@@ -117,4 +126,29 @@ public BinaryManagerRootDescriptor getDescriptor(Path configFile, boolean create
...
+
+    public PathStrategy getPathStategy(KeyStrategy keyStrategy, Path dir) {
+        String pathStrategyType = getPathStrategyType();
+        if (keyStrategy.useDeDuplication()) {
+            switch (pathStrategyType) {
+            case PATH_STRATEGY_TYPE_FULL:
+                return new PathStrategySubDirs(dir, descriptor.depth);
+            case PATH_STRATEGY_TYPE_SHORTENED:
+            default:
+                return new PathStrategyShortenedSubDirs(dir, descriptor.depth);
```
kevinleturc: The default switch case should be along with PATH_STRATEGY_TYPE_FULL.

nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/PathStrategyShortened.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.nio.file.Path;
+
+import org.apache.commons.codec.digest.DigestUtils;
+
+/**
+ * Extends the default {@link PathStrategy#safePath(String)} method to ensure resulting path and file name is not too
+ * long.
+ *
+ * @since 2021.13
```
ataillefer: 2021.14 or 10.10-HF57, same below

ataillefer (dismissed):

kevinleturc (dismissed):

kevinleturc (approved):

ataillefer (approved):

efge (approved):

PR #5052, opened 2022-01-27
NXP-30834: allow DocumentModelCSVWriter to write documents from other repositories

troger:


troger:
Same fix as in https://jira.nuxeo.com/browse/NXP-30615.

nuxeo-features/nuxeo-platform-csv-export/src/main/java/org/nuxeo/ecm/platform/csv/export/io/DocumentModelCSVWriter.java
```diff
@@ -59,6 +63,15 @@ public DocumentModelCSVWriter() {
 
     @Override
     protected void write(DocumentModel entity, CSVPrinter printer) throws IOException {
+        if (entity.getSessionId() == null || entity.getId() == null) {
+            // do not try to re-attach a detached or non-existing document
+            doWriteEntityBody(entity, printer);
+        } else {
+            withDocumentAttached(entity, () -> doWriteEntityBody(entity, printer));
```
NourNuxeo, line 70: wouldn't it be better to create the method when we have more uses cases? I feel like this method has been created "in case" we need to make more use of it with yet-to-come different runnables?

ataillefer (dismissed):

ataillefer:
Rebased on 10.10.

troger:
Won't do.

PR #5056, opened 2022-02-03
NXP-30804: lazy init S3BlobStore.useAsyncDigest

NourNuxeo:
Avoid NPE on S3BlobStore init.
No test is planned for this PR.
The fix shall be forward-ported to https://github.com/nuxeo/nuxeo-lts/pull/253 before merging.


addons/nuxeo-core-binarymanager-cloud/nuxeo-core-binarymanager-s3/src/main/java/org/nuxeo/ecm/blob/s3/S3BlobStore.java
```diff
@@ -128,7 +128,7 @@
     /** If true, include the object version in the key. */
     protected final boolean useVersion;
 
-    protected final boolean useAsyncDigest;
+    protected Boolean useAsyncDigest;
```
efge: The double-checked locking idiom requires the field to be `volatile` otherwise it's incorrect.

efge (changes requested):

efge (approved):

troger (approved):

PR #5151, opened 2022-07-24
[Java 17]: Swap Nashorn for GraalVM

sgammon:
Hey there esteemed Nuxeo authors,

I would like to respectfully propose a pathway to Java 17 support. We have an internal set of patches to this effect which we can offer. Consequently we are sharing them here, with this being the first patch, as a starting point for feedback. Since Nashorn is dropped entirely in JDK15+, a swap to GraalVM makes sense, but if this change is not appropriate for merging up to main Nuxeo please just let us know 😁

With feedback, it should be possible to keep this change backwards-compatible with JDKs down to 11, perhaps 8.

---

This changeset drops Nashorn-specific classes which are no longer available in JDK15+ in accordance with [JEP372][1], and replaces them with [JSR223][2]-compatible calls.

These calls work with [GraalVM][3] to support a seamless switch to Graal JS when running on that platform. For other JDK17 runtimes, Graal JS dependencies can be included directly to run on a stock stock [JVM][4].

Changes enclosed:
- Drop references to Nashorn classes, replace with JSR223 calls
- Configure GraalJS with predicate-based class filter
- Fixup references to script mirrors as maps
- Compatibility with pre-JDK17 via Maven deps

[1]: https://openjdk.org/jeps/372
[2]: https://jcp.org/en/jsr/detail?id=223
[3]: https://www.graalvm.org/22.1/reference-manual/js/NashornMigrationGuide/
[4]: https://www.graalvm.org/22.1/reference-manual/js/RunOnJDK/

### DIFF
--- a/modules/platform/nuxeo-automation/nuxeo-automation-scripting/src/main/java/org/nuxeo/automation/scripting/internals/AutomationScriptingServiceImpl.java
+++ b/modules/platform/nuxeo-automation/nuxeo-automation-scripting/src/main/java/org/nuxeo/automation/scripting/internals/AutomationScriptingServiceImpl.java
@@ -23,6 +23,7 @@ import static org.nuxeo.automation.scripting.api.AutomationScriptingConstants.DE
 import static org.nuxeo.automation.scripting.api.AutomationScriptingConstants.NASHORN_JAVA_VERSION;
 import static org.nuxeo.automation.scripting.api.AutomationScriptingConstants.NASHORN_WARN_CACHE;
 import static org.nuxeo.automation.scripting.api.AutomationScriptingConstants.NASHORN_WARN_CLASS_FILTER;
+import static org.nuxeo.automation.scripting.api.AutomationScriptingConstants.NASHORN_WARN_ENGINE;
 import static org.nuxeo.launcher.config.ConfigurationChecker.checkJavaVersion;
 
 import java.io.InputStream;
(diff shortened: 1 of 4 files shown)


modules/platform/nuxeo-automation/nuxeo-automation-scripting/src/main/java/org/nuxeo/automation/scripting/internals/AutomationScriptingServiceImpl.java
```diff
@@ -31,24 +32,19 @@
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
+import java.util.function.Predicate;
+
+import javax.script.*;
```
sgammon, line 39: i suspect wildcard imports are banned by static analysis / code style, right? should i apply a fix for this?

modules/platform/nuxeo-automation/nuxeo-automation-scripting/src/main/java/org/nuxeo/automation/scripting/internals/AutomationScriptingServiceImpl.java
```diff
@@ -174,17 +170,20 @@ protected ScriptEngine getScriptEngine() {
...
-        return nashorn.getScriptEngine(args, classLoader, classFilter);
+        ScriptEngine engine = new ScriptEngineManager().getEngineByName("JavaScript");
+        if (engine == null) {
+            throw new UnsupportedOperationException(NASHORN_WARN_ENGINE);
+        }
+        Bindings bindings = engine.getBindings(ScriptContext.ENGINE_SCOPE);
+        bindings.put("polyglot.js.allowHostAccess", true);
+        Predicate<String> classFilter = filter ? getClassFilter() : null;
+        if (classFilter != null) {
+            bindings.put("polyglot.js.allowHostClassLookup", classFilter);
```
sgammon, line 181: i may need to equip this engine with Nashorn's compatibility mode to make this mergable upstream.

efge:
While it's been removed from recent JDKs, Nashorn is now available as a separate project, see https://github.com/openjdk/nashorn. So there is no urgent need to swap in GraalVM at this point.

sgammon:
@efge good point, would you guys prefer i use that? i did not know it was available as a standalone project with later JDK support.

graalvm would be usable toward older JDKs, has a community version, and the JS VM can be used with stock JDKs. it also offers better isolation at the VM layer, avoiding the need for class filtering etc (although Nuxeo already implements this, so there's no code savings for that point, and GraalVM is admittedly slower on earlier JDKs, I think).

what do you think? thank you for your response, i'm excited to contribute and happy to follow the Nuxeo team's guidelines.

PR #5154, opened 2022-08-05
NXP-31123: Reject multipart uploads when disabled

NourNuxeo:


NourNuxeo:
This PR will not be backported. nuxeo 10.10 behaves otherwise (supports multipart/form-data)

PR #5217, opened 2023-01-05
fix-NXP-31459-aceinfo-removal

NourNuxeo:


nuxeo-services/nuxeo-permissions/src/test/java/org/nuxeo/ecm/permissions/TestPermissionListener.java
```diff
@@ -193,6 +197,25 @@ public void shouldUpdateDirectory() throws Exception {
...
+        acp.addACE(ACL.LOCAL_ACL, new ACE("fry", "Browse", true));
+        doc.setACP(acp, true);
+
+        // Remove the document
+        session.removeDocument(doc.getRef());
+        txFeature.nextTransaction();
+
+        // Check ACE info is removed
+        try (Session dirSession = directoryService.open(ACE_INFO_DIRECTORY)) {
+            DocumentModelList entries = dirSession.query(Collections.singletonMap("docId", doc.getId()));
```
NourNuxeo, line 214: If the operators don't add manually the docId index this will never find anything.
Maybe the listener should be disabled by default ?

nuxeo-services/nuxeo-permissions/src/main/java/org/nuxeo/ecm/permissions/PermissionGarbageCollectorListener.java
```diff
@@ -0,0 +1,63 @@
...
+        if (!(ctx instanceof DocumentEventContext)) {
+            return;
+        }
+
+        if (DOCUMENT_REMOVED.equals(event.getName())) {
+            cleanDirectory((DocumentEventContext) ctx);
+        }
+    }
+
+    protected void cleanDirectory(DocumentEventContext ctx) {
```
bdelbosc: Could it be renamed to cleanAceInfo?

nuxeo-services/nuxeo-permissions/src/main/java/org/nuxeo/ecm/permissions/PermissionGarbageCollectorListener.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Override
+    public void handleEvent(Event event) {
+        EventContext ctx = event.getContext();
+        if (!(ctx instanceof DocumentEventContext)) {
+            return;
+        }
+
+        ConfigurationService configService = Framework.getService(ConfigurationService.class);
+        if (DOCUMENT_REMOVED.equals(event.getName())
```
bdelbosc: I guess there is no way to know if the removed document hold and ACP and require a cleaning?

nuxeo-services/nuxeo-permissions/src/main/resources/OSGI-INF/listeners-contrib.xml
```diff
@@ -8,6 +8,11 @@
       <event>documentSecurityUpdated</event>
     </listener>
 
+    <listener name="permissionGarbageCollectorListener" async="true" postCommit="false"
```
bdelbosc, line 11: what about?
```enabled="${nuxeo.aceinfo.gc.enabled:=false}"```
this will avoid the creation and the submit of a ListenerWork, and its execution to do nothing, also, it will work using nuxeo.conf which is not the case with your contrib.

File: nuxeo-services/nuxeo-permissions/src/main/java/org/nuxeo/ecm/permissions/PermissionGarbageCollectorListener.java (hunk 3 above)
Comment (NourNuxeo) [reply nuxeo-services/nuxeo-permissions/src/main/java/org/nuxeo/ecm/permissions/PermissionGarbageCollectorListener.java]: At this point if I want to check the doc `ACP`:
```((DocumentEventContext) ctx).getSourceDocument().getACP();```
I get a `DocumentNotFoundException`

```
2023-01-06 12:14:26,335 [main] ERROR [EventServiceImpl] Exception during permissionGarbageCollectorListener sync listener execution (after 1ms), continuing to run other listeners
org.nuxeo.ecm.core.api.DocumentNotFoundException: 8d1a455f-f14f-4390-bbad-b7218ca83c76
	at org.nuxeo.ecm.core.storage.sql.coremodel.SQLSession.getDocumentByUUID(SQLSession.java:185) ~[classes/:?]
	at org.nuxeo.ecm.core.api.AbstractSession.resolveReference(AbstractSession.java:340) ~[classes/:?]
	at org.nuxeo.ecm.core.api.AbstractSession.getACP(AbstractSession.java:570) ~[classes/:?]
	at org.nuxeo.ecm.core.api.impl.DocumentModelImpl.getACP(DocumentModelImpl.java:759) ~[classes/:?]
	at org.nuxeo.ecm.permissions.PermissionGarbageCollectorListener.handleEvent(PermissionGarbageCollectorListener.java:53) ~[classes/:?]
```

I don't know if there is another way to find out.

bdelbosc: yes this is expected, the listener is async so the doc is already removed, just wondering if there is another event or if it's worth creating an additional event like ACP_REMOVED

bdelbosc (approved):

kevinleturc (approved):

guirenard (approved):

PR #5235, opened 2023-03-01
NXP-31716: reset startTime when GC ends

tmartins:


nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/blob/binary/BinaryGarbageCollector.java
```diff
@@ -81,4 +81,12 @@ public interface BinaryGarbageCollector {
      */
     boolean isInProgress();
 
+    /**
+     * Reset start time. This is required if the process throws an error before the end not to block the next call to
+     * the GC
+     *
+     * @since 2021.36
+     */
+    void reset();
```
guirenard:
```suggestion
    default void reset() {};
```

tmartins: ok I've added the default implementation with a warn if a subclass do not override the reset method.

kevinleturc (approved):

PR #5246, opened 2023-03-22
NXP-24524: Add ES fields for version created and description

guirenard:


nuxeo-core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/DocumentModel.java
```diff
@@ -917,6 +925,8 @@ class DocumentModelRefresh {
 
         public String checkinComment;
 
+        public Calendar checkinDate;
```
ataillefer, line 928: `@since`?

ataillefer (approved):

bdelbosc (approved):

